PR #18, opened 2018-05-22
Update test acs-engine version

karataliu:


tests/k8s-azure/Dockerfile
```diff
@@ -13,7 +13,7 @@ RUN git clone https://github.com/kubernetes/kubernetes . \
 RUN rm -rf _output/local/go
 
 FROM golang:1.8.3-stretch AS build_acs-engine
-ARG ACSENGINE_VERSION=v0.14.0
+ARG ACSENGINE_VERSION=v0.17.0
```
feiskyer: we should always use latest release, the release tag could be get by command:

```
curl -L -s https://api.github.com/repos/Azure/acs-engine/releases/latest | jq -r .tag_name
```

karataliu: Latest release sometimes have regression such as: https://github.com/Azure/acs-engine/issues/3022 ,which brings all NFS case to failure.

Would prefer manually updating the version, this is like what we do in using fixed os image version inside acs-engine.

feiskyer: looks reasonable

PR #23, opened 2018-06-08
Import loadbalancer document

### LINKED ISSUE #11: Add documentation for azure service annotations
Azure cloud provider supports a lot of annotations: https://github.com/kubernetes/kubernetes/blob/master/pkg/cloudprovider/providers/azure/azure_loadbalancer.go#L36-L83, we should document them.

Also refer https://github.com/kubernetes/website/issues/8154

karataliu:
Import azure_loadbalancer.md and update annotation section.

Move loadbalancer doc from https://github.com/kubernetes/kubernetes/blob/v1.12.0-alpha.0/pkg/cloudprovider/providers/azure/azure_loadbalancer.md

Fixes #11

### DIFF
--- a/docs/development-loadbalancer.md
+++ b/docs/development-loadbalancer.md
@@ -1,0 +1,77 @@
+# Azure LoadBalancer
+
+The way azure define LoadBalancer is different with GCE or AWS. Azure's LB can have multiple frontend IP refs. The GCE and AWS can only allow one, if you want more, you better to have another LB. Because of the fact, Public IP is not part of the LB in Azure. NSG is not part of LB in Azure either. However, you cannot delete them in parallel, Public IP can only be delete after LB's frontend IP ref is removed. 
+
+For different Azure Resources, such as LB, Public IP, NSG. They are the same tier azure resources. We need to make sure there is no connection in their own ensure loops. In another words, They would be eventually reconciled regardless of other resources' state. They should only depends on service state.
+
+Despite the ideal philosophy above, we have to face the reality. NSG depends on LB's frontend ip to adjust NSG rules. So when we want to reconcile NSG, the LB should contain the corresponding frontend ip config.
+
+And also, For Azure, we cannot afford to have more than 1 worker of service_controller. Because, different services could operate on the same LB, concurrent execution could result in conflict or unexpected result. For AWS and GCE, they apparently doesn't have the problem, they use one LB per service, no such conflict.
+
+There are two load balancers per availability set internal and external. There is a limit on number of services that can be associated with a single load balancer.
+By default primary load balancer is selected. Services can be annotated to allow auto selection of available load balancers. Service annotations can also be used to provide specific availability sets that host the load balancers. Note that in case of auto selection or specific availability set selection, when the availability set is lost in case of downtime or cluster scale down the services are currently not auto assigned to an available load balancer.
+Service Annotation for Auto and specific load balancer mode
+
+- service.beta.kubernetes.io/azure-load-balancer-mode" (__auto__|as1,as2...)
+
+## Introduce Functions
+
+- reconcileLoadBalancer(clusterName string, service *v1.Service, nodes []*v1.Node, wantLb bool) (*network.LoadBalancer, error)
+  - Go through lb's properties, update based on wantLb
+  - If any change on the lb, no matter if the lb exists or not
+    - Call az cloud to CreateOrUpdate on this lb, or Delete if nothing left
+  - return lb, err
+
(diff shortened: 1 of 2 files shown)


docs/development-loadbalancer.md
```diff
@@ -0,0 +1,157 @@
+# Azure LoadBalancer
```
colemickens: Can you document how the "Standard LB" support works? (Does it work with VMSS > 100, does it work with multiple VMSS > 100, does it do the same rule reconcilliation?)

docs/development-loadbalancer.md
```diff
@@ -0,0 +1,157 @@
...
+
+For different Azure Resources, such as LB, Public IP, NSG. They are the same tier azure resources. We need to make sure there is no connection in their own ensure loops. In another words, They would be eventually reconciled regardless of other resources' state. They should only depends on service state.
+
+Despite the ideal philosophy above, we have to face the reality. NSG depends on LB's frontend ip to adjust NSG rules. So when we want to reconcile NSG, the LB should contain the corresponding frontend ip config.
+
+And also, For Azure, we cannot afford to have more than 1 worker of service_controller. Because, different services could operate on the same LB, concurrent execution could result in conflict or unexpected result. For AWS and GCE, they apparently doesn't have the problem, they use one LB per service, no such conflict.
```
colemickens: How is this ensured? Does the azure controller do leader election the way KCM does?

docs/development-loadbalancer.md
```diff
@@ -0,0 +1,157 @@
...
+    Value type: string
+
+    Description: Specify the resource group of load balancer objects that are not in the same resource group as the cluster.
+
+1. Allowed service tags
+   ```
+    metadata:
+      name: my-service
+      annotations:
+        service.beta.kubernetes.io/azure-allowed-service-tags: ""
```
colemickens: Can you elaborate more on this annotation? I don't understand what the description means.

docs/development-loadbalancer.md
```diff
@@ -0,0 +1,157 @@
...
+1. Shared security rule
+   ```
+    metadata:
+      name: my-service
+      annotations:
+        service.beta.kubernetes.io/azure-shared-securityrule: ""
+    ```
+    Value type: string
+
+    Description: Specify that the service should be exposed using an Azure security rule that may be shared with other service, trading specificity of rules for an increase in the number of services that can be exposed. This relies on the Azure "augmented security rules" feature.
```
colemickens: What is "augmented security rules" feature? (Is it an Azure thing or a cloudprovider config thing?) Can you add a link to more info/docs/example?

docs/development-loadbalancer.md
karataliu: Yes.
https://github.com/kubernetes/kubernetes/blob/v1.12.0-alpha.0/cmd/cloud-controller-manager/app/controllermanager.go#L169

docs/development-loadbalancer.md
karataliu: Yes, Standard LB might have some special behavior. Basically this PR is to move the document from k8s repository only. 

@feiskyer  Do you have some comment on this? Suppose we could add the information in follow up PR.

docs/development-loadbalancer.md
karataliu: Introduced in https://github.com/kubernetes/kubernetes/pull/55740

@rjtsdl @itowlson  Could you please help elaborate what is 'service.beta.kubernetes.io/azure-shared-securityrule' for?

docs/development-loadbalancer.md
karataliu: The tags represent a group of security group settings:
https://docs.microsoft.com/en-us/azure/virtual-network/security-overview#service-tags

Will add a reference link.

docs/development-loadbalancer.md
karataliu: 'augmented security rules' itself is an Azure thing https://docs.microsoft.com/en-us/azure/virtual-network/security-overview#augmented-security-rules

docs/development-loadbalancer.md
feiskyer: yep, we may add it in a following PR.

@colemickens Standard load balancer supports multiple vmss/vmas with nodes up to 1000. It reconcile NSGs and Routes same like basic LB, but there're also differences of course:

- (1) It adds all nodes and masters (controlled by setting `excludeMasterFromStandardLB`) in same backend pools
- (2) It creates standard SKU floating IP
- (3) Because of (1), it doesn't support `service.beta.kubernetes.io/azure-load-balancer-mode`.

docs/development-loadbalancer.md
rjtsdl: @itowlson has more context :)

PR #24, opened 2018-06-08
Typo fix: configruation -> configuration

mirake:


feiskyer:
@mirake. Thanks for the fix.

/ok-to-test

PR #47, opened 2018-08-15
add azure internal and subnet service annotation test

fseldow:
Add tests for load balancer annotations
Add azure resource framework
Rename some of the functions


fseldow:
\retest

PR #57, opened 2018-09-06
Add documentation for Azure availability zones

feiskyer:
Add docs for AZ: https://github.com/kubernetes/features/issues/586.

/assign @justaugustus

### DIFF
--- a/docs/using-availability-zones.md
+++ b/docs/using-availability-zones.md
@@ -1,0 +1,218 @@
+# Availability Zones
+
+**Feature Status:** Alpha since v1.12.
+
+Kubernetes v1.12 adds support for [Azure availability zones (AZ)](https://azure.microsoft.com/en-us/global-infrastructure/availability-zones/). Nodes in availability zone will be added with label `failure-domain.beta.kubernetes.io/zone=<region>-<AZ>` and topology-aware provisioning is added for Azure managed disks storage class.
+
+**TOC:**
+
+<!-- TOC -->
+
+- [Availability Zones](#availability-zones)
+    - [Pre-requirements](#pre-requirements)
+    - [Node labels](#node-labels)
+    - [Load Balancer](#load-balancer)
+    - [Managed Disks](#managed-disks)
+        - [StorageClass examples](#storageclass-examples)
+        - [PV examples](#pv-examples)
+    - [Appendix](#appendix)
+    - [Reference](#reference)
+
+<!-- /TOC -->
+
+## Pre-requirements
+
+Because only standard load balancer is supported with AZ, it is a prerequisite to enable AZ for the cluster. It should be configured in Azure cloud provider configure file (e.g. `/etc/kubernetes/azure.json`):
+
+```json
+{
+    "loadBalancerSku": "standard",
+    ...
+}
+```
+
+If topology-aware provisioning feature is used, feature gate `VolumeScheduling` should be enabled on master components (e.g. kube-apiserver, kube-controller-manager and kube-scheduler).
+
+## Node labels
+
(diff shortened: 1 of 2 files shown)


docs/using-availability-zones.md
```diff
@@ -0,0 +1,220 @@
...
+Because only standard load balancer is supported with AZ, it is a prerequisite to enable AZ for the cluster. It should be configured in Azure cloud provider configure file (e.g. `/etc/kubernetes/azure.json`):
+
+```json
+{
+    "loadBalancerSku": "standard",
+    ...
+}
+```
+
+If topology-aware provisioning feature is used, feature gates `VolumeScheduling` and `DynamicProvisioningScheduling` should be enabled on master components (e.g. kube-apiserver, kube-controller-manager and kube-scheduler).
```
msau42: We've removed the DynamicProvisioningScheduling feature gate in 1.12.  So only VolumeScheduling is required.

docs/using-availability-zones.md
```diff
@@ -0,0 +1,220 @@
...
+## Load Balancer
+
+`loadBalancerSku` has been set to `standard` in cloud provider configure file, so standard load balancer and standard public IPs will be provisioned automatically for services with type `LoadBalancer`. Both load balancer and public IPs are zone redundant.
+
+## Managed Disks
+
+Zone-aware and topology-aware provisioning are supported for Azure managed disks. To support these features, a few options are added in AzureDisk storage class:
+
+- **zoned**: indicates whether new disks are provisioned with AZ. Default is true.
+- **zone** and **zones**: indicates which zones should be used to provision new disks (zone-aware provisioning). Only can be set if `zoned` is not false and allowedTopologies is not set.
```
msau42: Considering that we are deprecating the zone/zones parameters for aws/gce in 1.12, replacing it with allowedTopologies which is beta, and this is just newly being introduced for azure as alpha, do you want to consider leaving it out?

docs/using-availability-zones.md
```diff
@@ -0,0 +1,220 @@
...
+
+## Load Balancer
+
+`loadBalancerSku` has been set to `standard` in cloud provider configure file, so standard load balancer and standard public IPs will be provisioned automatically for services with type `LoadBalancer`. Both load balancer and public IPs are zone redundant.
+
+## Managed Disks
+
+Zone-aware and topology-aware provisioning are supported for Azure managed disks. To support these features, a few options are added in AzureDisk storage class:
+
+- **zoned**: indicates whether new disks are provisioned with AZ. Default is true.
```
msau42: Will this cause backwards compatibility issues for users that have StorageClasses before this change?

docs/using-availability-zones.md
```diff
@@ -0,0 +1,220 @@
+# Availability Zones
+
+**Feature Status:** Alpha since v1.12.
```
msau42: Is there a feature gate that users have to set to enable this?

docs/using-availability-zones.md
feiskyer: Yep, thanks for tip

docs/using-availability-zones.md
feiskyer: yep, let me remove it from this usage guide

docs/using-availability-zones.md
feiskyer: No, azure supports both zoned and unzoned nodes together. If there's no zoned nodes in the cluster, then the disks will still be provisioned without zone.

docs/using-availability-zones.md
feiskyer: Noop, this is enabled for v1.12, so that new zoned nodes could be joined easily after upgrading an existing cluster.

PR #69, opened 2018-11-09
fix typos

mikeweiwei:


feiskyer:
@mikeweiwei  Thanks for the fix.

PR #90, opened 2019-02-03
Update documentation on multiple scale set support

dharmab:
Correct an error in the documentation stating that you cannot add multiple scale sets to a load balancer. You can in fact do this as long as you are using the Load Balancer Standard SKU.

```release-note
NONE
```


justaugustus:
/lgtm

...but I'll allow others to validate and `/approve`:
/assign @feiskyer @andyzhangx 
/ok-to-test

PR #105, opened 2019-03-07
add timeout to ccm e2e

ritazh:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Add timeout to ccm e2e test suite

Fixes #

**Release note**:
```
Add timeout to ccm e2e test suite
```
cc @feiskyer

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -95,4 +95,4 @@ test-e2e: image hyperkube
 		-chyperkube_image=$(HYPERKUBE_IMAGE)
 
 test-ccm-e2e:
-	go test ./tests/e2e/
+	go test ./tests/e2e/ -timeout 0


Makefile
```diff
@@ -95,4 +95,4 @@ test-e2e: image hyperkube
 		-chyperkube_image=$(HYPERKUBE_IMAGE)
 
 test-ccm-e2e:
-	go test ./tests/e2e/
+	go test ./tests/e2e/ -timeout 0
```
andyzhangx: so what's the timeout value here?

ritazh: `0` disables the default 10m timeout. As we run this more in the future, we will get a sense of what timeout value we should set here.

feiskyer: Reasonable

PR #108, opened 2019-03-11
 Refine load balancer annotations and SLB docs

### LINKED ISSUE #87: Pre existing Nodes will be removed from the LB's backend on deallocation causing them to loose outbound IP at boot up
Running k8s `v1.11.5`
Following this flow

1) we create an AS with 10 VM, they all get a outbound IP defined by the AS outbound IP.
2) we install kubelet, kubelet starts registering to the ARM API using the public IP (outbound IP)
3) Eventually, all nodes are registered in k8s. 
4) A user creates a Service of Type LoadBalancer, the outboundIP if the nodes changes to the frontEnd IP of that LoadBalancer service.
5) Deallocate a VM, the node disappear from the backend pool and become NotReady. It doesn't have an outboundIP defined anymore

The step above takes some time.  > 24h I would say. 

here's a snapshot of the k8s LB backend pool after a while, as you can see, we should have `kn-es-0`, `kn-es-1`, `kn-es-2`, ect, in the list and many more. Maybe this is the expected behavior, but it causes the node to not have a valid Outbound IP when it gets allocated again. 

screen shot 2019-01-31 at 06 59 56

feiskyer:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

/kind documentation

 Refine load balancer annotations and SLB docs.

Fixes #87

**Release note**:
```
NONE
```

/hold


feiskyer:
/assign @andyzhangx @justaugustus 

@andyzhangx @justaugustus  Could you help to have a look at the PR?

justaugustus:
@feiskyer -- this is significantly better than what we had documented before.
I'll take an action item to copy-edit, but let's get these changes in ASAP.

/lgtm
/approve

PR #119, opened 2019-03-14
Set timeout to a larger value

feiskyer:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

/kind failing-test

The current e2e tests are pretty flaky, set the timeouts to large values for reducing the failing tests.

Fixes #

**Release note**:
```
NONE
```

### DIFF
--- a/tests/e2e/utils/service_utils.go
+++ b/tests/e2e/utils/service_utils.go
@@ -28,6 +28,10 @@ import (
 	clientset "k8s.io/client-go/kubernetes"
 )
 
+const (
+	serviceTimeout = 20 * time.Minute
+)
+
 // DeleteService deletes a service
 func DeleteService(cs clientset.Interface, ns string, serviceName string) error {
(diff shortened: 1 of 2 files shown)


tests/e2e/utils/service_utils.go
```diff
@@ -28,6 +28,10 @@ import (
 	clientset "k8s.io/client-go/kubernetes"
 )
 
+const (
+	serviceTimeout = 20 * time.Minute
```
andyzhangx: 10 min to delete a service is not enough?

feiskyer: it should work for deletion, but for creation, 10 minutes may be not working

feiskyer:
The autoscaling tests is still failing, but the basic test has been passed now.

@ritazh Let's get this in first? I think there may be some edge cases not handled correctly in autoscaling tests. We may fix that in another PR.

PR #125, opened 2019-03-22
Make autoscaler e2e test more stable

fseldow:
Autoscaler e2e test exceeds time or returns 'Unexpected EOF' error

```release-note
NONE
```

### DIFF
--- a/tests/e2e/autoscaling/autoscaler.go
+++ b/tests/e2e/autoscaling/autoscaler.go
@@ -17,7 +17,6 @@ limitations under the License.
 package autoscaling
 
 import (
-	"fmt"
 	"strconv"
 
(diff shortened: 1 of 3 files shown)


fseldow:
> It may take a couple minutes for the CLA signature to be fully registered; after that, please reply here with a new comment and we'll verify.  Thanks.

re-verify

feiskyer:
The failing case is `CSI volume limit information using mock driver should report attach limit when limit is bigger than 0`.

@andyzhangx Since this is actually a mock CSI, should we pre-setup something for the test?

feiskyer:
@fseldow pull-cloud-provider-azure-e2e is still flaky.  We have made it as optional now. Could you close this PR and open a new one?

PR #126, opened 2019-03-22
fix: Allow IMAGE_TAG to be overridden

jchauncey:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Allows the user to override the IMAGE_TAG value in the Makefile. This is useful when building ccm in a CI environment.

Fixes #

**Release note**:
```
Allow user to override the IMAGE_TAG var in the Makefile
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,7 +37,7 @@ K8S_BRANCH ?=
 CCM_E2E_ARGS ?= -ginkgo.skip=\\[Serial\\]\\[Slow\\]
 
 IMAGE_NAME=azure-cloud-controller-manager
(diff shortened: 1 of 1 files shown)


jchauncey:
cc @ritazh @feiskyer 

This is so we can build the MCR image in CI and specify the tag at build time.

feiskyer:
e2e is failing because of https://github.com/kubernetes/test-infra/pull/11887, let's get this in first since the e2e fixes would spend much more time

PR #134, opened 2019-03-28
add ensure load balancer e2e test

fseldow:
Add tests when updating service load balancer including:

1. assigning to specific IP when updating public service
2. updating internal IP when updating internal service
3. updating an internal service to a public service with assigned IP
4. no operation since no change in service when update

**Release note**:
```
none
```


tests/e2e/network/ensureloadbalancer.go
```diff
@@ -0,0 +1,276 @@
+/*
+Copyright 2018 The Kubernetes Authors.
```
feiskyer: nit: 2019

fseldow:
\reopen

PR #154, opened 2019-04-27
Migrate legacy provider code into repo packages

justaugustus:
Signed-off-by: Stephen Augustus <saugustus@vmware.com>

**What type of PR is this?**
/kind cleanup
/kind design
/kind feature

Fixes #

First pass at migrating the Azure cloud provider code to packages within this repo.

**Release note**:
```
Initial commit of out-of-tree Azure cloud-controller-manager
```


justaugustus:
/assign @feiskyer @andyzhangx 
/milestone v1.15
/priority important-soon

Putting a hold on this for discussion:
/hold

feiskyer:
CSI drivers for AzureFile/AzureDisk are still not ready yet (especially for Windows), and azure credential provider is also depending on auth package. Both of them should be going to beta in v1.15.

 I'm in favor of migrating this only after CSI and azure credential provider ready for out-of-tree (probably 1.16?).

@justaugustus @andyzhangx WDYT?

justaugustus:
@feiskyer -- this only copies the in-tree provider into this repo. The legacy provider would still exist [here](https://github.com/kubernetes/kubernetes/tree/master/staging/src/k8s.io/legacy-cloud-providers/azure) and be published [here](https://github.com/kubernetes/legacy-cloud-providers/tree/master/azure), so AzureFile/AzureDisk/credential provider could still vendor off of the in-tree.

feiskyer:
>  this only copies the in-tree provider into this repo. The legacy provider would still exist here and be published here, so AzureFile/AzureDisk/credential provider could still vendor off of the in-tree.

Right, but since cloud-provider-azure is not GA, new commits should still be pushed to k/k and then copy here.

andyzhangx:
@justaugustus it's better do vendoring in this cloud-provider-azure, new commits should still be pushed into `kubernetes/staging/src/k8s.io/legacy-cloud-providers/azure` since cloud-provider-azure is not GA, I think there is a long time to go, we don't want to maintain two copies of azure cloud provider code.

justaugustus:
Totally agree! I should've mentioned that this was more of an experiment to see how much effort/if we were in a good place to be able to move fully out of tree. 

Thanks for the feedback. I'm going to look over the KEPs and add a few more details as necessary.
/close

PR #286, opened 2020-01-26
Update out-of-tree examples

ialidzhikov:
**What type of PR is this?**
/kind documentation

Update out-of-tree examples

Fixes #

**Release note**:
```
NONE
```

### DIFF
--- a/examples/out-of-tree/kube-apiserver.yaml
+++ b/examples/out-of-tree/kube-apiserver.yaml
@@ -10,7 +10,7 @@ spec:
   hostNetwork: true
   containers:
     - name: kube-apiserver
-      image: k8s.gcr.io/hyperkube-amd64:v1.16.0-beta.1
+      image: k8s.gcr.io/hyperkube-amd64:v1.17.2
       imagePullPolicy: IfNotPresent
       command: ["/hyperkube", "kube-apiserver"]
       args:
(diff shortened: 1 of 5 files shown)


examples/out-of-tree/kube-apiserver.yaml
```diff
@@ -10,7 +10,7 @@ spec:
   hostNetwork: true
   containers:
     - name: kube-apiserver
-      image: k8s.gcr.io/hyperkube-amd64:v1.16.0-beta.1
+      image: k8s.gcr.io/hyperkube-amd64:v1.16.0
```
feiskyer: Thanks for the update.

since azure-cloud-node-manager:v0.4.1 only works with v1.17, could you replace the hyperkube tag with v1.17.2?

examples/out-of-tree/kube-controller-manager.yaml
```diff
@@ -10,7 +10,7 @@ spec:
   hostNetwork: true
   containers:
     - name: kube-controller-manager
-      image: k8s.gcr.io/hyperkube-amd64:v1.16.0-beta.1
+      image: k8s.gcr.io/hyperkube-amd64:v1.16.0
```
feiskyer: same here and below

examples/out-of-tree/kube-apiserver.yaml
feiskyer: ping @ialidzhikov

PR #333, opened 2020-05-05
use distroless/static as base image

sozercan:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature

> /kind flake

- Current base image (`debian-base:v1.0.0`) has quite a few vulnerabilities, moving to `distroless/static` for better security
- We were building both ccm and cnm for each image (with `make all`), building ccm or cnm now

**Release note**:
```
Use distroless/static as base image for CCM/CNM for Linux
```


Dockerfile
```diff
@@ -12,11 +12,11 @@
...
+FROM golang:1.13.9-stretch AS builder
 WORKDIR /go/src/sigs.k8s.io/cloud-provider-azure
 COPY . .
-RUN make ENABLE_GIT_COMMAND=true
+RUN make bin/azure-cloud-controller-manager ENABLE_GIT_COMMAND=true
 
-FROM k8s.gcr.io/debian-base:v1.0.0
-COPY --from=build /go/src/sigs.k8s.io/cloud-provider-azure/bin/azure-cloud-controller-manager /usr/local/bin
-RUN ln -s /usr/local/bin/azure-cloud-controller-manager /usr/local/bin/cloud-controller-manager
+FROM gcr.io/distroless/static@sha256:c6d5981545ce1406d33e61434c61e9452dad93ecd8397c41e89036ef977a88f4
```
feiskyer, line 20: where did the sha get here? could we use a tag instead?

Dockerfile
```diff
@@ -12,11 +12,11 @@
...
+COPY --from=builder /go/src/sigs.k8s.io/cloud-provider-azure/bin/azure-cloud-controller-manager /bin/cloud-controller-manager
```
feiskyer, line 21: could you keep the path same?

Dockerfile
sozercan: Distroless images doesn't have semantic version tags, and we shouldn't use latest as a best practice to have deterministic builds.

Dockerfile
sozercan: Path is just arbitrary here since base distroless image doesn't have a shell. I changed it to `/usr/local/bin`

PR #392, opened 2020-09-11
chore: set the cloud-provider flag to external

ZeroMagic:
**What type of PR is this?**
/kind flake

When using out-of-tree cloud provider, should set the flag --cloud-provider to external in kube-apiserver.

Fixes #

**Release note**:
```
NONE
```


examples/out-of-tree/kube-apiserver.yaml
```diff
@@ -26,7 +26,7 @@ spec:
         - "--bind-address=0.0.0.0"
         - "--client-ca-file=/etc/kubernetes/certs/ca.crt"
         - "--cloud-config=/etc/kubernetes/azure.json"
```
feiskyer, line 28: also remove L28

ZeroMagic: done

PR #456, opened 2020-12-29
feat: Add LogicalSectorSize support for UltraSSD

nearora-msft:
**What type of PR is this?**
/kind feature

Add LogicalSectorSize as a parameter so that azuredisk-csi can utilize it to set the logicalSectorSize for ultraSSD
related to : https://github.com/kubernetes-sigs/azuredisk-csi-driver/issues/559

Fixes #

**Release note**:
```
none
```

### DIFF
--- a/pkg/azureclients/diskclient/interface.go
+++ b/pkg/azureclients/diskclient/interface.go
@@ -26,7 +26,7 @@ import (
 
 const (
 	// APIVersion is the API version for compute.
-	APIVersion = "2019-11-01"
+	APIVersion = "2020-06-30"
 	// AzureStackCloudAPIVersion is the API version for Azure Stack
 	AzureStackCloudAPIVersion = "2019-07-01"
 	// AzureStackCloudName is the cloud name of Azure Stack
--- a/pkg/provider/azure_managedDiskController.go
+++ b/pkg/provider/azure_managedDiskController.go
@@ -67,6 +67,8 @@ type ManagedDiskOptions struct {
 	DiskIOPSReadWrite string
 	// Throughput Cap (MBps) for UltraSSD disk
 	DiskMBpsReadWrite string
+	// Logical sector size in bytes for Ultra disks
+	LogicalSectorSize int32
 	// if SourceResourceID is not empty, then it's a disk copy operation(for snapshot)
 	SourceResourceID string
 	// The type of source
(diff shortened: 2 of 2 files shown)


pkg/azureclients/diskclient/interface.go
```diff
@@ -26,7 +26,7 @@ import (
 
 const (
 	// APIVersion is the API version for compute.
-	APIVersion = "2019-11-01"
+	APIVersion = "2020-06-30"
```
feiskyer, line 29: @andyzhangx  do you know whether this API version is available on all clouds?

andyzhangx: need to confirm with storage team, also for this feature, should test on csi driver first, let's hold this PR now.

nearora-msft: Already tested with the csi driver. Was able to create ultra disks successfully.

nearora-msft: @andyzhangx Now that we have the confirmation that this version is available on all the clouds, can we stop holding this PR?

andyzhangx: @nearora-msft could you port this code change to csi driver vendor first and then add csi driver related change? I would like to confirm it works in CSI driver e2e test first, thanks.

nearora-msft: Created PR : https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/661

pkg/provider/azure_managedDiskController.go
```diff
@@ -67,6 +67,8 @@ type ManagedDiskOptions struct {
 	DiskIOPSReadWrite string
 	// Throughput Cap (MBps) for UltraSSD disk
 	DiskMBpsReadWrite string
+	// Logical sector size in bytes for Ultra disks
+	LogicalSectorSize int
```
andyzhangx, line 71: since `diskProperties.CreationData.LogicalSectorSize` is int32, what about set it as `int32`, then you don't need to convert

nearora-msft: Done

andyzhangx (approved):
/lgtm
/approve

pkg/provider/azure_managedDiskController.go
```diff
@@ -135,13 +137,20 @@ func (c *ManagedDiskController) CreateManagedDisk(options *ManagedDiskOptions) (
 			diskMBpsReadWrite = int64(v)
 		}
 		diskProperties.DiskMBpsReadWrite = to.Int64Ptr(diskMBpsReadWrite)
+
+		if options.LogicalSectorSize != 0 {
+			diskProperties.CreationData.LogicalSectorSize = to.Int32Ptr(options.LogicalSectorSize)
```
andyzhangx, line 142: it's better add a logging here to show `LogicalSectorSize`, that would be easier for debugging, WDYT?

nearora-msft: Wouldn't it be logged in the azuredisk logs that have the grpc request and response?

andyzhangx: it will be logged in grpc request, while this is the actual code part taking effect, since there are very few people would use this feature, it's worthwhile add a logging here

nearora-msft: Added

andyzhangx (approved):
/lgtm
/approve

PR #460, opened 2021-01-11
feat: add CreateAccount option in EnsureStorageAccount

andyzhangx:
**What type of PR is this?**
/kind feature

feat: add CreateAccount option in EnsureStorageAccount
This PR:
 - add CreateAccount option in EnsureStorageAccount
 - if CreateAccount is false(by default), use a random matching account if user does not provide an account name(current behavior is always use first matching account)

Fixes #

**Release note**:
```
none
```


feiskyer (approved):
/lgtm

PR #477, opened 2021-01-20
feat: vm client changes for Azure Stack Hub support

JesusAlvarezTorres:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

/kind feature

VM client changes for Azure Stack Hub support.

Fixes #

Part of a series of changes to add Azure Stack Hub support to all supported ARM resource clients.

**Release note**:
```
feat: vm client changes for Azure Stack Hub support
```

### DIFF
--- a/pkg/azureclients/vmclient/azure_vmclient.go
+++ b/pkg/azureclients/vmclient/azure_vmclient.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"strings"
 	"time"
 
 	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2020-06-30/compute"
@@ -42,6 +43,7 @@ var _ Interface = &Client{}
 type Client struct {
 	armClient      armclient.Interface
 	subscriptionID string
+	cloudName      string
 
 	// Rate limiting configures.
 	rateLimiterReader flowcontrol.RateLimiter
@@ -56,7 +58,11 @@ type Client struct {
 func New(config *azclients.ClientConfig) *Client {
 	baseURI := config.ResourceManagerEndpoint
 	authorizer := config.Authorizer
-	armClient := armclient.New(authorizer, baseURI, config.UserAgent, APIVersion, config.Location, config.Backoff)
+	apiVersion := APIVersion
+	if strings.EqualFold(config.CloudName, AzureStackCloudName) {
+		apiVersion = AzureStackCloudAPIVersion
+	}
+	armClient := armclient.New(authorizer, baseURI, config.UserAgent, apiVersion, config.Location, config.Backoff)
 	rateLimiterReader, rateLimiterWriter := azclients.NewRateLimiter(config.RateLimitConfig)
 
 	klog.V(2).Infof("Azure VirtualMachine client (read ops) using rate limit config: QPS=%g, bucket=%d",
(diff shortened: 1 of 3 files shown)


andyzhangx:
/ok-to-test

@JesusAlvarezTorres would you sign the CLA first?
cc @songjiaxun

JesusAlvarezTorres:
> /ok-to-test
> 
> @JesusAlvarezTorres would you sign the CLA first?
> cc @songjiaxun

Thanks @andyzhangx , just signed the CLA

pkg/azureclients/vmclient/azure_vmclient.go
```diff
@@ -71,6 +77,7 @@ func New(config *azclients.ClientConfig) *Client {
 		rateLimiterReader: rateLimiterReader,
 		rateLimiterWriter: rateLimiterWriter,
 		subscriptionID:    config.SubscriptionID,
+		cloudName:         config.CloudName,
```
andyzhangx, line 80: why add a new field here, what about use `config.CloudName` directly, refer to

https://github.com/kubernetes-sigs/cloud-provider-azure/blob/083cf6de2feb34b98873082c60b6cd0eaefc925b/pkg/azureclients/diskclient/azure_diskclient.go#L62-L64

pkg/azureclients/vmclient/azure_vmclient.go
```diff
@@ -42,6 +43,7 @@ var _ Interface = &Client{}
 type Client struct {
 	armClient      armclient.Interface
 	subscriptionID string
+	cloudName      string
```
feiskyer, line 46: this is not used?

pkg/azureclients/vmclient/azure_vmclient.go
JesusAlvarezTorres: for consistency, there are 2 clients (for example the interface client) that use an additional API version (the compute api version). In those cases the switch to the Azure Stack API version happens using this field as the config is not available on the method that uses the compute API version. I kept the implementation the same across all clients.

pkg/azureclients/vmclient/azure_vmclient.go
JesusAlvarezTorres: It is used on the new unit test and in other clients that use the compute API version (interface client for example). As mentioned on my reply to @andyzhangx above, I added cloudName to all clients for consistency.

pkg/azureclients/vmclient/azure_vmclient.go
andyzhangx: all clients are initialized from azure cloud config, what's the benefit to add a redundant field(`cloudName`), only for unit test?

https://github.com/kubernetes-sigs/cloud-provider-azure/blob/74926ac455ea163d9f0a3a7b3f3dd7c4332dc785/pkg/provider/azure.go#L557

pkg/azureclients/vmclient/azure_vmclient.go
feiskyer: thanks for clarification, then it makes sense

pkg/azureclients/vmclient/azure_vmclient.go
JesusAlvarezTorres: For unit test yes, and in this case for consistency with other clients like the interface client that actually do need the new field. For example the change in the interface client (already merged) uses the new field to check for the cloud name in a method outside of the initialization.
https://github.com/kubernetes-sigs/cloud-provider-azure/blob/9aae7de0440b8a87e29224ca94e9ef8c47f93e21/pkg/azureclients/interfaceclient/azure_interfaceclient.go#L190

andyzhangx: thanks for the clarification, lgtm.

andyzhangx (approved):
/lgtm
/approve

PR #486, opened 2021-01-21
output rerr to log rather than previous err from GET

devigned:
**What type of PR is this?**
>
> /kind api-change

/kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

The error that shows up in the logs when vmss fails to update is not the error from the PUT, but rather the error from the previous `ss.getVMSS(vmssName, azcache.CacheReadTypeDefault)`. This is _very_ misleading.

**Release note**:
```
Output the actual error when VMSS PUT fails rather than the error from the previous GET
```


CecileRobertMichon (commented):
/lgtm

CecileRobertMichon:
Do we need to get this backported to the in-tree provider?

PR #490, opened 2021-01-24
feat: skip account matching with special tags

andyzhangx:
**What type of PR is this?**
/kind feature

feat: skip account matching with special tags
This PR also select first matching account to make sure account matching is deterministic.

Fixes #

**Release note**:
```
feat: skip account matching with special tags
```


feiskyer (approved):
/lgtm

PR #525, opened 2021-02-27
Allow disabling AzureStackCloud API versions when using AzureStackCloud config on public cloud

feiskyer:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake
/kind bug

When using AzureStackCloud config on the public cloud to customize ARM endpoints, the API versions for AzureStackCloud would be too old for public Cloud environment. Hence, this PR adds another new cloud config option `DisableAzureStackCloud` to allow disable AzureStackCloud API versions.

Refer https://github.com/Azure/AKS/issues/2148#issuecomment-786972895.

The issue only happens on Kubernetes v1.20.x with out-of-tree cloud provider enabled.

**Release note**:
```
Allow disabling AzureStackCloud API versions when using AzureStackCloud config on public cloud
```

### DIFF
--- a/pkg/provider/azure.go
+++ b/pkg/provider/azure.go
@@ -239,6 +239,11 @@ type Config struct {
 	// DisableAvailabilitySetNodes disables VMAS nodes support when "VMType" is set to "vmss".
 	DisableAvailabilitySetNodes bool `json:"disableAvailabilitySetNodes,omitempty" yaml:"disableAvailabilitySetNodes,omitempty"`
 
+	// DisableAzureStackCloud disables AzureStackCloud support. It should be used
+	// when setting AzureAuthConfig.Cloud with "AZURESTACKCLOUD" to customize ARM endpoints
+	// while the cluster is not running on AzureStack.
+	DisableAzureStackCloud bool `json:"disableAzureStackCloud,omitempty" yaml:"disableAzureStackCloud,omitempty"`
+
 	// Tags determines what tags shall be applied to the shared resources managed by controller manager, which
 	// includes load balancer, security group and route table. The supported format is `a=b,c=d,...`. After updated
 	// this config, the old tags would be replaced by the new ones.
@@ -643,6 +648,7 @@ func (az *Cloud) getAzureClientConfig(servicePrincipalToken *adal.ServicePrincip
 		ResourceManagerEndpoint: az.Environment.ResourceManagerEndpoint,
 		Authorizer:              autorest.NewBearerAuthorizer(servicePrincipalToken),
 		Backoff:                 &retry.Backoff{Steps: 1},
+		DisableAzureStackCloud:  az.Config.DisableAzureStackCloud,
 	}
 
 	if az.Config.CloudProviderBackoff {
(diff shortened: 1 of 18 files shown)


pkg/provider/azure.go
```diff
@@ -239,6 +239,11 @@ type Config struct {
 	// DisableAvailabilitySetNodes disables VMAS nodes support when "VMType" is set to "vmss".
 	DisableAvailabilitySetNodes bool `json:"disableAvailabilitySetNodes,omitempty" yaml:"disableAvailabilitySetNodes,omitempty"`
 
+	// DisableAzureStackCloud disables AzureStackCloud support. It should be used
+	// when setting AzureAuthConfig.Cloud with "AZURESTACKCLOUD" to customize ARM endpoints
+	// while the cluster is not running on AzureStack.
+	DisableAzureStackCloud bool `json:"disableAzureStackCloud,omitempty" yaml:"disableAzureStackCloud,omitempty"`
```
andyzhangx, line 245: how can I apply this in CSI driver, we use `az.InitializeCloudFromSecret()` or `azure.NewCloudWithoutFeatureGates(f)` to get cloud object: 
https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/master/pkg/azuredisk/azure.go#L58

feiskyer: you could set `disableAzureStackCloud` in secret data part together with other cloud config options.

feiskyer: or, another option is setting cloud.config.DisableAzureStackCloud out side of the secret

andyzhangx: why not add a new parameter in `InitializeCloudFromSecret` and `NewCloudWithoutFeatureGates`,
`InitializeCloudFromSecret` read secret, while `NewCloudWithoutFeatureGates` read file, we have 3 CSI drivers at least, it's better change as little as possible in CSI driver side.

andyzhangx: Azure Disk CSI driver already has a flag `--support-azure-stack`, so if these two funcs have a new paramter(`supportAzureStack`), that's minimum change.

Also what about change in AKS rp setting for `azure-cloud-controller-manager`, and `azure-cloud-node-manager`? Shall we disable by default or only disable in AKS rp?

andyzhangx: example from Azure Disk CSI driver:
```
        - name: azuredisk
          image: mcr.microsoft.com/k8s/csi/azuredisk-csi:latest
          imagePullPolicy: IfNotPresent
          args:
            ...
            - "--support-azure-stack=true"
```

shall we also add a similar flag for `azure-cloud-controller-manager` and `azure-cloud-node-manager`, that's easier for configure.

feiskyer: cloud provider is not using InitializeCloudFromSecret now, hence more changes are needed if we change InitializeCloudFromSecret.

And the cloud name is setting in cloud config file (actually all azure specific configures are inside config file). Hence, it would be consistent when setting this new config together.

What's the overhead of setting this config inside secret? as what I understand, it would need no code changes besides vendoring the new library.

andyzhangx (approved):
/lgtm

PR #528, opened 2021-03-01
Route updates wait a while for new routes to take effect

### LINKED ISSUE #527: kubenet route table latency causing DNS resolution failure right after node creation
**What happened**:

When using kubenet network plugin together with cluster-autoscaler, new Pods scheduled to newly created node would get DNS resolution failure.

After talking with Azure Networking team, we confirmed this is because the route updates haven't take effect yet when the async updating returns success.

Route updates should take effect when the node is marked as network ready state.

**How to reproduce it**:

Create a cluster with kubenet plugin and enable cluster autoscaler. Then enable HPA for the workload Pods.

**Environment**:
- Kubernetes version (use `kubectl version`): 1.20 and 1.19.
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

feiskyer:
**What type of PR is this?**
>
> /kind api-change
/kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

When using kubenet network plugin together with cluster-autoscaler, new Pods scheduled to newly created node would get DNS resolution failure.

After talking with Azure Networking team, we confirmed this is because the route updates haven't take effect yet when the async updating returns success.

Since there's no API exposed to check whether routes take effect or not, this PR adds a delay after route updating requests (defautl is 30s). It also adds a new configuration option, so the delay could be tuned in the future.

Fixes #527

**Release note**:
```
Fixed routes not created issues before Pod scheduling. When using kubenet, 1) cloud-node-manager supports "--wait-routes=true" to indicate a node would wait for route updates before accepting Pod scheduling and  2) route controller would wait a while for new routes to take effect (default is 30s). 
```

### DIFF
--- a/cmd/cloud-node-manager/app/config/config.go
+++ b/cmd/cloud-node-manager/app/config/config.go
@@ -69,4 +69,8 @@ type Config struct {
 	// ClientConnection specifies the kubeconfig file and client connection
 	// settings for the proxy server to use when communicating with the apiserver.
 	ClientConnection componentbaseconfig.ClientConnectionConfiguration
+
+	// WaitForRoutes indicates whether the node should wait for routes to be created on Azure.
+	// If true, the node condition "NodeNetworkUnavailable" would be set to true on initialization.
+	WaitForRoutes bool
 }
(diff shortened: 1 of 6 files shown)


andyzhangx (approved):
/lgtm

PR #518, opened 2021-02-25
Enable more linters

CecileRobertMichon:
**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespaces from that line:
>
> /kind api-change
> /kind bug

/kind cleanup

> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

**What this PR does / why we need it**: Fix code and enable more linter. PR is already pretty big so kept some commented out and will follow up with other PRs if there are no objections. 

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:
Kept the commits separate so it's easier to review. Some of the linters required no changes, some (errorlint, goimports) required a lot of changes across many files, but mostly repetitive. 

No logic / behavior changes.

**Release note**:
```
NONE
```

PR #531, opened 2021-03-02
enable maligned linter

CecileRobertMichon:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

/kind cleanup

**What this PR does / why we need it**: Fix code and maligned gosec linter. Follow up from #518.

Fixes #

**Release note**:
```
NONE
```


pkg/provider/azure.go
```diff
@@ -173,35 +176,13 @@ type Config struct {
 	// the cloudprovider will try to add all nodes to a single backend pool which is forbidden.
 	// In other words, if you use multiple agent pools (scale sets), you MUST set this field.
 	PrimaryScaleSetName string `json:"primaryScaleSetName,omitempty" yaml:"primaryScaleSetName,omitempty"`
-	// Enable exponential backoff to manage resource request retries
-	CloudProviderBackoff bool `json:"cloudProviderBackoff,omitempty" yaml:"cloudProviderBackoff,omitempty"`
```
feiskyer, line 177: is this `maligned` rule required? this change would move such correlated configuration options to different places and would make future changes harder. Could we keep it disabled?

CecileRobertMichon: it's not "required" but it's best practice. It detects structs that would take less memory if their fields were sorted.

For example in this case 
```
pkg/provider/azure.go:131:13: struct of size 800 bytes could be of size 776 bytes (maligned)
```

feiskyer (commented):
/lgtm
/approve

pkg/provider/azure.go
feiskyer: OK, let's add this then.

PR #564, opened 2021-03-17
fix: check disk state before attach disk

andyzhangx:
**What type of PR is this?**
 /kind bug

fix: check disk state before attach disk

Should check both diskState and ManagedBy field before attach disk

Fixes #

**Release note**:
```
fix: check disk state before attach disk
```

/hold


ZeroMagic (approved):
/lgtm

PR #576, opened 2021-03-30
chore: use aks-ubuntu-16.04 distro for dual-stack and ipv6

aramase:
Signed-off-by: Anish Ramasekar <anish.ramasekar@gmail.com>

**What type of PR is this?**
/kind failing-test

Using `ubuntu` distro doesn't honor `runUnattendedUpgradesOnBootstrap: false` and the nodes get rebooted. This causes the API server reachability tests to fail for dual-stack. Switching to use `aks-ubuntu-16.04` instead.

Fixes #

**Release note**:
```
NONE
```


feiskyer (commented):
/lgtm
/approve

PR #579, opened 2021-04-06
Update compute API version to 2020-12-01

feiskyer:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

/kind cleanup

Update compute API version to 2020-12-01.

Fixes #

**Release note**:
```
Update Azure compute API version to 2020-12-01
```

### DIFF
--- a/pkg/azureclients/diskclient/azure_diskclient.go
+++ b/pkg/azureclients/diskclient/azure_diskclient.go
@@ -64,11 +64,7 @@ func New(config *azclients.ClientConfig) *Client {
 		apiVersion = AzureStackCloudAPIVersion
 	}
 
-	if config.ExtendedLocation != nil {
-		apiVersion = AzureEdgeZonesAPIVersion
-	}
-
-	klog.V(2).Infof("Azure DisksClient using API version: %s", AzureEdgeZonesAPIVersion)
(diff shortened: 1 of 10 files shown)


pkg/azureclients/diskclient/azure_diskclient.go
```diff
@@ -64,11 +64,7 @@ func New(config *azclients.ClientConfig) *Client {
 		apiVersion = AzureStackCloudAPIVersion
 	}
 
-	if config.ExtendedLocation != nil {
-		apiVersion = AzureEdgeZonesAPIVersion
-	}
-
-	klog.V(2).Infof("Azure DisksClient using API version: %s", AzureEdgeZonesAPIVersion)
+	klog.V(2).Infof("Azure DisksClient using API version: %s", apiVersion)
```
andyzhangx, line 67: `2020-12-01` is available on all regions including gov cloud?

andyzhangx: should be very cautious about this version change unless we could make sure it's available on all regions otherwise there could be break

feiskyer: yes, available now in all clouds.

PR #593, opened 2021-04-19
fix: potential race condition in detach disk

andyzhangx:
**What type of PR is this?**
/kind bug

fix: potential race condition in detach disk

Original detach disk logic hols `diskMap` data too long time and it would have potential race condition in such case. This PR only tries one(holding a lock) if it's a triable error, and unlock after retry. If there is still failure, k8s volume controller would retry in an outer loop.

Fixes #

/hold

**Release note**:
```
fix: potential race condition in detach disk
```

### DIFF
--- a/pkg/provider/azure_controller_common.go
+++ b/pkg/provider/azure_controller_common.go
@@ -312,6 +312,7 @@ func (c *controllerCommon) DetachDisk(diskName, diskURI string, nodeName types.N
 	}
 
 	c.lockMap.LockEntry(node)
+	defer c.lockMap.UnlockEntry(node)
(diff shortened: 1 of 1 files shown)


feiskyer (approved):
/lgtm

feiskyer:
@andyzhangx do we need to cherry-pick this to k/k?

andyzhangx:
> @andyzhangx do we need to cherry-pick this to k/k?

there is risk cherrying pick to k/k since the code logic is quite different now, also k/k is still attach one by one disk, the race condition is quite little and easy to recover, so answer is No.

PR #595, opened 2021-04-19
fix: arm node provider Windows initialization

JesusAlvarezTorres:
**What type of PR is this?**
>
> /kind api-change
/kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

PR fixes an issue on cloud node manager initialization of Windows nodes through ARM. Host name of the node manager Windows container is not equal to the node name, thus zone initialization cannot use the OS host name thus requires passing the node name as a parameter.

Fixes #

**Release note**:
```
fix: arm node provider Windows initialization

```

### DIFF
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -74,6 +74,6 @@ func (np *IMDSNodeProvider) InstanceType(ctx context.Context, name types.NodeNam
 // GetZone returns the Zone containing the current failure zone and locality region that the program is running in
 // In most cases, this method is called from the kubelet querying a local metadata service to acquire its zone.
 // If the node is not running with availability zones, then it will fall back to fault domain.
-func (np *IMDSNodeProvider) GetZone(ctx context.Context) (cloudprovider.Zone, error) {
+func (np *IMDSNodeProvider) GetZone(ctx context.Context, name types.NodeName) (cloudprovider.Zone, error) {
 	return np.azure.GetZone(ctx)
 }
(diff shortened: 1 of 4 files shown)


feiskyer:
There is a node-name arg [here](https://github.com/kubernetes-sigs/cloud-provider-azure/blob/master/cmd/cloud-node-manager/app/options/options.go#L127), @JesusAlvarezTorres have you set it on Windows?

JesusAlvarezTorres:
> There is a node-name arg [here](https://github.com/kubernetes-sigs/cloud-provider-azure/blob/master/cmd/cloud-node-manager/app/options/options.go#L127), @JesusAlvarezTorres have you set it on Windows?

Yes, I am setting that node-name arg on Windows. The problem however is not that argument, the problem comes from this line: https://github.com/kubernetes-sigs/cloud-provider-azure/blob/bfc60a7659c734a293baa6d8f96ce56afa8a07b1/pkg/provider/azure_zones.go#L151 
That is the current code path for get zone if not using IMDS. 
In Linux this works because the container can be configured to use host network in which case this call resolves to the host name. 
In Window's case host networking is not supported so this call resolves to the pod name which results in failed ARM calls when trying to fetch a node with the name of the Windows pod. With this change the ARM code path for Windows hosts is changed to do a GetZoneByNodeName instead of GetZone.

feiskyer (commented):
/ok-to-test
/lgtm
/approve

PR #609, opened 2021-04-27
chore: only logs rate limiting configurations when rate limit is enabled

feiskyer:
cherry-pick of #608 to release-0.7 (for k8s 1.20).

**Release note**:
```
chore: only logs rate limiting configurations when rate limit is enabled
```


feiskyer:
@nilo19 we need to enable pre-submit jobs for cherry-pick PRs.

PR #651, opened 2021-06-02
Add ability to reconcile Nodes from a single CNM replica

Danil-Grigorev:
**What type of PR is this?**

/kind api-change

This PR allows to serve multiple `Node` initialization from single `CNM` replica running outside of Nodes VM network. This is required to match the pattern in-tree CCM for Azure are doing, and simplifies migration process.

Fixes #

**Release note**:
```
Change of the --node-name behavior. If omitted, defaults to VM hostname. Setting it to emply value disables requirement for running it from inside the VM. Example Azure out-of-tree CCM is populating the value from "spec.nodeName" of the Pod.
```

### DIFF
--- a/cmd/cloud-node-manager/app/options/options.go
+++ b/cmd/cloud-node-manager/app/options/options.go
@@ -57,6 +57,8 @@ const (
 	CloudControllerManagerPort = 10263
 	// defaultNodeStatusUpdateFrequencyInMinute is the default frequency at which the manager updates nodes' status.
 	defaultNodeStatusUpdateFrequencyInMinute = 5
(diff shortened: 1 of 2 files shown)


feiskyer:
I think there would be some issues here because CNM would talk to IMDS to fetch some node metadata. Why are you proposing this change?
/hold

feiskyer:
ping @Danil-Grigorev on the questions above?

Danil-Grigorev:
@feiskyer You are right, will close this PR. Investigation showed that CCM cloud-node controller is sufficient to do the initial initialization for the first cluster Node, and CNM could later be started on any of the following.

PR #655, opened 2021-06-10
WIP: Adds docker buildx support for azure-cloud-node-manager image

claudiubelu:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

/sig windows
/kind feature

``docker buildx`` will allow us to build Windows images on a Linux node. Using this approach, we can build any Windows image for any OS Version without having to have the latest Windows node (typically, you cannot build Windows Container Images for OS Versions newer than the host's), and without having to rely on Hyper-V Support (typically, you cannot build Windows Container Images for OS Versions older than the host's without Hyper-V Isolation).

The image building approach is similar to the one used to build the Windows pause images and the E2E test images in Kubernetes:

- when using ``docker buildx`` for Windows images, the output type must be ``registry``,  which means that the image is being built and pushed.
- ``docker buildx`` does not add any ``os.version`` information in the images, which is necessary for manifest lists containing Windows images. We have to set it ourselves.
- pulling ``mcr.microsoft.com/windows/servercore:1809`` just to get ``netapi32.dll`` from it can be a bit inefficient, especially when considering the images is ~2GB. Instead, we can use ``gcr.io/k8s-staging-e2e-test-images/windows-servercore-cache:1.0-linux-amd64-$OSVERSION`` as a source for the dll, which is a scratch image containing a few commonly needed DLLs.

Fixes #

**Release note**:
```
TBA
```


Makefile
```diff
@@ -126,6 +147,29 @@ push: push-ccm-image push-node-image
 .PHONY: push-images
 push-images: push-ccm-image push-node-image
 
+.PHONY: push-node-manager-manifest
+push-node-manager-manifest:
+	docker manifest create --amend $(NODE_MANAGER_IMAGE) $(ALL_NODE_MANAGER_IMAGES)
+	docker manifest annotate --os linux --arch amd64 $(NODE_MANAGER_IMAGE) $(NODE_MANAGER_LINUX_IMAGE)
+	# For Windows images, we also need to include the "os.version" in the manifest list, so the Windows node can pull the proper image it needs.
+	# At the moment, docker manifest annotate doesn't allow us to set the os.version, so we'll have to it ourselves. The manifest list can be found locally as JSONs.
+	# See: https://github.com/moby/moby/issues/41417
```
aravindhp, line 156: Do we still need this given https://github.com/docker/cli/pull/2578 has merged?

andyzhangx: seems it's in `20.10.0`, that could make the logic t quite simple

claudiubelu: Right, I've updated this, so that we'll be using ``docker manifest annotate --os-version`` instead.

aravindhp: Thank you @claudiubelu. LGTM from me.

Makefile
```diff
@@ -37,6 +37,11 @@ STAGING_REGISTRY := gcr.io/k8s-staging-provider-azure
 K8S_VERSION ?= v1.18.0-rc.1
 HYPERKUBE_IMAGE ?= gcrio.azureedge.net/google_containers/hyperkube-amd64:$(K8S_VERSION)
 
+# The OS Version for the Windows images: 1809, 2004, 20H2
+OSVERSION ?= 1809
+ALL_OSVERSIONS = 1809 2004 20H2
```
feiskyer, line 42: Could you add WINDOWS prefix for these two envs?

PR #685, opened 2021-06-30
chore: increase log level for creating private endpoint in account

andyzhangx:
**What type of PR is this?**
 /kind documentation

chore: increase log level for creating private endpoint in account

Fixes #

**Release note**:
```
none
```

### DIFF
--- a/pkg/provider/azure_storageaccount.go
+++ b/pkg/provider/azure_storageaccount.go
@@ -277,7 +277,7 @@ func (az *Cloud) EnsureStorageAccount(accountOptions *AccountOptions, genAccount
 }
 
 func (az *Cloud) createPrivateEndpoint(ctx context.Context, accountName string, accountID *string, privateEndpointName string) error {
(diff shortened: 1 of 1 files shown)


pkg/provider/azure_storageaccount.go
```diff
@@ -277,7 +277,7 @@ func (az *Cloud) EnsureStorageAccount(accountOptions *AccountOptions, genAccount
 }
 
 func (az *Cloud) createPrivateEndpoint(ctx context.Context, accountName string, accountID *string, privateEndpointName string) error {
-	klog.V(5).Infof("Creating private endpoint(%s) for account (%s)", privateEndpointName, accountName)
+	klog.V(2).Infof("Creating private endpoint(%s) for account (%s)", privateEndpointName, accountName)
```
feiskyer, line 280: this is actually decreasing the log level

feiskyer (commented):
/lgtm

pkg/provider/azure_storageaccount.go
andyzhangx: make it print by default in csi driver

PR #716, opened 2021-07-21
fix: Disable zones on azure stack cloud

### LINKED ISSUE #715: Cloud controller manger crashes during startup on Azure Stack because of no zones
**What happened**:
Recent version of Azure cloud controller manager panicing during startup because of no zones support on Azure Stack
Error comes from this line: https://github.com/kubernetes-sigs/cloud-provider-azure/blob/13f66db1f3eb7dd00f7e35d0b281d4ca17965c61/pkg/provider/azure.go#L579

It should starts up successfully.

**How to reproduce it**:
Try to run CCM built after https://github.com/kubernetes-sigs/cloud-provider-azure/commit/7a105d9fd572e21dbc8b16af8182a8993d7b2007 this patch on Azure Stack

https://docs.microsoft.com/en-us/azure-stack/user/azure-stack-network-differences?view=azs-2102 - this doc say that Avalability Zones not yet supported on StackCloud.

**Environment**:
- kube 1.21.1, openshift 4.9
- ccm built from https://github.com/kubernetes-sigs/cloud-provider-azure/commit/61c726d73edb4204d608612fb8c198f493f0995d this revision

lobziik:
**What type of PR is this?**
/kind bug

Disabling zones check and other zone related bit on azure stack cloud due to it is not supported on this platform.
https://docs.microsoft.com/en-us/azure-stack/user/azure-stack-network-differences?view=azs-2102 

Fixes #715

**Release note**:
```
Zones logic on Azure Stack Cloud platform was disabled because it does not supported on this platform.
```

### DIFF
--- a/pkg/provider/azure.go
+++ b/pkg/provider/azure.go
@@ -573,14 +573,18 @@ func (az *Cloud) InitializeCloudFromConfig(config *Config, fromSecret, callFromC
 		az.routeUpdater = newDelayedRouteUpdater(az, routeUpdateInterval)
 		go az.routeUpdater.run()
 
-		// wait for the success first time of syncing zones
-		err = az.syncRegionZonesMap()
-		if err != nil {
-			klog.Errorf("InitializeCloudFromConfig: failed to sync regional zones map for the first time: %s", err.Error())
-			return err
-		}
+		// Azure Stack does not support zone at the moment
+		// https://docs.microsoft.com/en-us/azure-stack/user/azure-stack-network-differences?view=azs-2102
+		if !az.isStackCloud() {
+			// wait for the success first time of syncing zones
+			err = az.syncRegionZonesMap()
+			if err != nil {
+				klog.Errorf("InitializeCloudFromConfig: failed to sync regional zones map for the first time: %s", err.Error())
+				return err
+			}
 
-		go az.refreshZones(az.syncRegionZonesMap)
+			go az.refreshZones(az.syncRegionZonesMap)
+		}
 	}
 
 	return nil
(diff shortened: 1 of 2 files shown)


andyzhangx (commented):
/ok-to-test

pkg/provider/azure.go
```diff
@@ -573,14 +573,18 @@ func (az *Cloud) InitializeCloudFromConfig(config *Config, fromSecret, callFromC
...
 
-		// wait for the success first time of syncing zones
-		err = az.syncRegionZonesMap()
-		if err != nil {
-			klog.Errorf("InitializeCloudFromConfig: failed to sync regional zones map for the first time: %s", err.Error())
-			return err
-		}
+		// Azure Stack does not support zone at the moment
+		// https://docs.microsoft.com/en-us/azure-stack/user/azure-stack-network-differences?view=azs-2102
+		if !az.isStackCloud() {
```
feiskyer, line 578: Could we add another configure for this feature? Some deployments on Azure are using AzureStack config to customize ARM endpoints, hence this check would break those deployments

feiskyer (changes requested):

pkg/provider/azure.go
lobziik: https://github.com/kubernetes-sigs/cloud-provider-azure/blob/master/pkg/provider/azure.go#L175 -
```
	// DisableAzureStackCloud disables AzureStackCloud support. It should be used
	// when setting AzureAuthConfig.Cloud with "AZURESTACKCLOUD" to customize ARM endpoints
	// while the cluster is not running on AzureStack.
	DisableAzureStackCloud bool `json:"disableAzureStackCloud,omitempty" yaml:"disableAzureStackCloud,omitempty"
```

`isStackCloud` also check `disableAzureStackCloud` config parameter. This behaviour might be disabled using this config parameter, as described in comment.

feiskyer: thanks for adding this. LGTM then.

feiskyer (approved):
Thanks for the fix. Could you add some unit tests for the new logic? that would help to ensure the fixes won't be broken by future changes

feiskyer:
let's get this merged to unblock azurestack

/lgtm

PR #718, opened 2021-07-22
fix: avoid wrapping RawError twice

marwanad:
**What type of PR is this?**
>
> /kind api-change
/kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

The sender decorator builds a local `Error` type that embeds the RawError returned from the Azure service. However, in the client's `sendRequest` method it wraps this error again under the `Error` type.

```
retry.GetError(response, err)
```
This would return a double-wrapped error the consumers:
```
Retriable: false, RetryAfter: 0s, HTTPStatusCode: 409, RawError: Retriable: false, RetryAfter: 0s, HTTPStatusCode: 409, RawError: {"error":{"code":"ReadOnlyDisabledSubscription","message":"The subscription 'xxx' is disabled and therefore marked as read only. You cannot perform any write actions on this subscription until it is re-enabled."}}
 
```

Initially was thinking of safe guarding the behaviour but because `Error` doesn't implement the error interface, it seemed to be a bit of hassle.

With the fix in place, the top-level consumers will get the expected error body:

```
Retriable: false, RetryAfter: 0s, HTTPStatusCode: 409, RawError: {​​​​​​​"error":{​​​​​​​"code":"ReadOnlyDisabledSubscription","message":"The subscription 'xxx' is disabled and therefore marked as read only. You cannot perform any write actions on this subscription until it is re-enabled."}​​​​​​​}​​​​​​​
```

Fixes #

**Release note**:
```
N/A
```

### DIFF
--- a/pkg/retry/azure_retry.go
+++ b/pkg/retry/azure_retry.go
@@ -168,15 +168,19 @@ func doBackoffRetry(s autorest.Sender, r *http.Request, backoff *Backoff) (resp 
 		// 3) request has been throttled
 		// 4) request contains non-retriable errors
 		// 5) request has completed all the retry steps
-		if rerr == nil || !rerr.Retriable || rerr.IsThrottled() || backoff.isNonRetriableError(rerr) || backoff.Steps == 1 {
-			return resp, rerr.Error()
+		if rerr == nil {
+			return resp, nil
+		}
+
+		if !rerr.Retriable || rerr.IsThrottled() || backoff.isNonRetriableError(rerr) || backoff.Steps == 1 {
+			return resp, rerr.RawError
 		}
 
 		if !delayForBackOff(backoff, r.Context().Done()) {
 			if r.Context().Err() != nil {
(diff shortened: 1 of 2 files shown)


marwanad:
@feiskyer any reason we don't embed the ServiceError type instead of RawError? It seems that every code-path should return this.
https://github.com/Azure/go-autorest/blob/82fc42f340cec32d30ce97f9277f9b8aaf8885df/autorest/azure/azure.go#L50-L57

feiskyer (commented):
looks good

/retest

feiskyer:
>  any reason we don't embed the ServiceError type instead of RawError?

ideally, every azure api should report ServiceError, but we're not actually sure about that. So even with a ServiceError, we would still need another RawError if the responded error type couldn't be serialized to ServiceError.

feiskyer (commented):
/lgtm
/approve

PR #733, opened 2021-08-04
Add a "code" dimension to the Azure API error metric

marwanad:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
/kind feature
> /kind flake

This adds a more detailed "Code" label to the azure API errors metrics which is useful to classify the reason behind failed calls. For example, Quota, SubnetExhaustion, etc.
 
This can help unlock certain auto-healing scenarios.

The PR also changes some existing parsing of an LB error to be more robust and build on the same pattern using `autorest.ServiceError` instead of duplicating the fields.

Fixes #

**Release note**:
```
azure_api_request_errors metric now has an added "code" label which provides more details on the errors encountered.
```

### DIFF
--- a/pkg/azureclients/containerserviceclient/azure_containerserviceclient.go
+++ b/pkg/azureclients/containerserviceclient/azure_containerserviceclient.go
@@ -96,7 +96,7 @@ func (c *Client) Get(ctx context.Context, resourceGroupName string, managedClust
 	}
 
 	result, rerr := c.getManagedCluster(ctx, resourceGroupName, managedClusterName)
-	_ = mc.Observe(rerr.Error())
+	mc.Observe(rerr)
 	if rerr != nil {
 		if rerr.IsThrottled() {
 			// Update RetryAfterReader so that no more requests would be sent until RetryAfter expires.
@@ -157,7 +157,7 @@ func (c *Client) List(ctx context.Context, resourceGroupName string) ([]containe
 	}
 
 	result, rerr := c.listManagedCluster(ctx, resourceGroupName)
-	_ = mc.Observe(rerr.Error())
+	mc.Observe(rerr)
 	if rerr != nil {
 		if rerr.IsThrottled() {
 			// Update RetryAfterReader so that no more requests would be sent until RetryAfter expires.
@@ -318,7 +318,7 @@ func (c *Client) CreateOrUpdate(ctx context.Context, resourceGroupName string, m
 	}
 
 	rerr := c.createOrUpdateManagedCluster(ctx, resourceGroupName, managedClusterName, parameters, etag)
-	_ = mc.Observe(rerr.Error())
+	mc.Observe(rerr)
 	if rerr != nil {
 		if rerr.IsThrottled() {
 			// Update RetryAfterReader so that no more requests would be sent until RetryAfter expires.
@@ -393,7 +393,7 @@ func (c *Client) Delete(ctx context.Context, resourceGroupName string, managedCl
 	}
 
 	rerr := c.deleteManagedCluster(ctx, resourceGroupName, managedClusterName)
-	_ = mc.Observe(rerr.Error())
+	mc.Observe(rerr)
 	if rerr != nil {
 		if rerr.IsThrottled() {
 			// Update RetryAfterReader so that no more requests would be sent until RetryAfter expires.
--- a/pkg/azureclients/vmssclient/azure_vmssclient.go
+++ b/pkg/azureclients/vmssclient/azure_vmssclient.go
@@ -103,7 +103,7 @@ func (c *Client) Get(ctx context.Context, resourceGroupName string, VMScaleSetNa
 	}
 
 	result, rerr := c.getVMSS(ctx, resourceGroupName, VMScaleSetName)
-	_ = mc.Observe(rerr.Error())
+	mc.Observe(rerr)
 	if rerr != nil {
 		if rerr.IsThrottled() {
 			// Update RetryAfterReader so that no more requests would be sent until RetryAfter expires.
@@ -164,7 +164,7 @@ func (c *Client) List(ctx context.Context, resourceGroupName string) ([]compute.
 	}
 
 	result, rerr := c.listVMSS(ctx, resourceGroupName)
-	_ = mc.Observe(rerr.Error())
+	mc.Observe(rerr)
 	if rerr != nil {
 		if rerr.IsThrottled() {
 			// Update RetryAfterReader so that no more requests would be sent until RetryAfter expires.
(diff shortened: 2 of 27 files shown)


pkg/azureclients/containerserviceclient/azure_containerserviceclient.go
```diff
@@ -96,7 +96,7 @@ func (c *Client) Get(ctx context.Context, resourceGroupName string, managedClust
 	}
 
 	result, rerr := c.getManagedCluster(ctx, resourceGroupName, managedClusterName)
-	_ = mc.Observe(rerr.Error())
```
marwanad, line 99: This method used to return a error, but it's never used. Switched that to get rid of the return type.

pkg/azureclients/vmssclient/azure_vmssclient.go
```diff
@@ -288,39 +288,30 @@ func (c *Client) CreateOrUpdateAsync(ctx context.Context, resourceGroupName stri
 
 // WaitForCreateOrUpdateResult waits for the response of the create or update request
 func (c *Client) WaitForCreateOrUpdateResult(ctx context.Context, future *azure.Future, resourceGroupName string) (*http.Response, error) {
-	mc := metrics.NewMetricContext("vmss", "wait_for_create_or_update_result", resourceGroupName, c.subscriptionID, "")
-	res, err := c.armClient.WaitForAsyncOperationResult(ctx, future, "VMSSWaitForCreateOrUpdateResult")
-	_ = mc.Observe(err)
-	return res, err
+	return c.WaitForAsyncOperationResult(ctx, future, resourceGroupName, "wait_for_create_or_update_result", "VMSSWaitForCreateOrUpdateResult")
```
marwanad, line 291: Cleaned up the duplication and have everything call into `WaitForAsyncOperationResult`

pkg/azureclients/vmssclient/azure_vmssclient.go
```diff
@@ -288,39 +288,30 @@ func (c *Client) CreateOrUpdateAsync(ctx context.Context, resourceGroupName stri
...
+	return c.WaitForAsyncOperationResult(ctx, future, resourceGroupName, "wait_for_start_instances_result", "VMSSWaitForStartInstancesResult")
 }
 
 // WaitForAsyncOperationResult waits for the response of the request
-func (c *Client) WaitForAsyncOperationResult(ctx context.Context, future *azure.Future) (*http.Response, error) {
-	return c.armClient.WaitForAsyncOperationResult(ctx, future, "VMSSWaitForAsyncOperationResult")
+func (c *Client) WaitForAsyncOperationResult(ctx context.Context, future *azure.Future, resourceGroupName, request, asycOpName string) (*http.Response, error) {
+	mc := metrics.NewMetricContext("vmss", request, resourceGroupName, c.subscriptionID, "")
+	res, err := c.armClient.WaitForAsyncOperationResult(ctx, future, asycOpName)
+	mc.Observe(retry.NewErrorOrNil(false, err))
```
marwanad, line 313: `retry.NewErrorOrNil(false, err)` is some helper method which returns nil if err is nil and creates a retry.Error object wrapping the error if not.

This is just to avoid having to duplicate the `mc.Observe` calls.

pkg/provider/azure_backoff.go
```diff
@@ -384,7 +384,7 @@ func (az *Cloud) DeletePublicIP(service *v1.Service, pipResourceGroup string, pi
 }
 
 // DeleteLB invokes az.LoadBalancerClient.Delete with exponential backoff retry
-func (az *Cloud) DeleteLB(service *v1.Service, lbName string) error {
+func (az *Cloud) DeleteLB(service *v1.Service, lbName string) *retry.Error {
```
marwanad, line 387: Changed this to be able to parse the error Message nicer in azure_errors

pkg/retry/azure_error.go
```diff
@@ -327,47 +327,85 @@ func HasStatusForbiddenOrIgnoredError(err error) bool {
...
+	if err == nil || !isErrorLoadBalancerInUseByVirtualMachineScaleSet(err.RawError.Error()) {
+		return "", "", nil
 	}
 
-	rawErrorMap := make(map[string]*RawErrorContainer)
-	err := json.Unmarshal([]byte(matches[1]), &rawErrorMap)
-	if err != nil {
-		return nil, err
+	reg := regexp.MustCompile(`.*/subscriptions/(?:.*)/resourceGroups/(.*)/providers/Microsoft.Compute/virtualMachineScaleSets/(.+).`)
+	matches := reg.FindStringSubmatch(err.ServiceErrorMessage())
```
marwanad, line 337: That's the change for the load balancer error parsing logic. We don't need the `RawErrorContainer` since its doing the unnecessary round trip of converting to a string and building into a local struct.

pkg/retry/azure_error.go
```diff
@@ -327,47 +327,85 @@ func HasStatusForbiddenOrIgnoredError(err error) bool {
...
+func classifyErrorCode(sre azure.ServiceError) string {
+	if sre.Code == OperationNotAllowed {
+		return getOperationNotAllowedReason(sre.Message)
+	}
+	return sre.Code
+}
+
+// getOperationNotAllowedReason attempts to better classify OperationNotAllowed errors
+// by looking at the message
+func getOperationNotAllowedReason(msg string) string {
```
marwanad, line 406: This does a further classification on "OperationNotAllowed" errors to detect if the error is a quota one.

marwanad (commented):
commenting important parts

pkg/retry/azure_error.go
```diff
@@ -327,47 +327,85 @@ func HasStatusForbiddenOrIgnoredError(err error) bool {
...
-	if !IsErrorLoadBalancerInUseByVirtualMachineScaleSet(rawError) {
-		return "", "", nil
+const (
+	// OperationNotAllowed is an umbrella errrfor a lot of errors
+	OperationNotAllowed string = "OperationNotAllowed"
+	// QuotaExceeded falls under OperationNotAllowed error code but we make it more specific here
+	QuotaExceeded string = "QuotaExceeded"
+)
+
+// ServiceRawError wraps the RawError field satisfying autorest.ServiceError
```
marwanad, line 358: RawError under this package will always embed either azure.ServiceError or azure.DetailedError.

We don't care about azure.DetailedError for now since they seem to arise from ARM-context outside the providers.

pkg/azureclients/vmssclient/azure_vmssclient.go
feiskyer: much cleaner now, thanks for the improvement

feiskyer (approved):
/lgtm
/approve

PR #743, opened 2021-08-09
fix: ensure VM with provisioningState=Creating and powerState=Stopped  is treated as Creating instead of Stopped

### LINKED ISSUE #742: VMSS instance with powerstate stopped may be still under creating
**What happened**:

From Azure Virtual Machine [docs](https://docs.microsoft.com/en-us/azure/virtual-machines/states-billing), stopped power state means:

> The Virtual Machine is allocated on a host but not running. Also called PoweredOff state or Stopped (Allocated). This can be result of invoking the PowerOff API operation or invoking shutdown from within the guest OS. The Stopped state may also be observed briefly during VM creation or while starting a VM from Deallocated state.

This means the provisioning state should also be checked to determine a VM is stopped or not in `InstanceShutdownByProviderID()`.

https://github.com/kubernetes-sigs/cloud-provider-azure/blob/884b6c1124e6a18ef723b050dee00b19230c1c5f/pkg/provider/azure_instances.go#L258-L259

VM with provisioningState=Creating and powerState=Stopped should be treated as creating.

**How to reproduce it**:

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

feiskyer:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake
/kind bug

fix: ensure VM with provisioningState=Creating and powerState=Stopped  is treated as Creating instead of Stopped

Fixes #742

**Release note**:
```
fix: ensure VM with provisioningState=Creating and powerState=Stopped  is treated as Creating instead of Stopped
```

### DIFF
--- a/pkg/provider/azure_instances.go
+++ b/pkg/provider/azure_instances.go
@@ -28,6 +28,7 @@ import (
 	cloudprovider "k8s.io/cloud-provider"
 	"k8s.io/klog/v2"
 
+	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2020-12-01/compute"
 	azcache "sigs.k8s.io/cloud-provider-azure/pkg/cache"
 	"sigs.k8s.io/cloud-provider-azure/pkg/consts"
 )
@@ -253,10 +254,22 @@ func (az *Cloud) InstanceShutdownByProviderID(ctx context.Context, providerID st
 
 		return false, err
 	}
-	klog.V(5).Infof("InstanceShutdownByProviderID gets power status %q for node %q", powerStatus, nodeName)
+	klog.V(3).Infof("InstanceShutdownByProviderID gets power status %q for node %q", powerStatus, nodeName)
+
+	provisioningState, err := az.VMSet.GetProvisioningStateByNodeName(string(nodeName))
+	if err != nil {
+		// Returns false, so the controller manager will continue to check InstanceExistsByProviderID().
+		if errors.Is(err, cloudprovider.InstanceNotFound) {
+			return false, nil
+		}
+
+		return false, err
+	}
+	klog.V(3).Infof("InstanceShutdownByProviderID gets provisioning state %q for node %q", provisioningState, nodeName)
 
 	status := strings.ToLower(powerStatus)
-	return status == vmPowerStateStopped || status == vmPowerStateDeallocated || status == vmPowerStateDeallocating, nil
+	provisioningSucceeded := strings.EqualFold(strings.ToLower(provisioningState), strings.ToLower(string(compute.ProvisioningStateSucceeded)))
+	return provisioningSucceeded && (status == vmPowerStateStopped || status == vmPowerStateDeallocated || status == vmPowerStateDeallocating), nil
 }
 
 // InstanceShutdown returns true if the instance is shutdown according to the cloud provider.
--- a/pkg/provider/azure_instances.go
+++ b/pkg/provider/azure_instances.go
@@ -219,7 +219,7 @@ func (az *Cloud) InstanceExists(ctx context.Context, node *v1.Node) (bool, error
 	providerID := node.Spec.ProviderID
 	if providerID == "" {
 		var err error
-		providerID, err = az.getNodeProviderIDByNodeName(ctx, types.NodeName(node.Name))
+		providerID, err = cloudprovider.GetInstanceProviderID(ctx, az, types.NodeName(node.Name))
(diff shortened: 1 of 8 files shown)


pkg/provider/azure_instances.go
```diff
@@ -253,9 +254,14 @@ func (az *Cloud) InstanceShutdownByProviderID(ctx context.Context, providerID st
 
 		return false, err
 	}
-	klog.V(5).Infof("InstanceShutdownByProviderID gets power status %q for node %q", powerStatus, nodeName)
+	klog.V(5).Infof("InstanceShutdownByProviderID gets power status %q and provisioning state %q for node %q", powerStatus, provisioningState, nodeName)
 
 	status := strings.ToLower(powerStatus)
+	if status == vmPowerStateStopped && strings.EqualFold(strings.ToLower(provisioningState), strings.ToLower(string(compute.ProvisioningStateCreating))) {
```
marwanad, line 260: I don't think we should be doing/logging an "assumption". This is addressing a symptom rather than the cause.

I'd prefer if we the code is changed to:

```
return provisioningState == Succeeded && (status == vmPowerStateStopped || status == vmPowerStateDeallocated || status == vmPowerStateDeallocating)
```

pkg/provider/azure_standard.go
```diff
@@ -526,25 +526,31 @@ func (as *availabilitySet) GetInstanceIDByNodeName(name string) (string, error)
...
 // GetPowerStatusByNodeName returns the power state of the specified node.
-func (as *availabilitySet) GetPowerStatusByNodeName(name string) (powerState string, err error) {
+func (as *availabilitySet) GetPowerStatusByNodeName(name string) (powerState string, provisioningState string, err error) {
 	vm, err := as.getVirtualMachine(types.NodeName(name), azcache.CacheReadTypeDefault)
 	if err != nil {
-		return powerState, err
+		return powerState, provisioningState, err
 	}
 
+	if vm.VirtualMachineProperties == nil || vm.VirtualMachineProperties.ProvisioningState == nil {
```
marwanad, line 535: I'd prefer if we return a the nil status to the caller rather than default to something. This code is only called by `InstanceShutdownByProviderID` which is called by the cloudprovider's by "shutdownInCloudProvider".

I don't see it being an issue if we just return false instead of assuming it's stopped. Why would we want to taint a node that is under deletion with a shutdown annotation prematurely? 

It should already get the unreachable/non-executable taint

pkg/provider/azure_instances.go
marwanad: to reword my point: we should only act on **terminal states**  and avoid acting on transient states because they are more susceptible to latencies/API caching.

feiskyer: because of caching, there're high chances the transient states are cached for instances. waiting another 10 minutes for caching refreshing might be too long

marwanad: We only care about the shutdown state for the sake of adding the annotation right? We aren't modifying the cache state anywhere with that change. This change is just for returning the result to `InstancesShutDown` call which shouldn't affect anything else apart from the taint.

Looking at the controller-manager code, once the node reports Ready - the taint will be cleared up by that control loop independent of the cached state in cloudprovider.

pkg/provider/azure_standard.go
feiskyer: confirmed ProvisioningState would never be empty, would change the logs message

pkg/provider/azure_instances.go
```diff
@@ -244,7 +245,7 @@ func (az *Cloud) InstanceShutdownByProviderID(ctx context.Context, providerID st
 		return false, err
 	}
 
-	powerStatus, err := az.VMSet.GetPowerStatusByNodeName(string(nodeName))
+	powerStatus, provisioningState, err := az.VMSet.GetPowerStatusByNodeName(string(nodeName))
```
qike-ms, line 248: Can we create a new function to get provisioning state instead of piggyback the same?

pkg/provider/azure_instances.go
feiskyer: makes sense. changed according to above suggestions

pkg/provider/azure_instances.go
feiskyer: added the new interface

PR #784, opened 2021-08-31
feat: add AllowBlobPublicAccess setting in storage account creation

andyzhangx:
#### What type of PR is this?
/kind feature

#### What this PR does / why we need it:
feat: add AllowBlobPublicAccess setting in storage account creation
This PR also add `ctx context.Context` parameter in some functions.

#### Which issue(s) this PR fixes:

Fixes #

#### Does this PR introduce a user-facing change?

```release-note
feat: add AllowBlobPublicAccess setting in storage account creation
```

#### Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.:

```docs
feat: add AllowBlobPublicAccess setting in storage account creation
```

PR #785, opened 2021-08-31
feat: Add support to check and create storage accounts when accountOptions has accountName

mukhoakash:

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/pkg/provider/azure_storageaccount.go
+++ b/pkg/provider/azure_storageaccount.go
@@ -42,8 +42,8 @@ const PrivateDNSZoneName = "privatelink.file.core.windows.net"
 // AccountOptions contains the fields which are used to create storage account.
 type AccountOptions struct {
 	Name, Type, Kind, ResourceGroup, Location string
+	EnableHTTPSTrafficOnly                    bool
 	// indicate whether create new account when Name is empty
-	EnableHTTPSTrafficOnly                  bool
 	CreateAccount                           bool
 	EnableLargeFileShare                    bool
 	CreatePrivateEndpoint                   bool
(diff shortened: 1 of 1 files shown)


mukhoakash:
@andyzhangx  I will add the changes to pass contex once [this](https://github.com/kubernetes-sigs/cloud-provider-azure/pull/784) PR gets merged.

Working on adding unittests for the same.

pkg/provider/azure_storageaccount.go
```diff
@@ -118,25 +118,54 @@ func (az *Cloud) GetStorageAccesskey(account, resourceGroup string) (string, err
...
+	if accountOptions == nil {
+		return "", fmt.Errorf("account options is nil")
+	}
+
+	accountName := accountOptions.Name
+
+	if len(accountName) == 0 {
+		return "", fmt.Errorf("Account name not provided in account options")
+	} else {
+		accountExists, err := az.checkIfAccountExists(accountOptions)
```
andyzhangx, line 133: if you only need `checkIfAccountNotExists`, I think you could add this function into your CSI driver, it's not necessary to add this function here.

mukhoakash: I'll move the function `checkIfAccountNotExists` to the driver. But I would need the function to create the storage account.

andyzhangx: what about add a new field "CreateIfNotExists"(by default: false) in AccountOptions, that would be a minor change

mukhoakash: The EnsureStorageAccount function is designed in a way to work when accountName is empty in accountOptions. For my case, I would need to create an account with that particular accountName if it doesnt exist. So, the `createStorageAccount` function might be needed after all?

mukhoakash:
Abandoning this PR since we decided on modifying the EnsureStorageAccount function to accomodate the requirement.

PR #803, opened 2021-09-12
feat: reduce vmss cache refresh in parallel disk attach/detach

andyzhangx:
#### What type of PR is this?
/kind feature

feat: reduce vmss cache refresh in parallel disk attach/detach
This PR adds a lock in get vmss operation, in 1K disk attach/detach load test, it could reduce ~20%(67/338=20%) vmss cache refresh operations. It's not necessary to refresh cache every  time when nodeName is not found in vmss cache, especially in parallel workloads.

Fixes #

/hold

```release-note
feat: reduce vmss cache refresh in parallel disk attach/detach
```

```docs
feat: reduce vmss cache refresh in parallel disk attach/detach
```

### DIFF
--- a/pkg/provider/azure_vmss.go
+++ b/pkg/provider/azure_vmss.go
@@ -76,6 +76,8 @@ type ScaleSet struct {
 	vmssCache                 *azcache.TimedCache
 	vmssVMCache               *sync.Map // [resourcegroup/vmssname]*azcache.TimedCache
 	availabilitySetNodesCache *azcache.TimedCache
+	// lockMap in cache refresh
+	lockMap *lockMap
 }
 
 // newScaleSet creates a new ScaleSet.
@@ -93,6 +95,7 @@ func newScaleSet(az *Cloud) (VMSet, error) {
 		Cloud:           az,
 		availabilitySet: as,
 		vmssVMCache:     &sync.Map{},
+		lockMap:         newLockMap(),
 	}
 
 	if !ss.DisableAvailabilitySetNodes {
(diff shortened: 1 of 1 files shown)


feiskyer:
>  It's not necessary to refresh cache every time when nodeName is not found in vmss cache, especially in parallel workloads.

If the new nodes are created continuously by autoscaler, then the cache needs to be refreshed. or else, newly joined nodes may be deleted.

pkg/provider/azure_vmss.go
```diff
@@ -76,6 +76,8 @@ type ScaleSet struct {
 	vmssCache                 *azcache.TimedCache
 	vmssVMCache               *sync.Map // [resourcegroup/vmssname]*azcache.TimedCache
 	availabilitySetNodesCache *azcache.TimedCache
+	// lockMap in cache refresh
+	lockMap *lockMap
```
feiskyer, line 80: how is this map set and refreshed?

andyzhangx: it's in L98:
```
lockMap:         newLockMap(),
```

andyzhangx:
> > It's not necessary to refresh cache every time when nodeName is not found in vmss cache, especially in parallel workloads.
> 
> If the new nodes are created continuously by autoscaler, then the cache needs to be refreshed. or else, newly joined nodes may be deleted.

@feiskyer this PR does not change the original logic, this PR make vmss list operation as serial by using lock, and then it would take second look in vmss cache, that would improve the cache hit ratio when there are large mount of vmss list operations in parallel.

feiskyer (approved):
/lgtm

feiskyer:
> @feiskyer this PR does not change the original logic, this PR make vmss list operation as serial by using lock, and then it would take second look in vmss cache, that would improve the cache hit ratio when there are large mount of vmss list operations in parallel.

Reasonable. LGTM.

PR #824, opened 2021-09-30
chore: support multi-arch cloud-node-manager image

mainred:
#### What type of PR is this?
/kind feature

changes included in this PR:
- support multi-arch cloud-node-manager image
- adjust the Makefile target to be more user-friendly

Currently, we introduce only multi-arch cloud-node-manager images as required for AKS arm64 support for cluster nodes, and that will be pretty handy when cloud-controller-manager is also required in the future.

Fixes #

```release-note
chore: support multi-arch cloud-node-manager image
```

```docs

```


mainred:
@feiskyer , part of the targets has been verified through
https://dev.azure.com/AzureContainerUpstream/Kubernetes/_build/results?buildId=28240&view=results

andyzhangx (commented):
/ok-to-test

Makefile
```diff
@@ -104,56 +107,61 @@ build-ccm-image: docker-pull-prerequisites ## Build controller-manager image.
...
 .PHONY: push-images
-push-images: push-ccm-image push-node-image ## Push all images.
+push-images: push-ccm-image push-all-node-images ## Push all images.
+
+.PHONY: push
+push: push-ccm-image push-all-node-images push-all-windows-node-images ## Push all images.
 
 .PHONY: push-node-manager-manifest
-push-node-manager-manifest: ## Create and push a manifest list containing all the Windows and Linux images.
+push-node-manager-manifest: push-all-node-images push-all-windows-node-images ## Create and push a manifest list containing all the Windows and Linux images.
```
mainred, line 160: I did not manage to test push-node-manager-manifest as building windows image raised the following error
```
error: failed to solve: rpc error: code = Unknown desc = operating system is not supported
```

But test after removing windows related code works passed.

Makefile
```diff
@@ -18,6 +18,9 @@ SHELL=/bin/bash -o pipefail
 BIN_DIR=bin
 PKG_CONFIG=.pkg_config
 
+ARCH ?= amd64
+LINUX_ARCHS = amd64 arm arm64 ppc64le s390x
```
feiskyer, line 22: could you also add arm/arm64 for Windows images?

Makefile
feiskyer: Could you try adding the following lines to install qemu as part of the build?

```
docker run --privileged --rm tonistiigi/binfmt --uninstall qemu-aarch64
docker run --rm --privileged tonistiigi/binfmt --install all
```

PR #929, opened 2021-12-08
fix: detach disk should return error when throttled

andyzhangx:
#### What type of PR is this?
/kind bug

fix: detach disk should return error when throttled

Original case is that when there is detach disk batch operation, the rest detach disk(with empty operation) may return success even it's throttled.

Fixes #

```release-note
fix: detach disk should return error when throttled
```

```docs
fix: detach disk should return error when throttled
```


ZeroMagic (approved):
/lgtm

PR #955, opened 2021-12-29
chore: add vnetRG, vnetName, subnetName in account settings

andyzhangx:
#### What type of PR is this?
/kind cleanup

chore: add vnetRG, vnetName, subnetName in account settings
these 3 settings could be read from `azure.json`, while in some k8s clusters, e.g. ARO, these settings are not defined, and moreover, there are more than 1 subnet in one cluster, in such scenario, we should allow users to define those settings in storage class, one cloud config file could not define anything, and if it's not defined in storage class, still read default settings from cloud config file.

Fixes #

```release-note
chore: add vnetRG, vnetName, subnetName in account settings
```

```docs
chore: add vnetRG, vnetName, subnetName in account settings
```


ZeroMagic (approved):
/lgtm

PR #1156, opened 2022-02-17
Fix: ensure lb config untouched if config is not changed

MartinForReal:
#### What type of PR is this?

/kind bug

use patch request to update wanted changes and left other field untouched

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/azureclients/loadbalancerclient/azure_loadbalancerclient.go
+++ b/pkg/azureclients/loadbalancerclient/azure_loadbalancerclient.go
@@ -256,29 +256,42 @@ func (c *Client) createOrUpdateLB(ctx context.Context, resourceGroupName string,
 		"Microsoft.Network/loadBalancers",
 		loadBalancerName,
 	)
-	decorators := []autorest.PrepareDecorator{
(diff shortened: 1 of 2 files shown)


pkg/azureclients/loadbalancerclient/azure_loadbalancerclient.go
```diff
@@ -256,29 +256,42 @@ func (c *Client) createOrUpdateLB(ctx context.Context, resourceGroupName string,
 		"Microsoft.Network/loadBalancers",
 		loadBalancerName,
 	)
-	decorators := []autorest.PrepareDecorator{
-		autorest.WithPathParameters("{resourceID}", map[string]interface{}{"resourceID": resourceID}),
-		autorest.WithJSON(parameters),
-	}
-	if etag != "" {
-		decorators = append(decorators, autorest.WithHeader("If-Match", autorest.String(etag)))
+	resp, rerr := c.armClient.GetResource(ctx, resourceID, "")
```
feiskyer, line 259: could we avoid querying requests in the write API? Generally, the GET requests should be controlled by the reconciling loop and hence caches could be used to reduce number of API calls.

MartinForReal:
LB rest api doesn't support patch request. close it now.

PR #1213, opened 2022-03-01
Chores: remove stale LB info from local cache if lb is updated by resource provider

MartinForReal:
#### What type of PR is this?

/kind bug

Fixes #

```release-note
NONE
```

```docs

```


pkg/provider/azure_loadbalancer.go
```diff
@@ -1384,6 +1385,7 @@ func (az *Cloud) reconcileLoadBalancer(clusterName string, service *v1.Service,
 		}
 		if changed {
 			dirtyLb = true
+			_ = az.lbCache.Delete(*lb.Name)
```
feiskyer, line 1388: The LB is not updated yet here (it would be updated after `CreateOrUpdateLB()` is called), hence caches shouldn't be refreshed here and below.

PR #1227, opened 2022-03-04
Set LegacyServiceAccountTokenNoAutoGeneration to false for master/1.24 aks-engine configs

marosset:
…

Signed-off-by: Mark Rossetti <marosset@microsoft.com>

#### What type of PR is this?
/kind bug

Sets `LegacyServiceAccountTokenNoAutoGeneration=false` for controller-manager feature gates as a mitigation to preserve previous SA token generation behaviors for aks-engine.

https://github.com/kubernetes/enhancements/pull/2800/files introduced this feature gate and set true-by-default.

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/tests/k8s-azure/manifest/linux-dual-stack-iptables-kubenet.json
+++ b/tests/k8s-azure/manifest/linux-dual-stack-iptables-kubenet.json
@@ -24,7 +24,7 @@
                     "--max-pods": "110"
                 },
                 "controllerManagerConfig": {
-                    "--feature-gates": "IPv6DualStack=true"
+                    "--feature-gates": "IPv6DualStack=true,LegacyServiceAccountTokenNoAutoGeneration=false"
                 },
                 "addons": [
                     {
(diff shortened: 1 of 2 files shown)


marosset:
We made the same updates to aks-engine templates used in Windows tests

https://github.com/kubernetes-sigs/windows-testing/pull/312 and https://github.com/kubernetes-sigs/windows-testing/pull/313

aramase (commented):
/lgtm

marosset:
Let me update the PR jobs too

marosset:
> Let me update the PR jobs too

Hmm it looks like the PR jobs are using https://raw.githubusercontent.com/kubernetes-sigs/cloud-provider-azure/master/tests/k8s-azure/manifest/linux.json which is shared between different versions of K8s.
Because of this I cannot add the new feature gate here.

I think we either need to update aks-engine to add this featuregate for v1.24-alpha.3+ or break up the configs.
I'm leaning towards updating aks-engine.

--Mark

PR #1356, opened 2022-03-25
fix: disk attach/detach failure when operation is preempted

andyzhangx:
#### What type of PR is this?
/kind bug
fix: always check result code even returns http 200 success in VM update

There are cases that vm update returns `200` httpStatusCode, while it actually failed with `OperationPreempted`, vm update operation should return failure in such condition. Original code only checks http response which is incorrect.

```
resultType httpStatusCode resultCode	                         errorDetails
1	          200	            OperationPreempted		Operation execution has been preempted by a more recent operation.
```

Fixes #

```release-note
fix: disk attach/detach failure when operation is preempted
```

```docs
fix: disk attach/detach failure when operation is preempted
```


feiskyer (commented):
/lgtm
/approve

PR #1396, opened 2022-04-01
[release-1.0]fix: fallback to tcp if appprotocol is not supported

MartinForReal:
#### What type of PR is this?
/kind bug
/kind regression

Fall back to tcp if user assigned probe protocol is not supported.

Fixes #

```release-note
The cloud provider will create TCP probe rule if the probe protocol is not supported. 
```

```docs

```


feiskyer (commented):
/lgtm
/approve

PR #1300, opened 2022-03-21
Refactor store and use sync.map as data store

MartinForReal:
Signed-off-by: MartinForReal <fanshangxiang@gmail.com>

#### What type of PR is this?

/kind feature

#### What this PR does / why we need it:
remove k/client-go/tools/cache dependency and provide a more performant cache.

#### Which issue(s) this PR fixes:

Fixes #

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

#### Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.:

```docs

```

PR #1402, opened 2022-04-02
Rollback pr #1300 Refactor store and use sync.map as data store

MartinForReal:
#### What type of PR is this?
/kind bug

cocurrent get requests are generated if sync.map is used. 

Fixes #

As @jwtty @nilo19 mentioned, The code change may introduce a regression bug.

update entry logic doesn't fulfill the requirement documented here 
More than one routine will update cache store at the same time.

https://github.com/kubernetes-sigs/cloud-provider-azure/blob/4f83f549efd056c3ed9cc5cbacab2915c6374b6a/pkg/cache/azure_cache.go#L144-L147

The original code had held a Mutex lock before entered into update section.

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/cache/azure_cache.go
+++ b/pkg/cache/azure_cache.go
@@ -20,6 +20,8 @@ import (
 	"fmt"
 	"sync"
 	"time"
+
+	"k8s.io/client-go/tools/cache"
 )
 
 // AzureCacheReadType defines the read type for cache data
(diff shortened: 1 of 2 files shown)


andyzhangx:
@MartinForReal why roll back the original PR? could you provide more details?

MartinForReal:
I've updated description. FYI @andyzhangx I also raised another pr to fix it. But it is not validated in a long running process. Both pr is ok. Because here we only have one worker which processed svc update event. But These fixes are essential to other project which depends on this library.

andyzhangx (approved):
/lgtm

PR #1491, opened 2022-04-14
Add DeletePEConnection to plsClient

jwtty:
#### What type of PR is this?
/kind feature

When deleting a PLS, need to delete all existing PE connections first. So add this client interface.

Fixes #

```release-note
Add azure private link service client interface to delete a private endpoint connection.
```

```docs

```

### DIFF
--- a/pkg/azureclients/privatelinkserviceclient/azure_privatelinkserviceclient_test.go
+++ b/pkg/azureclients/privatelinkserviceclient/azure_privatelinkserviceclient_test.go
@@ -311,11 +311,49 @@ func TestDelete(t *testing.T) {
 	}
 }
 
+func TestDeletePEConnection(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+
+	tests := []struct {
+		description  string
+		armClientErr *retry.Error
+		expectedErr  *retry.Error
+	}{
+		{
+			description:  "Delete should report the throttling error",
+			armClientErr: &retry.Error{HTTPStatusCode: http.StatusTooManyRequests},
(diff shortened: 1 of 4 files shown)


feiskyer (approved):
/lgtm

pkg/azureclients/privatelinkserviceclient/azure_privatelinkserviceclient_test.go
```diff
@@ -319,6 +350,13 @@ func getTestPrivateLinkService(name string) network.PrivateLinkService {
 	}
 }
 
+func getTestPrivateEndpointConnection(PLSName string, PEConnName string) network.PrivateEndpointConnection {
+	return network.PrivateEndpointConnection{
+		ID:   to.StringPtr(fmt.Sprintf("/subscriptions/subscriptionID/resourceGroups/rg/providers/"+PLSResourceType+"/%s/%s/%s", PLSName, PEConnResourceType, PEConnName)),
```
lzhecheng, line 355: Why not like this? Current one seems a bit strange :confused:
```suggestion
		ID:   to.StringPtr(fmt.Sprintf("/subscriptions/subscriptionID/resourceGroups/rg/providers/%s/%s/%s/%s", PLSResourceType, PLSName, PEConnResourceType, PEConnName)),
```

jwtty: I was following the above `getTestPrivateLinkService` function. But I don't see any problem changing it.

lzhecheng: I think all of they can be changed.

PR #1500, opened 2022-04-15
Reduce listPIP calls

jwtty:
#### What type of PR is this?
/kind feature

In current code, `ListPIP` is called every time in `determinePublicIPName()` function when service specifies loadBalancerIP. This function can be called in multiple loops when reconciling loadBalancer and may cause throttling. 

Please find origin issue #1136 

Fixes #

```release-note
Reduce ListPublicIPAddress call count by avoiding each call in loops
```

```docs
```


feiskyer (commented):
/lgtm

feiskyer (approved):

PR #1508, opened 2022-04-18
Add support for specifying probe protocol / probe port via annotation per service port

vsabella:
Signed-off-by: Vito Sabella <vito@axon.com>


pkg/provider/azure_loadbalancer.go
```diff
@@ -1928,33 +1928,85 @@ func (az *Cloud) buildHealthProbeRulesForPort(annotations map[string]string, por
...
+		return nil, fmt.Errorf("failed to parse annotation %s: %w", consts.BuildHealthProbeAnnotationKeyForPort(port.Port, consts.HealthProbeParamsProtocol), err)
+	}
+
+	// 2. If not specified, look up from AppProtocol
+	// Note - this order is to remain compatible with previous versions
+	if protocol == nil {
+		protocol = port.AppProtocol
+	}
+
+	// 3. If protocol is still nil, check the global annotation
```
vsabella, line 1952: @MartinForReal I don't think this order makes sense. I would imagine the order we want is:

1. Local Port Annotation
2. Global Annotation
3. appProtocol in port
4. Default of TCP

However it seems that would be a breaking change. In the previous code it only applies the global annotation if appProtocol is not set. We want to set appProtocol for lots of reasons not just health checking.

Ideally it would be:
1. per-port annotation
2. global annotation
3. appProtocol
4. default (TCP)

If we agree to make the breaking change then I'll make a second PR to allow us to globally set the health probe protocol and probe port... Otherwise its fine to leave this as is.

MartinForReal: IMHO, I think the proper order would be:
Local Port Annotation
appProtocol in port
Global Annotation (option is removed from doc and code is not remove for backward compatibility)
Default of TCP

MartinForReal (commented):
We may need to refactor this part because if we enabled port/protocol customization, there will be duplicated probe rules. we may need to check if new probe rule already exists in probe rule sets.

pkg/provider/azure_loadbalancer.go
vsabella: That sounds good to me especially as the global annotation is deprecated :)

vsabella:
> We may need to refactor this part because if we enabled port/protocol customization, there will be duplicated probe rules. we may need to check if new probe rule already exists in probe rule sets.(Line 2097 in azure loadbalancer.go )

I'm not sure I get under what circumstances this happens - is it when we are set to externalTrafficPolicy: Local ? 

If you can give me an example i'm happy to add some tests to check it out.

MartinForReal:
In the Istio scenario, A customized probe rule whose target port is 150xx will be generated for port 80.  And a probe rule whose target port is 150xx will be generated for port 150xx (default provider behavior). And these two rules may have different configurations (probe protocol is different). I think we can avoid this rule conflict by changing the probe rule names.

vsabella:
> In the Istio scenario, A customized probe rule whose target port is 150xx will be generated for port 80. And a probe rule whose target port is 150xx will be generated for port 150xx (default provider behavior). And these two rules may have different configurations (probe protocol is different). I think we can avoid this rule conflict by changing the probe rule names.

Ah yes, I see your point. Is that really a problem though other than just being duplicated? Like a capacity or performance issue with LB

We could just make the key the probe configuration vs the backend configuration as it is now. That's what you see with pod status monitor / externalTrafficPolicy: Local

MartinForReal:
> 

ok

vsabella:
@MartinForReal Let me write up a spec for the change in rule generation to rule out duplicates. I'll link a wiki page here we can comment on

I'll ping you in a day or so once we have time to have something to review

vsabella:
> @MartinForReal Let me write up a spec for the change in rule generation to rule out duplicates. I'll link a wiki page here we can comment on
> 
> I'll ping you in a day or so once we have time to have something to review

Still working on it - had to take a few days off but I'll circulate the draft spec soon.

PR #1701, opened 2022-05-09
chore: remove GA IPv6DualStack feature-gate

jackfrancis:
#### What type of PR is this?

/kind cleanup

This PR removes references to the `IPv6DualStack` feature-gate as it is now locked to true and GA in upstream Kubernetes.

See:

- https://github.com/kubernetes/kubernetes/pull/109435

This change will break any future v0.7 releases of cloud-provider-azure because the feature-flag was not default to true until 1.21.0, so we'll have to be careful when we cherry-pick changes there. I observe that we are still releasing that version:

- https://github.com/kubernetes-sigs/cloud-provider-azure/releases/tag/v0.7.21

Fixes #

```release-note
remove GA IPv6DualStack feature-gate
```

```docs

```

### DIFF
--- a/pkg/provider/azure.go
+++ b/pkg/provider/azure.go
@@ -34,7 +34,6 @@ import (
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
-	utilfeature "k8s.io/apiserver/pkg/util/feature"
 	"k8s.io/client-go/informers"
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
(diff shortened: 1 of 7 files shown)


pkg/provider/azure.go
```diff
@@ -374,7 +373,7 @@ func NewCloud(configReader io.Reader, callFromCCM bool) (cloudprovider.Interface
 	if err != nil {
 		return nil, err
 	}
-	az.ipv6DualStackEnabled = utilfeature.DefaultFeatureGate.Enabled(consts.IPv6DualStack)
+	az.ipv6DualStackEnabled = true
```
CecileRobertMichon, line 376: Should we go one step further and deprecate/remove that field from Cloud config?

jackfrancis: This should be feasible as it doesn't appear to be a user-exposed configuration vector (we always relied upon the k8s feature-gate interface).

@feiskyer @lzhecheng @nilo19 thoughts?

feiskyer:
@jackfrancis could we only remove this in maser branch (to keep same pace as Kubernetes 1.25)?

pkg/provider/azure.go
feiskyer: yes, it could be removed  (this is not part of cloud config file).

PR #1745, opened 2022-05-19
helm: update tolerations and nodeSelectors

jackfrancis:
#### What type of PR is this?

/kind feature

This PR updates node tolerations and nodeSelectors in the helm chart so that the implementation works with 1.24.

Also removed an unnecessary toleration for the Windows cloud-node-manager deployment (Windows control plane nodes are not supported AFAIK).

Fixes #

```release-note
NONE
```

```docs

```


helm/cloud-provider-azure/templates/cloud-provider-azure.yaml
```diff
@@ -430,10 +438,6 @@ spec:
       tolerations:
       - key: CriticalAddonsOnly
         operator: Exists
-      - key: node-role.kubernetes.io/master
```
jackfrancis, line 433: @marosset @jsturtevant can you confirm that this control plane toleration is not needed for the Windows cloud-node-manager pod (i.e., there is no such thing as a "Windows control plane node" in the k8s ecosystem)

jsturtevant: That's correct, the windows `nodeSelector` above on line 437 is enough to land these on windows.

jackfrancis:
@lzhecheng these updates are needed to enable functionality on 1.24 clusters

PR #1863, opened 2022-06-17
fix: remove auto approval restriction

### LINKED ISSUE #1859: AKS Private Link Service requires global visbility
#### What happened:
Creating a private link service as [follows](https://github.com/kubernetes-sigs/cloud-provider-azure/issues/872):
```
apiVersion: v1
kind: Service
metadata:
  name: my-app
  namespace: my-namespace
  annotations:
    service.beta.kubernetes.io/azure-load-balancer-internal: "true"
    service.beta.kubernetes.io/azure-pls-create: "true"
    # Does not work without: service.beta.kubernetes.io/azure-pls-visibility: "*"
    # service.beta.kubernetes.io/azure-pls-visibility: "<UUID>" does not work either
    service.beta.kubernetes.io/azure-pls-auto-approval: "<UUID>"
spec:
  type: LoadBalancer
  ports:
  - name: http
    port: 8080
    protocol: TCP
    targetPort: 8080
  selector:
    app: my-app
```

The auto approval does not apply without pls-visibility set to *. The following controller event is surfaced: `reconcilePLSVisibility: autoApproval only takes effect when visibility is set to "*"`

I should be able to specify visibility and approval to the same subscription, removing the need for global visibility. This capability is present in the console: 
Screen Shot 2022-06-16 at 1 30 30 PM

```
service.beta.kubernetes.io/azure-pls-visibility: *
service.beta.kubernetes.io/azure-pls-auto-approval: "<UUID>"
``` 
is the third option and not ideal.

Apply service definition above.

N/A

#### Environment:
AKS

- Kubernetes version (use `kubectl version`): AKS 1.22.4
- Install tools: N/A
- Others:

jwtty:
#### What type of PR is this?
/kind bug

Remove redundant restriction on pls autoApproval and visibility.
User can specify a list of subscriptions for visibility and a subset of this list for autoApproval.

Fixes #1859 

```release-note
Remove redundant restriction on pls autoApproval and visibility.
User can specify a list of subscriptions for visibility (e.g. "sub1 sub2") and a subset of this list for autoApproval (e.g. "sub1").
```

```docs

```

### DIFF
--- a/pkg/provider/azure_privatelinkservice.go
+++ b/pkg/provider/azure_privatelinkservice.go
@@ -451,11 +451,8 @@ func reconcilePLSVisibility(
 	service *v1.Service,
 ) (bool, error) {
 	changed := false
-	visibilitySubs, anyoneCanView := getPLSVisibility(service)
+	visibilitySubs, _ := getPLSVisibility(service)
 	autoApprovalSubs := getPLSAutoApproval(service)
-	if !anyoneCanView && len(autoApprovalSubs) > 0 {
-		return false, fmt.Errorf("reconcilePLSVisibility: autoApproval only takes effect when visibility is set to \"*\"")
-	}
 
 	if existingPLS.Visibility == nil || existingPLS.Visibility.Subscriptions == nil {
 		if len(visibilitySubs) != 0 {
--- a/pkg/provider/azure_privatelinkservice_test.go
+++ b/pkg/provider/azure_privatelinkservice_test.go
@@ -1075,16 +1075,6 @@ func TestReconcilePLSVisibility(t *testing.T) {
 		assert.NoError(t, err)
 	})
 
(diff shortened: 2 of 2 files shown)


pkg/provider/azure_privatelinkservice.go
```diff
@@ -451,11 +451,8 @@ func reconcilePLSVisibility(
 	service *v1.Service,
 ) (bool, error) {
 	changed := false
-	visibilitySubs, anyoneCanView := getPLSVisibility(service)
+	visibilitySubs, _ := getPLSVisibility(service)
```
lzhecheng, line 454: Since it is the only place to call `getPLSVisibility(), why not just return `visibilityList`?

jwtty: I want to keep it in case it's needed in the future.

PR #2060, opened 2022-07-20
fix: don't set vm entry as nil in cache when vm is in Deleting state to avoid VMSSList throttling

andyzhangx:
#### What type of PR is this?
/kind bug

fix: don't set vm entry as nil in cache when vm is in Deleting state to avoid VMSSList throttling
When VM is in Deleting state, original behavior is setting as nil, and then with this PR(https://github.com/kubernetes-sigs/cloud-provider-azure/pull/1578), Deleting state VM will be not found in cache, then it would trigger VMSSList again, found VM is still in Deleting state, return nil, trigger VMSSList again and again, finally get VMSSList throttling.

Fixes #

```release-note
fix: don't set vm entry as nil in cache when vm is in Deleting state to avoid VMSSList throttling
```

```docs
fix: don't set vm entry as nil in cache when vm is in Deleting state to avoid VMSSList throttling
```

### DIFF
--- a/pkg/provider/azure_vmss_cache.go
+++ b/pkg/provider/azure_vmss_cache.go
@@ -223,11 +223,9 @@ func (ss *ScaleSet) newVMSSVirtualMachinesCache(resourceGroupName, vmssName, cac
 				virtualMachine: &vm,
 				lastUpdate:     time.Now().UTC(),
 			}
-			// set cache entry to nil when the VM is under deleting.
 			if vm.VirtualMachineScaleSetVMProperties != nil &&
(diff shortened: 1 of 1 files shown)


pkg/provider/azure_vmss_cache.go
```diff
@@ -226,8 +226,7 @@ func (ss *ScaleSet) newVMSSVirtualMachinesCache(resourceGroupName, vmssName, cac
 			// set cache entry to nil when the VM is under deleting.
```
MartinForReal, line 226: remove the comments as well?

andyzhangx: fixed.

PR #2161, opened 2022-08-10
Revert "fix: don't set vm entry as nil in cache when vm is in Deleting state to avoid VMSSList throttling"

andyzhangx:
Reverts kubernetes-sigs/cloud-provider-azure#2060
This PR returns InstanceNotFound when vm in cache is nil, this PR would fix following issue in another way:
When VM is in Deleting state, original behavior is setting as nil, and then with this PR(https://github.com/kubernetes-sigs/cloud-provider-azure/pull/1578), Deleting state VM will be not found in cache, then it would trigger VMSSList again, found VM is still in Deleting state, return nil, trigger VMSSList again and again, finally get VMSSList throttling.

```release-note
none
```

```docs
none
```


feiskyer (approved):
/lgtm

Thanks for the fix.

PR #2172, opened 2022-08-15
make ConvertResourceGroupNameToLower exportable

jackfrancis:
#### What type of PR is this?

/kind cleanup

This PR converts the existing `convertResourceGroupNameToLower` func in the `pkg/provider` library to be exportable: `ConvertResourceGroupNameToLower`.

The reason for this is that this string transformation logic is intrinsic to how cloud-provider-azure identifies a node's `providerID` property, and capz needs to be able to similarly recognize node `providerID` data. If this func is exportable that means that capz can simply import it and use it, to ensure that it will always be up-to-date with how cloud-provider-azure performs this work.

Refer to this capz PR which is manually duplicating this logic:

https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2536

Once this PR lands, we can convert the above capz surface area to simply call this exported cloud-provider-azure function.

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/provider/azure_standard.go
+++ b/pkg/provider/azure_standard.go
@@ -534,9 +534,9 @@ func (as *availabilitySet) GetInstanceIDByNodeName(name string) (string, error) 
 	}
 
 	resourceID := *machine.ID
-	convertedResourceID, err := convertResourceGroupNameToLower(resourceID)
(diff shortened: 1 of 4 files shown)


feiskyer (approved):
thanks for addressing this

PR #2288, opened 2022-09-08
VMSS Flex Support: Embed FlexScaleSet logic into existing CCM execution flow

zmyzheng:
#### What type of PR is this?
/kind feature

This PR embeds FlexScaleSet logic into existing CCM execution flow. Key points of the change:

1. Change AvailabilitySetNodesCache to NonUniformNodesCache
2. Change isNodeManagedByAvailabilitySet to getVMManagementType
3. Integrate FlexScaleSet logic into existing CCM execution flow
- vmType == standard: only AvSet VMs or standalone VMs (no change)
- vmType == vmss && DisableAvailabilitySetNodes == true && EnableVmssFlexNodes == false: only VMSS Uniform VMs (for current AKS with Uniform Scale Set)
- vmType == vmss && DisableAvailabilitySetNodes == false && EnableVmssFlexNodes == true: mix mode: AvSet + Uniform + Flex

Refer to #639 

```release-note
NONE
```

```docs

```


pkg/provider/azure_vmss_cache.go
```diff
@@ -294,24 +313,36 @@ func (ss *ScaleSet) deleteCacheForNode(nodeName string) error {
 	return nil
 }
 
-func (ss *ScaleSet) newAvailabilitySetNodesCache() (*azcache.TimedCache, error) {
+func (ss *ScaleSet) newNonVmssUniformNodesCache() (*azcache.TimedCache, error) {
```
andyzhangx, line 316: what does this func name(`newNonVmssUniformNodesCache`) mean?

andyzhangx: it's vmss flex + avset, why vmss uniform is excluded? is `vmss uniform` logic more similar to vmss flex?

pkg/provider/azure_vmss_cache.go
```diff
@@ -294,24 +313,36 @@ func (ss *ScaleSet) deleteCacheForNode(nodeName string) error {
...
-					vmNames.Insert(to.String(vm.Name))
-					vmList = append(vmList, vm)
+				if vm.OsProfile != nil && vm.OsProfile.ComputerName != nil {
+					if vm.VirtualMachineScaleSet != nil {
+						vmssFlexVMNodeNames.Insert(strings.ToLower(to.String(vm.OsProfile.ComputerName)))
+						if vm.ID != nil {
+							vmssFlexVMProviderIDs.Insert(ss.ProviderName() + "://" + to.String(vm.ID))
+						}
+					} else {
+						avSetVMNodeNames.Insert(strings.ToLower(to.String(vm.OsProfile.ComputerName)))
```
andyzhangx, line 341: it's changing avset cache logic already, originally it's `to.String(vm.Name)`, now it's `strings.ToLower(to.String(vm.OsProfile.ComputerName))`

do you have example what's the diff between these two values, and why use `strings.ToLower`

andyzhangx (commented):
/hold
hold a while since it's already changing the existing logic, need to run csi driver e2e test before merge.

pkg/provider/azure_controller_common.go
```diff
@@ -114,7 +114,7 @@ type ExtendedLocation struct {
 // getNodeVMSet gets the VMSet interface based on config.VMType and the real virtual machine type.
 func (c *controllerCommon) getNodeVMSet(nodeName types.NodeName, crt azcache.AzureCacheReadType) (VMSet, error) {
 	// 1. vmType is standard, return cloud.VMSet directly.
-	if c.cloud.VMType == consts.VMTypeStandard {
+	if c.cloud.VMType == consts.VMTypeStandard || c.cloud.VMType == consts.VMTypeVmssFlex {
```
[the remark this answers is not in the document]
zmyzheng: No, VMSS Flex is using a different implementation called FlexScaleSet. 

If you take a look at the below logic:
image

https://github.com/kubernetes-sigs/cloud-provider-azure/pull/2288/files#diff-5f537390df5562c47fe9c7f698fc4ae0ee6a08a51272c344e0d78d21cf0cf8dbR616

you will find c.cloud.VMSet has been set to the correct implementation based on the vm type. So no matter c.cloud.VMType is standard of vmssflex, it should always use c.cloud.VMSet

pkg/provider/azure_vmss_cache.go
```diff
@@ -294,24 +313,36 @@ func (ss *ScaleSet) deleteCacheForNode(nodeName string) error {
 	return nil
 }
 
-func (ss *ScaleSet) newAvailabilitySetNodesCache() (*azcache.TimedCache, error) {
+func (ss *ScaleSet) newNonVmssUniformNodesCache() (*azcache.TimedCache, error) {
 	getter := func(key string) (interface{}, error) {
-		vmNames := sets.NewString()
+		klog.V(2).Infof("calling getter function of newNonVmssUniformNodesCache with key: %s", key)
```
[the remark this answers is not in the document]
zmyzheng: Done

pkg/provider/azure_vmss_cache.go
zmyzheng: newNonVmssUniformNodesCache contains vmss flex vm and avset vms.

The cache newNonVmssUniformNodesCache is used to replace the old availabilitySetNodesCache to distinguish the node type based on the nodeName.

One Key difference between vmssflex and vmss uniform is: in Flex, each VM is a top-level azure resource; while in Uniform, each VM is a sub-resource of VMSS. This means, the VM List API can only return VMSS Flex VMs, but it cannot return VMSS Uniform VMs. We utilize this feature to distinguish Flex VM and Uniform VM. The logic is:

- Use VM List API to retrieve all Visible VMs inside the resource group. The visible VMs include Flex VMs and AvSet VMs.
- for each returned VM, check whether it has a VMSS ID. If it has, this means it is a Flex VM. If not, this means it is an AvSet VM
- When checking a node type by NodeName, we first check if this node is inside the nonVmssUniformNodesCache. If not, it means it is a Uniform VM. If it is inside the nonVmssUniformNodesCache, then further check if the node has VMSS ID or not. If it has Vmss ID, it means it is FLex VM. If not, it is AvSet VM

pkg/provider/azure_vmss_cache.go
zmyzheng: After testing different types of VMs (Flex, Uniform, AvSet), I found out the naming conventions of nodeName. More specifically:
- No matter which node type it is, the nodeName is always equal to lower case of OS HostName (namely, strings.ToLower(to.String(vm.OsProfile.ComputerName))). This is because the kubelet report the nodeName using OS HostName when a new node just joined the cluster.
- When using AvSet VM, the OS hostname is always the same as VM name. This is why using vm name for avSet vm also works.

So, to unify them, we should always use OS HostName.

pkg/provider/azure_vmss_cache.go
```diff
@@ -294,24 +313,36 @@ func (ss *ScaleSet) deleteCacheForNode(nodeName string) error {
 	return nil
 }
 
-func (ss *ScaleSet) newAvailabilitySetNodesCache() (*azcache.TimedCache, error) {
+func (ss *ScaleSet) newNonVmssUniformNodesCache() (*azcache.TimedCache, error) {
 	getter := func(key string) (interface{}, error) {
-		vmNames := sets.NewString()
+		klog.V(6).Infof("refresh the cache of NonVmssUniformNodesCache")
+		vmssFlexVMNodeNames := sets.NewString()
```
andyzhangx, line 319: `NonVmssUniformNodesCache` means it's not VMSS uniform cache? while current test failures are all on VMSS uniform nodes, so that's the issue?

andyzhangx:
in our failed csi driver test cases, `DisableAvailabilitySetNodes` is set as `false`, so the following case is broken?
```
For pure vmss uniform cluster, vm type should be set to vmss and DisableAvailabilitySetNodes should be set to true. (no change)
```

actually `DisableAvailabilitySetNodes` just means we don't want to list avset since we could make sure there are no avset in the cluster, it does not mean current cluster is vmss uniform or vmss flex.

```
I0914 06:21:45.179932       1 azuredisk.go:201] cloud: AzurePublicCloud, location: uksouth, rg: kubetest-zshepzqk, VMType: vmss, PrimaryScaleSetName: 2679k8s00, PrimaryAvailabilitySetName: , DisableAvailabilitySetNodes: false
```

pkg/provider/azure_vmss_cache.go
andyzhangx: and I found following logs in vmss uniform node pool cluster:
```
I0914 10:15:02.899048       1 azure_vmss_cache.go:388] Node k8s-agentpool-25348871-vmss000000 has joined the cluster since the last VM cache refresh in nonVmssUniformNodesEntry, refreshing the cache
```

The log is strange since one vmss uniform node is not found and then it tries to refresh nonVmssUniformNodesEntry

pkg/provider/azure_vmss_cache.go
```diff
@@ -322,43 +353,120 @@ func (ss *ScaleSet) newAvailabilitySetNodesCache() (*azcache.TimedCache, error)
...
+	return azcache.NewTimedcache(time.Duration(ss.Config.NonVmssUniformNodesCacheTTLInSeconds)*time.Second, getter)
 }
 
-func (ss *ScaleSet) isNodeManagedByAvailabilitySet(nodeName string, crt azcache.AzureCacheReadType) (bool, error) {
-	// Assume all nodes are managed by VMSS when DisableAvailabilitySetNodes is enabled.
+func (ss *ScaleSet) getVMManagementTypeByNodeName(nodeName string, crt azcache.AzureCacheReadType) (VMManagementType, error) {
 	if ss.DisableAvailabilitySetNodes {
-		klog.V(6).Infof("Assuming node %q is managed by VMSS since DisableAvailabilitySetNodes is set to true", nodeName)
-		return false, nil
+		return ManagedByVmssUniform, nil
```
andyzhangx, line 375: this is actually not correct, `DisableAvailabilitySetNodes` means current cluster does not have avset (save vmas list operation), while it may have vmss uniform or vmss flex.

pkg/provider/azure_vmss_cache.go
zmyzheng: Per discussion, this issue is resolved by using the correct OrchetrationMode value

pkg/provider/azure_vmss_cache.go
zmyzheng: There indeed is a sperate cache for VMSS Flex nodes: https://github.com/kubernetes-sigs/cloud-provider-azure/blob/master/pkg/provider/azure_vmssflex.go#L53

The NonVmssUniformNodesCache inside azure_vmss.go is only used to distinguish Flex, Uniform and AvSet nodes. VM perperties are not retrieved from NonVmssUniformNodesCache

zmyzheng:
> in our failed csi driver test cases, `DisableAvailabilitySetNodes` is set as `false`, so the following case is broken?
> 
> ```
> For pure vmss uniform cluster, vm type should be set to vmss and DisableAvailabilitySetNodes should be set to true. (no change)
> ```
> 
> actually `DisableAvailabilitySetNodes` just means we don't want to list avset since we could make sure there are no avset in the cluster, it does not mean current cluster is vmss uniform or vmss flex.
> 
> ```
> I0914 06:21:45.179932       1 azuredisk.go:201] cloud: AzurePublicCloud, location: uksouth, rg: kubetest-zshepzqk, VMType: vmss, PrimaryScaleSetName: 2679k8s00, PrimaryAvailabilitySetName: , DisableAvailabilitySetNodes: false
> ```

Per discussion, the root cause is the value of OrchestrationMode is empty. The lastest code has fixed this issue

pkg/provider/azure_vmss_cache.go
zmyzheng: Yes and No. The name DisableAvailabilitySetNodes is quite missleading. The starting point to have this flag is just to reduce API calls by not checking node types. Below is the detailed explanation:

- When CCM start to support AKS, all the nodes are VMSS Uniform nodes, so there is no need to distinguish node type at that time. 
- Later when CCM needs to support CAPZ, Uniform nodes and AvSet nodes are used together, so CCM added the availabilitySetNodesCache to distinguish them. More specifically, it added an availabilitySetNodesCache by using the VM List API to batch query all AvSet  (including standalone VMs) VM names in the resource group. After this, CCM does a check for the node type before each function execution. The side effect is, CCM needs to refresh the availabilitySetNodesCache periodically, which is an  additional cost. 
- Considering AKS only contains Uniform nodes, DisableAvailabilitySetNodes flag is added to bypass the node type check to reduce the unnecessary API call.
- Now, to support VMSS Flex, we should make the CCM to have backward compatilibity and does not affect the performance of existing execution. So, here if DisableAvailabilitySetNodes is set to true, we will not check the node type and assume all nodes are Uniform nodes, which is excactly the behavior before VMSS Flex is supported. In this way, the existing AKS cluster's performance using Uniform will not be affected.

pkg/provider/azure_vmss.go
```diff
@@ -1404,6 +1470,83 @@ func (ss *ScaleSet) EnsureHostsInPool(service *v1.Service, nodes []*v1.Node, bac
 	return nil
 }
 
+// EnsureHostsInPool ensures the given Node's primary IP configurations are
+// participating in the specified LoadBalancer Backend Pool.
+func (ss *ScaleSet) EnsureHostsInPool(service *v1.Service, nodes []*v1.Node, backendPoolID string, vmSetNameOfLB string) error {
+	if ss.DisableAvailabilitySetNodes {
```
andyzhangx, line 1476: @nilo19 can you take a look at this change?

pkg/provider/azure_vmss_cache.go
```diff
@@ -47,12 +53,23 @@ type vmssEntry struct {
...
+	clusterNodeNames      sets.String
 }
 
+type VMManagementType string
+
+const (
+	ManagedByVmssUniform VMManagementType = "ManagedByVmssUniform"
+	ManagedByVmssFlex    VMManagementType = "ManagedByVmssFlex"
+	ManagedByAvSet       VMManagementType = "ManagedByAvSet"
+	ManagedByUnknown     VMManagementType = "ManagedByUnknown"
```
andyzhangx, line 70: naming `Unknown` would be better IMO

pkg/provider/azure_vmss_cache.go
```diff
@@ -322,43 +353,120 @@ func (ss *ScaleSet) newAvailabilitySetNodesCache() (*azcache.TimedCache, error)
...
+	cachedVmssFlexVMs := cached.(nonVmssUniformNodesEntry).vmssFlexVMNodeNames
+
+	if cachedAvSetVMs.Has(nodeName) {
+		return ManagedByAvSet, nil
+	}
+	if cachedVmssFlexVMs.Has(nodeName) {
+		return ManagedByVmssFlex, nil
+	}
+	return ManagedByVmssUniform, nil
+
```
andyzhangx, line 405: remove empty line

pkg/provider/azure_vmss_cache.go
```diff
@@ -33,6 +34,11 @@ import (
 	"sigs.k8s.io/cloud-provider-azure/pkg/consts"
 )
 
+var (
+	// ErrorManagedByUnknown indicates an instance management type is unknown.
+	ErrorManagedByUnknown = errors.New("vm management type unknown")
```
andyzhangx, line 39: this var is not used?

pkg/provider/azure_vmss_cache.go
```diff
@@ -322,43 +353,120 @@ func (ss *ScaleSet) newAvailabilitySetNodesCache() (*azcache.TimedCache, error)
...
+	cachedAvSetVMs := cached.(nonVmssUniformNodesEntry).avSetVMNodeNames
+	cachedVmssFlexVMs := cached.(nonVmssUniformNodesEntry).vmssFlexVMNodeNames
+
+	if cachedAvSetVMs.Has(nodeName) {
+		return ManagedByAvSet, nil
+	}
+	if cachedVmssFlexVMs.Has(nodeName) {
+		return ManagedByVmssFlex, nil
+	}
+	return ManagedByVmssUniform, nil
```
andyzhangx, line 404: is there a way to check whether current nodeName is `ManagedByVmssUniform`, it's now just return `ManagedByVmssUniform` by default.

andyzhangx (commented):
```
The name DisableAvailabilitySetNodes is kind of misleading: it actually means disable both AvailabilitySet nodes and VMSS Flex nodes
```

I think there is problem for this since `DisableAvailabilitySetNodes` only means disable avset VM list now, while in the future, it would also disable vmss flex list, that may break in the future.

The motivation for adding `DisableAvailabilitySetNodes` flag is on AKS, avset and vmss cannot be enabled on one cluster, while for vmss uniform and vmss flex, both could be enabled on one cluster?

zmyzheng:
> ```
> The name DisableAvailabilitySetNodes is kind of misleading: it actually means disable both AvailabilitySet nodes and VMSS Flex nodes
> ```
> 
> I think there is problem for this since `DisableAvailabilitySetNodes` only means disable avset VM list now, while in the future, it would also disable vmss flex list, that may break in the future.
> 
> The motivation for adding `DisableAvailabilitySetNodes` flag is on AKS, avset and vmss cannot be enabled on one cluster, while for vmss uniform and vmss flex, both could be enabled on one cluster?

I will leave this question to @feiskyer for confirmation.

My main consideration here is backward compatibility. Before my code change, there is a way (set DisableAvailabilitySetNodes to true) to tell CCM all nodes are Uniform nodes and node type check can be bypassed to reduce the API call. After adding my code change, CCM should also have the ability to do so. This will make sure the latest CCM code can continue to be used in the old all Uniform node AKS cluster and the behavior and performance are exactly the same as before. I agree continuing to use the flag DisableAvailabilitySetNodes is a little misleading (it would be nice if it was named as something like "EnableAllUniformNodesCluster"), however, since it is already named in such as way, looks like we don't have better option to keep both backward compatibility and naming clarity.

pkg/provider/azure_vmss_cache.go
zmyzheng: Renamed it to "ManagedByUnknownVMSet"

pkg/provider/azure_vmss_cache.go
zmyzheng: Thanks for catching this. Just removed it.

pkg/provider/azure_vmss_cache.go
zmyzheng: The previous code is doing the same: as long as it is not in the cache, directly treat it as Uniform node: https://github.com/kubernetes-sigs/cloud-provider-azure/pull/2288/files?diff=split&w=0#diff-e6715e02b5aa5f8b693bf3e14dd99dd9a160f0f0f1481df96891af24fc3a5fffL340

pkg/provider/azure_vmss_cache.go
zmyzheng: done

andyzhangx:
> > ```
> > The name DisableAvailabilitySetNodes is kind of misleading: it actually means disable both AvailabilitySet nodes and VMSS Flex nodes
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I think there is problem for this since `DisableAvailabilitySetNodes` only means disable avset VM list now, while in the future, it would also disable vmss flex list, that may break in the future.
> > The motivation for adding `DisableAvailabilitySetNodes` flag is on AKS, avset and vmss cannot be enabled on one cluster, while for vmss uniform and vmss flex, both could be enabled on one cluster?
> 
> I will leave this question to @feiskyer for confirmation.
> 
> My main consideration here is backward compatibility. Before my code change, there is a way (set DisableAvailabilitySetNodes to true) to tell CCM all nodes are Uniform nodes and node type check can be bypassed to reduce the API call. After adding my code change, CCM should also have the ability to do so. This will make sure the latest CCM code can continue to be used in the old all Uniform node AKS cluster and the behavior and performance are exactly the same as before. I agree continuing to use the flag DisableAvailabilitySetNodes is a little misleading (it would be nice if it was named as something like "EnableAllUniformNodesCluster"), however, since it is already named in such as way, looks like we don't have better option to keep both backward compatibility and naming clarity.

@zmyzheng I think the main concern is whether it's supported to enable vmss uniform and vmss flex on one AKS cluster in the future? If yes, we should not use `DisableAvailabilitySetNodes` to disallow vmss flex list all. We could add new flags, e.g. `DisableVMSSUniformNodes`, `DisableVMSSFlexNodes` flags, that would be compatible, and more straigntforward.

zmyzheng:
> > > ```
> > > The name DisableAvailabilitySetNodes is kind of misleading: it actually means disable both AvailabilitySet nodes and VMSS Flex nodes
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > I think there is problem for this since `DisableAvailabilitySetNodes` only means disable avset VM list now, while in the future, it would also disable vmss flex list, that may break in the future.
> > > The motivation for adding `DisableAvailabilitySetNodes` flag is on AKS, avset and vmss cannot be enabled on one cluster, while for vmss uniform and vmss flex, both could be enabled on one cluster?
> > 
> > 
> > I will leave this question to @feiskyer for confirmation.
> > My main consideration here is backward compatibility. Before my code change, there is a way (set DisableAvailabilitySetNodes to true) to tell CCM all nodes are Uniform nodes and node type check can be bypassed to reduce the API call. After adding my code change, CCM should also have the ability to do so. This will make sure the latest CCM code can continue to be used in the old all Uniform node AKS cluster and the behavior and performance are exactly the same as before. I agree continuing to use the flag DisableAvailabilitySetNodes is a little misleading (it would be nice if it was named as something like "EnableAllUniformNodesCluster"), however, since it is already named in such as way, looks like we don't have better option to keep both backward compatibility and naming clarity.
> 
> @zmyzheng I think the main concern is whether it's supported to enable vmss uniform and vmss flex on one AKS cluster in the future? If yes, we should not use `DisableAvailabilitySetNodes` to disallow vmss flex list all. We could add new flags, e.g. `DisableVMSSUniformNodes`, `DisableVMSSFlexNodes` flags, that would be compatible, and more straigntforward.

I agree adding new flags is more straightforward. However, one conner case I can think of is, how do we handle the existing AKS clusters? The existing clusters only knows the "DisableAvailabilitySetNodes "flag and asumes all nodes are Uniform and skip node type check when "DisableAvailabilitySetNodes" is set to true. If we choose to add new flags, after existing AKS cluster upgrades to the new CCM, the behavior will change: it will start to check node type, which is unnecessary for those existing all Uniform node clusters. If this node type check sounds OK to you, then I'm OK to add new flags.

feiskyer:
> I think there is problem for this since DisableAvailabilitySetNodes only means disable avset VM list now, while in the future, it would also disable vmss flex list, that may break in the future.

Correct. DisableAvailabilitySetNodes is only for VMAS nodes, it should not be used for VMSS flex nodes.

> Add a new vm type "vmssflex" for the K8s cluster with all nodes from VMSS Flex.

Could you make the vmss flex supported via `vmss` vmType? Adding another mode would make it not be able to support different kinds of nodes together. A new `DisableVMSSFlexNodes` could be added to disable vmss flex in case flex nodes are not desired for the cluster (and vmss flex should be disabled by default on the alpha/beta stage).

> However, one conner case I can think of is, how do we handle the existing AKS clusters? The existing clusters only knows the "DisableAvailabilitySetNodes "flag and asumes all nodes are Uniform and skip node type check when "DisableAvailabilitySetNodes" is set to true.

Checking node type is ok. When the VMAS or VMSS Flex are disabled, our caches won't need to invoke the related APIs, hence reduce the chances of throttling issues (this is why DisableAvailabilitySetNodes=true when the cluster only want to use VMSS Uniform).

pkg/consts/consts.go
```diff
@@ -27,6 +27,8 @@ const (
 	VMTypeVMSS = "vmss"
 	// VMTypeStandard is the vmas vm type
 	VMTypeStandard = "standard"
+	// VMTypeVmssFlex is the vmssflex vm type
+	VMTypeVmssFlex = "vmssflex"
```
feiskyer, line 31: as commented, let's avoid adding a new "vmssflex" vmType.

pkg/provider/azure.go
```diff
@@ -610,7 +613,12 @@ func (az *Cloud) InitializeCloudFromConfig(config *Config, fromSecret, callFromC
 		az.MaximumLoadBalancerRuleCount = consts.MaximumLoadBalancerRuleCount
 	}
 
-	if strings.EqualFold(consts.VMTypeVMSS, az.Config.VMType) {
+	if strings.EqualFold(consts.VMTypeVmssFlex, az.Config.VMType) {
+		az.VMSet, err = newFlexScaleSet(az)
```
feiskyer, line 617: let's create a new `VMSSFlexSet` here for vmss flex when DisableVMSSFlexNodes==false?

feiskyer: then in the following controllers, you could invoke the VMSet implementation based on the real node type.

pkg/provider/azure_vmss_cache.go
```diff
@@ -294,24 +307,36 @@ func (ss *ScaleSet) deleteCacheForNode(nodeName string) error {
 	return nil
 }
 
-func (ss *ScaleSet) newAvailabilitySetNodesCache() (*azcache.TimedCache, error) {
+func (ss *ScaleSet) newNonVmssUniformNodesCache() (*azcache.TimedCache, error) {
```
feiskyer, line 310: Could you separate VMSSFlex caches from VMAS/Standalone VM caches for more clear?

pkg/consts/consts.go
zmyzheng: Done

pkg/provider/azure_vmss_cache.go
zmyzheng: nonVmssUniformNodesCache is only used for checking node type here

zmyzheng:
Per discussion with @feiskyer offline, below are the latest configurations:
image

pkg/provider/azure_vmss.go
```diff
@@ -852,15 +935,20 @@ func extractResourceGroupByVMSSNicID(nicID string) (string, error) {
 
 // GetPrimaryInterface gets machine primary network interface by node name and vmSet.
 func (ss *ScaleSet) GetPrimaryInterface(nodeName string) (network.Interface, error) {
-	managedByAS, err := ss.isNodeManagedByAvailabilitySet(nodeName, azcache.CacheReadTypeDefault)
```
feiskyer, line 855: why changing cache type to unsafe here?

pkg/provider/azure_vmss.go
```diff
@@ -74,9 +74,13 @@ type ScaleSet struct {
...
 
-	vmssCache                 *azcache.TimedCache
-	vmssVMCache               *sync.Map // [resourcegroup/vmssname]*azcache.TimedCache
-	availabilitySetNodesCache *azcache.TimedCache
+	flexScaleSet VMSet
+
+	vmssCache   *azcache.TimedCache
+	vmssVMCache *sync.Map // [resourcegroup/vmssname]*azcache.TimedCache
+
+	nonVmssUniformNodesCache *azcache.TimedCache
```
feiskyer, line 82: could you add a comment for the new cache?

pkg/provider/azure_vmss.go
```diff
@@ -74,9 +74,13 @@ type ScaleSet struct {
 	// this also allows for clusters with both VM and VMSS nodes.
 	availabilitySet VMSet
 
-	vmssCache                 *azcache.TimedCache
-	vmssVMCache               *sync.Map // [resourcegroup/vmssname]*azcache.TimedCache
-	availabilitySetNodesCache *azcache.TimedCache
+	flexScaleSet VMSet
```
feiskyer, line 77: could you add a comment when flexScaleSet would be used?

pkg/provider/azure_vmss.go
```diff
@@ -92,18 +96,22 @@ func newScaleSet(az *Cloud) (VMSet, error) {
...
 		vmssVMCache:     &sync.Map{},
 		lockMap:         newLockMap(),
 	}
 
-	if !ss.DisableAvailabilitySetNodes {
-		ss.availabilitySetNodesCache, err = ss.newAvailabilitySetNodesCache()
-		if err != nil {
-			return nil, err
-		}
+	ss.nonVmssUniformNodesCache, err = ss.newNonVmssUniformNodesCache()
```
feiskyer, line 112: I think we still need the old check: `if !ss.DisableAvailabilitySetNodes || ss.EnableVmssFlexNodes`

pkg/provider/azure_vmss.go
```diff
@@ -352,18 +370,29 @@ func (ss *ScaleSet) getVmssVMByInstanceID(resourceGroup, scaleSetName, instanceI
...
+		return ss.flexScaleSet.GetInstanceIDByNodeName(name)
+	}
 
 	vm, err := ss.getVmssVM(name, azcache.CacheReadTypeUnsafe)
 	if err != nil {
+		// special case: during scaling in, if the vm is deleted and nonVmssUniformNodesCache is refreshed,
+		// then getVMManagementTypeByNodeName will return ManagedByVmssUniform no matter what the actual managementType is.
+		// In this case, if it is actually a non vmss uniform node, return InstanceNotFound
+		if errors.Is(err, ErrorNotVmssInstance) {
+			return "", cloudprovider.InstanceNotFound
```
feiskyer, line 394: since the type has been checked above, why a node falls into this condition?

pkg/provider/azure_vmss_cache.go
```diff
@@ -47,12 +47,23 @@ type vmssEntry struct {
...
+	clusterNodeNames      sets.String
 }
 
+type VMManagementType string
+
+const (
+	ManagedByVmssUniform  VMManagementType = "ManagedByVmssUniform"
+	ManagedByVmssFlex     VMManagementType = "ManagedByVmssFlex"
+	ManagedByAvSet        VMManagementType = "ManagedByAvSet"
+	ManagedByUnknownVMSet VMManagementType = "ManagedByUnknownVMSet"
```
feiskyer, line 64: should we call this "ManagedByStandalone"?

pkg/provider/azure_vmss_cache.go
```diff
@@ -322,43 +347,119 @@ func (ss *ScaleSet) newAvailabilitySetNodesCache() (*azcache.TimedCache, error)
...
+	cachedVmssFlexVMProviderIDs := cached.(nonVmssUniformNodesEntry).vmssFlexVMProviderIDs
+	cachedAvSetVMProviderIDs := cached.(nonVmssUniformNodesEntry).avSetVMProviderIDs
+
+	if cachedAvSetVMProviderIDs.Has(providerID) {
+		return ManagedByAvSet, nil
+	}
+	if cachedVmssFlexVMProviderIDs.Has(providerID) {
+		return ManagedByVmssFlex, nil
+	}
+	return ManagedByUnknownVMSet, fmt.Errorf("getVMManagementTypeByProviderID : failed to check the providerID %s management type", providerID)
```
feiskyer, line 426: since standalone vm is also supported, I don't think we should return an error here. Could you recheck the changes in other places won't break standalone VM (they are simply reusing the same logic as VMAS, except with an empty vm set name)?

feiskyer (commented):
Have you run the tests on CSI repo again?

pkg/provider/azure_vmss.go
zmyzheng: A node management type (by uniform, flex or avset) cannot be changed after creation, so there is no need to fresh cache when checking it.

pkg/provider/azure_vmss.go
zmyzheng: Done

pkg/provider/azure_vmss.go
zmyzheng: Done

pkg/provider/azure_vmss.go
zmyzheng: May I know why do we need the check? Please note the initialization of nonVmssUniformNodesCache only defines the getter function of how to refresh the cache as well as the timeout, but it does not take action to refresh the cache. So, we don't need to worry about the extra api cost.

pkg/provider/azure_vmss.go
zmyzheng: This is based on a discovery of my testing. This applies to a special case:

Step 1: a scaling in operation deletes a node
Step 2: the nonVmssUniformNodesCache is refreshed by timeout

In the above case, the nonVmssUniformNodesCache does not contains the node name of this node  anymore so it assumes this is a uniform node. This is the case where it falls to this condition

pkg/provider/azure_vmss_cache.go
zmyzheng: No, both AvailabilitySet node and individual nodes will be distinguished as ManagedByAvSet, which is exactly the same as what is before this PR.

ManagedByUnknownVMSet is for handling the case where the node is neither flex nor uniform or avset. (For exmaple, in the future, we have a new node type)

pkg/provider/azure_vmss_cache.go
zmyzheng: The standalone VM will go to the logic of returning ManagedByAvSet, which is the same as before.

Returning ManagedByUnknownVMSet only happens in two cases:
1.  the ProviderID is somehow in wrong format for whatever error
2. a new VM type is supported in the future and we cannot distinguish its providerID

zmyzheng:
> Have you run the tests on CSI repo again?

I updated the code in CSI driver and triggered the test again: https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/1512

pkg/provider/azure.go
```diff
@@ -214,6 +216,9 @@ type Config struct {
 	CloudProviderBackoffRetries int `json:"cloudProviderBackoffRetries,omitempty" yaml:"cloudProviderBackoffRetries,omitempty"`
 	// Backoff duration
 	CloudProviderBackoffDuration int `json:"cloudProviderBackoffDuration,omitempty" yaml:"cloudProviderBackoffDuration,omitempty"`
+	// NonVmssUniformNodesCacheTTLInSeconds sets the Cache TTL for NonVmssUniformNodesCacheTTLInSeconds
+	// if not set, will use default value
+	NonVmssUniformNodesCacheTTLInSeconds int `json:"nonVmssUniformNodesCacheTTLInSeconds,omitempty" yaml:"nonVmssUniformNodesCacheTTLInSeconds,omitempty"`
```
andyzhangx, line 221: need to clarify the relationship between `NonVmssUniformNodesCacheTTLInSeconds` and `AvailabilitySetNodesCacheTTLInSeconds` in the code comment.

zmyzheng: AvailabilitySetNodesCacheTTLInSeconds is not used any more

pkg/provider/azure_vmss.go
feiskyer: if there are only VMSS in the cluster, the data structure doesn't need to be initialized. This would avoid the related APIs are invoked mistakenly by future changes.

pkg/provider/azure_vmss.go
feiskyer: then is it possible for "scaling up" in step1 also falling into this condition? If so, `cloudprovider.InstanceNotFound` should not be returned here.

Are there any possible to trigger cache refreshing instead of default to "cloudprovider.InstanceNotFound" here?

pkg/provider/azure_vmss_cache.go
feiskyer: would this cause issues when the cluster has some [unmanaged nodes](https://cloud-provider-azure.sigs.k8s.io/topics/cross-resource-group-nodes/#unmanaged-nodes) (including virtual-kubelet nodes)?

pkg/provider/azure_vmss.go
zmyzheng: Make sense. I just changed the code to add the check !ss.DisableAvailabilitySetNodes || ss.EnableVmssFlexNodes

pkg/provider/azure_vmss.go
zmyzheng: No. During scaling up, it first check if the node name exists in the cached cluster node names at the time of last cache refresh. It it does not exist, then it means this is a new node and we will trigger force cache refresh, so we don't need to worry about returning "cloudprovider.InstanceNotFound" in this case
image

pkg/provider/azure_vmss_cache.go
zmyzheng: This funciton is only called by ss.GetNodeNameByProviderID. The existing code before this PR has the same issue for handling unmanaged nodes if I understand correctly so this change does not break any eixisting cases.

feiskyer (commented):
/lgtm
/approve

andyzhangx (approved):
/hold cancel

PR #2316, opened 2022-09-16
[release-1.24] Retry if response StatusCode 200 and ContentLength -1

lzhecheng:

#### What type of PR is this?
/kind bug

ARM will possibly return StatusCode 200 and ContentLength -1 which is a bug. However, cloudprovider can handle it better by retrying.

Fixes #

```release-note
ARM will possibly return StatusCode 200 and ContentLength -1 which is a bug. However, cloudprovider can handle it better by retrying.
```

```docs

```


feiskyer (commented):
/lgtm

Issue #2296, opened 2022-09-13
Unnecessary recreating resources because of ARM returns StatusCode 200 and ContentLength -1


lzhecheng:
#### What happened:
Sometimes, ARM returns StatusCode 200 and ContentLength -1 to CCM's GET request but actually the resource exists. This results in CCM thinks that the resource doesn't exist and create a new one, e.g. LB.

#### What you expected to happen:
This problem's root cause is an ARM bug but CCM can better handle it by retrying GET in such situation. Then if timeout, CCM should return error.

#### Environment:

- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Install tools:
- Network plugin and version (if this is a network-related bug):
- Others:

PR #2363, opened 2022-09-22
fix: Do not retry regional when request is not GET

jwtty:
#### What type of PR is this?
/kind bug

Only check response body length when request is GET.

related #2296 

```release-note
Only check response body length when request is GET. Avoids unnecessary retry when deleting an LB.
```

```docs

```

### DIFF
--- a/pkg/azureclients/armclient/azure_armclient.go
+++ b/pkg/azureclients/armclient/azure_armclient.go
@@ -145,7 +145,6 @@ func New(authorizer autorest.Authorizer, clientConfig azureclients.ClientConfig,
 	client.client.Sender = autorest.DecorateSender(client.client,
 		autorest.DoCloseIfError(),
 		retry.DoExponentialBackoffRetry(backoff),
-		DoHackRegionalRetryDecorator(client),
 		DoDumpRequest(10),
 	)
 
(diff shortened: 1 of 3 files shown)


MartinForReal:
send function accept decorator as parameters and maybe we can move DoHackRegionalRetryDecorator to GetResource function ?

feiskyer:
> send function accept decorator as parameters and maybe we can move DoHackRegionalRetryDecorator to GetResource function ?

yes, this is another way to achieve the same goal. Which approach do you prefer?

pkg/azureclients/armclient/azure_armclient.go
```diff
@@ -362,7 +353,7 @@ func (c *Client) GetResource(ctx context.Context, resourceID string, decorators
 		return nil, retry.NewError(false, err)
 	}
 
-	return c.Send(ctx, request)
```
MartinForReal, line 365: c.send(ctx,request,doHackRegionalRetryForGET)?

feiskyer: +1. prefer keeping RegionalRetry still as a decorator

MartinForReal:
<html>
<body>


/test pull-cloud-provider-azure-e2e-ccm-vmss-capz
--


</body>
</html>

MartinForReal:
LGTM

PR #2406, opened 2022-09-27
Get latest service before UpdateLoadBalancer and serialize LB reconcile

jwtty:
#### What type of PR is this?
/kind bug

Avoid race condition between node-sync-loop routine and service reconcile routine.
Avoid UpdateLoadBalancer() using stale service spec to update Load Balancers.

Fixes #

```release-note
Avoid race condition between node-sync-loop routine and service reconcile routine.
Avoid UpdateLoadBalancer() using stale service spec to update Load Balancers.
```

```docs

```

### DIFF
--- a/pkg/provider/azure_loadbalancer.go
+++ b/pkg/provider/azure_loadbalancer.go
@@ -31,6 +31,7 @@ import (
 	"github.com/Azure/go-autorest/autorest/to"
 
 	v1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
 	cloudprovider "k8s.io/cloud-provider"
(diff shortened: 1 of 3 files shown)


pkg/provider/azure_loadbalancer.go
```diff
@@ -176,6 +199,17 @@ func (az *Cloud) UpdateLoadBalancer(ctx context.Context, clusterName string, ser
 		klog.V(5).InfoS("UpdateLoadBalancer Finish", "service", serviceName, "cluster", clusterName, "service_spec", service, "error", err)
 	}()
 
+	// In case UpdateLoadBalancer gets stale service spec, retrieve the lastest from lister
```
lzhecheng, line 202:
```suggestion
	// In case UpdateLoadBalancer gets stale service spec, retrieve the latest from lister
```

jwtty: Good catch! Fixed. Thx

lzhecheng:
lgtm

PR #2421, opened 2022-09-28
feat: add PremiumV2LRS disk type support

andyzhangx:
#### What type of PR is this?
/kind feature

feat: add PremiumV2LRS disk type support

Fixes #

```release-note
feat: add PremiumV2LRS disk type support
```

```docs
feat: add PremiumV2LRS disk type support
```


pkg/azureclients/vmssclient/interface.go
```diff
@@ -28,7 +28,7 @@ import (
 
 const (
 	// APIVersion is the API version for VMSS.
-	APIVersion = "2020-12-01"
+	APIVersion = "2022-03-01"
```
feiskyer, line 31: do we need to update the apiversion in package name as well, e.g. L23 `"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2021-07-01/compute"` is still using `2021-07-01`.

andyzhangx: I have updated to latest `2021-12-01` package, actually there is no api changes between these two packages for disk functionality, that's why `2021-07-01` also works. Anyway, I have updated to latest package.

feiskyer (commented):
/lgtm

PR #2445, opened 2022-10-05
Deepcopy results of cache Get()

lzhecheng:

#### What type of PR is this?
/kind bug

Considering thread safety, results of cache Get() should be deepcopied before used.
* Support deepcopy and add UT for all types
* Add TestGet UT
* Update current UT
* Make Entry structs exported to adjust for deepcopy
* Delete cacheKey for VMSS VM and VMSS deletion

Fixes #

```release-note
Considering thread safety, results of cache Get() should be deepcopied before used.
```

```docs

```

### DIFF
--- a/pkg/cache/azure_cache.go
+++ b/pkg/cache/azure_cache.go
@@ -22,6 +22,8 @@ import (
 	"time"
 
 	"k8s.io/client-go/tools/cache"
+
+	"sigs.k8s.io/cloud-provider-azure/pkg/util/deepcopy"
 )
 
 // AzureCacheReadType defines the read type for cache data
@@ -120,8 +122,14 @@ func (t *TimedCache) getInternal(key string) (*AzureCacheEntry, error) {
 	return newEntry, nil
 }
 
-// Get returns the requested item by key.
+// Get returns the requested item by key with deep copy.
 func (t *TimedCache) Get(key string, crt AzureCacheReadType) (interface{}, error) {
+	data, err := t.get(key, crt)
+	copied := deepcopy.Copy(data)
+	return copied, err
+}
+
+func (t *TimedCache) get(key string, crt AzureCacheReadType) (interface{}, error) {
 	entry, err := t.getInternal(key)
 	if err != nil {
 		return nil, err
--- a/pkg/util/deepcopy/deepcopy.go
+++ b/pkg/util/deepcopy/deepcopy.go
@@ -1,0 +1,139 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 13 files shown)


pkg/cache/azure_cache.go
```diff
@@ -21,6 +21,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/mohae/deepcopy"
```
feiskyer, line 24: it is not actively maintained, could you add a new deep copy utils instead?

lzhecheng: How about this lib? It supports deepcopy and has many users.
https://github.com/jinzhu/copier

lzhecheng: @feiskyer I spent some time debugging with this new lib and realized that it has a bug about nested struct deep copy. And the fix PR is still pending. https://github.com/jinzhu/copier/pull/157
So I decided to use the first lib and put the deepcopy.go into our lib, not in vendor. PTAL. Also please check if there's any copyright issue. Thanks.

pkg/cache/azure_cache.go
```diff
@@ -157,6 +159,45 @@ func (t *TimedCache) Get(key string, crt AzureCacheReadType) (interface{}, error
 	return entry.Data, nil
 }
 
+// Get returns the requested item by key with deep copy.
+func (t *TimedCache) GetWithDeepCopy(key string, crt AzureCacheReadType) (interface{}, error) {
```
lodrem, line 163: Would it be better to decouple `CopyObject` and `Get` since `Get` and `GetWithDeepCopy` share most of their code?

```go
v, _ := cache.Get(...)

// If need mutation
cv := xx.CopyObject(v)
```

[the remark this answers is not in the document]
lzhecheng: Makes sense. Updated.

pkg/util/deepcopy/deepcopy.go
```diff
@@ -0,0 +1,107 @@
...
+import (
+	"reflect"
+)
+
+type deepCopyInterface interface {
+	DeepCopy() interface{}
+}
+
+// Copy deepcopies from v.
+func Copy(src interface{}) interface{} {
```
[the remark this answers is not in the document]
lzhecheng: Updated.

pkg/cache/azure_cache.go
lzhecheng: Good idea. But now I use `Get()` for all references. PTAL

pkg/cache/azure_cache.go
```diff
@@ -120,8 +122,14 @@ func (t *TimedCache) getInternal(key string) (*AzureCacheEntry, error) {
 	return newEntry, nil
 }
 
-// Get returns the requested item by key.
+// Get returns the requested item by key with deep copy.
 func (t *TimedCache) Get(key string, crt AzureCacheReadType) (interface{}, error) {
+	data, err := t.get(key, crt)
+	copied := deepcopy.Copy(data)
```
andyzhangx, line 128: any specifical reason why you want to do this? @lzhecheng 
we found there is OOM in large scale test, every time it ends up copying the VMSS cache whenever a single VM is gotten from it

lzhecheng: It's for thread safety.

PR #2452, opened 2022-10-06
Add support for specifying probe protocol / probe port via annotation per service port

### LINKED ISSUE #1505: Unable to set Loadbalancer Service Health Probe Port
Reopen from #1499 

When configuring Health Checks for Azure Load Balancer you can specify the path and type, but it is not possible to specify the port of the health probe.

Istio Ingress Gateway and similar services do not expose health checks on their actual endpoints (as it can operate across multiple protocols) but instead expose a general health check on a different status port for the gateway.

In AWS this is easily possible with the service.beta.kubernetes.io/aws-load-balancer-healthcheck-port annotation but it does not seem possible in Azure.

This prevents us from configuring health check probes against Istio which need to be targeting the "Status Ready" port not the actual port of the backend.

## Example

Istio hosted on two ports: Port 80 and Port 443 will not serve any traffic until a Virtual Service is deployed
The health check needs to be a HTTP health check against the status port, let's say that's 15021
The probe protocol and endpoint and everything can be set, but the health check port cannot be.

For the following spec:
```
spec:
  ports:
    - name: http2
      protocol: TCP
      port: 80
      targetPort: 8080
    - name: https
      protocol: TCP
      port: 443
      targetPort: 8443
```

The generated probe needs to look like this:

image

rainest:
#### What type of PR is this?

/kind feature

This adds `service.beta.kubernetes.io/port_<num>_health-probe_protocol` and `service.beta.kubernetes.io/port_<num>_health-probe_port` annotations for LoadBalancer Services. These allow overriding the health check port or protocol for a given port. This is useful for ports that cannot respond affirmatively to the health checks and expose health information elsewhere.

Fixes #1505 

This continues https://github.com/kubernetes-sigs/cloud-provider-azure/pull/1508 and attempts to address the open issue with it. I rebased onto upstream master and added commits to de-duplicate checks and correct an apparent error in the original docs (`/` is the default check path, not `/healthz`).

I'm not quite sure I understand the duplication concern raised in https://github.com/kubernetes-sigs/cloud-provider-azure/pull/1508#pullrequestreview-944084110, and it looks like the line indicated is maybe outdated. As I read it, we pass the rule name (for [internal](https://github.com/kubernetes-sigs/cloud-provider-azure/blob/d92742e0ca76326feaae69e175dea1d3a56945b5/pkg/provider/azure_loadbalancer.go#L2193) and [external](https://github.com/kubernetes-sigs/cloud-provider-azure/blob/d92742e0ca76326feaae69e175dea1d3a56945b5/pkg/provider/azure_loadbalancer.go#L2239) LBs) to the check generator, which then just [uses that string as-is](https://github.com/kubernetes-sigs/cloud-provider-azure/blob/d92742e0ca76326feaae69e175dea1d3a56945b5/pkg/provider/azure_loadbalancer.go#L2136) for the probe name.

That name's [generator](https://github.com/kubernetes-sigs/cloud-provider-azure/blob/d92742e0ca76326feaae69e175dea1d3a56945b5/pkg/provider/azure_standard.go#L282-L301) will return something like `SERVICE_UID-??-tcp-443` (the service, subnet, appProtocol, and port--not sure about the second's appearance, but it's optional anyway). That doesn't seem like it can conflict because there will only be one probe per LB rule name, and the the probes are named for the rule they check, not the port they check.

For example, if we have a Service with HTTP appProtocol ports on 8443 and 8009, and want to override the former to use an HTTP check on 8009 instead, the resulting probes would be:

* `ABC-123-http-8443` (an HTTP check on port 8009)
* `ABC-123-http-8009` (an HTTP check on port 8009)

Based on the discussion after, my additional change looks at the generated probes and only keeps those with unique properties, so that the above would only generate one probe. Naming is still a bit iffy, since you'll still get a probe named after only one of the rules, even though it's pulling double-duty. What name scheme would be desirable (changing it requires updating a bunch of tests, so I wanted to confirm before writing it)? I was thinking either `ABC-123-http-8009-8443` (it indicates all ports it checks) or `ABC-123-http-8009-10-5-/`, indicating all of the probe properties. The latter doesn't work great, however, since it needs to account for HTTP paths, which could get quite unwieldy. It could indicate the port and protocol only, but that _would_ result in name conflicts.

```release-note
Added: support for new annotations **service.beta.kubernetes.io/port_<num>_health-probe_protocol** and **service.beta.kubernetes.io/port_<num>_health-probe_port** to allow explicitly setting the health probe protocol individually for each service port. Useful for services like Istio which have health check seperate from the main service port.
```

### DIFF
--- a/site/content/en/topics/loadbalancer.md
+++ b/site/content/en/topics/loadbalancer.md
@@ -33,6 +33,8 @@ Below is a list of annotations supported for Kubernetes services with type `Load
 | `service.beta.kubernetes.io/azure-load-balancer-health-probe-interval` | Health probe interval | Refer to the detailed docs [here](#custom-load-balancer-health-probe) | v1.21 and later  with out-of-tree cloud provider  |
 | `service.beta.kubernetes.io/azure-load-balancer-health-probe-num-of-probe` | The minimum number of unhealthy responses of health probe  |  Refer to the detailed docs [here](#custom-load-balancer-health-probe) |	v1.21 and later  with out-of-tree cloud provider|
 | `service.beta.kubernetes.io/azure-load-balancer-health-probe-request-path` | Request path of the health probe | Refer to the detailed docs [here](#custom-load-balancer-health-probe) | v1.20 and later  with out-of-tree cloud provider|
+| `service.beta.kubernetes.io/port_{port}_health-probe_protocol` | Health probe protocol | {port} is the port number in the service. Explicit protocol for the health probe for the service port {port}, overriding port.appProtocol if set.  Refer to the detailed docs [here](#custom-load-balancer-health-probe) | v1.21 and later with out-of-tree cloud provider|
+| `service.beta.kubernetes.io/port_{port}_health-probe_port` | Health probe protocol |  port} is the port number in the service. Explicit port for the health probe for the service port {port}, overriding the default value.  Refer to the detailed docs [here](#custom-load-balancer-health-probe) | v1.21 and later with out-of-tree cloud provider|
 | `service.beta.kubernetes.io/port_{port}_health-probe_interval` | Health probe interval |  {port} is port number of service.  Refer to the detailed docs [here](#custom-load-balancer-health-probe) | v1.21 and later  with out-of-tree cloud provider|
 | `service.beta.kubernetes.io/port_{port}_health-probe_num-of-probe` | The minimum number of unhealthy responses of health probe  | {port} is port number of service. Refer to the detailed docs [here](#custom-load-balancer-health-probe) |	v1.21 and later with out-of-tree cloud provider|
 | `service.beta.kubernetes.io/port_{port}_health-probe_request-path` | Request path of the health probe | {port} is port number of service.  Refer to the detailed docs [here](#custom-load-balancer-health-probe) | v1.20 and later with out-of-tree cloud provider|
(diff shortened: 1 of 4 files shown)


MartinForReal:
@rainest Thanks for your PR! Could you please merge the commit into one commit?

rainest:
@MartinForReal I can once it's complete, yes. Did you have comments on the de-duplication/naming question and any further work needed there? Or is the change introduced in https://github.com/kubernetes-sigs/cloud-provider-azure/pull/2452/commits/7f80566de8e76f6295f5a09c92012a731f6708d6 all of what you were looking for?

MartinForReal:
Yes! That's what I'm looking for.

rainest:
Alright, sounds good, squashed and rebased onto the current tip of master here.

MartinForReal:
LGTM. 
/assign @feiskyer

MartinForReal:
please rebase the branch in order to fix error reported by ci pipeline.

rainest:
FYI I'll be out away from internet access through the 25th, so if this needs further changes I'll address them then.

rainest:
@MartinForReal checking in on this, are there any further changes it would need?

MartinForReal:
I can't think of any, @feiskyer @nilo19 Could you please take a look?

site/content/en/topics/loadbalancer.md
```diff
@@ -33,6 +33,8 @@ Below is a list of annotations supported for Kubernetes services with type `Load
 | `service.beta.kubernetes.io/azure-load-balancer-health-probe-interval` | Health probe interval | Refer to the detailed docs [here](#custom-load-balancer-health-probe) | v1.21 and later  with out-of-tree cloud provider  |
 | `service.beta.kubernetes.io/azure-load-balancer-health-probe-num-of-probe` | The minimum number of unhealthy responses of health probe  |  Refer to the detailed docs [here](#custom-load-balancer-health-probe) |	v1.21 and later  with out-of-tree cloud provider|
 | `service.beta.kubernetes.io/azure-load-balancer-health-probe-request-path` | Request path of the health probe | Refer to the detailed docs [here](#custom-load-balancer-health-probe) | v1.20 and later  with out-of-tree cloud provider|
+| `service.beta.kubernetes.io/port_{port}_health-probe_protocol` | Health probe protocol | {port} is the port number in the service. Explicit protocol for the health probe for the service port {port}, overriding port.appProtocol if set.  Refer to the detailed docs [here](#custom-load-balancer-health-probe) | v1.21 and later with out-of-tree cloud provider|
```
nilo19, line 36: I think it would be odd to make the variable in the annotation name. I prefer something like `"service.beta.kubernetes.io/port-health-probe-protocol": "port1:protocol1,port2:protocol2"`.

nilo19: OK, we have already made such pattern in other annotations. To change all of them will consume extra time. Let's keep it.

MartinForReal:
<html>
<body>


/test pull-cloud-provider-azure-e2e-ccm-vmss-ip-lb-capz
--


</body>
</html>

rainest:
Unsure what would have broken the one [test for `service.beta.kubernetes.io/azure-load-balancer-resource-group`](https://github.com/rainest/cloud-provider-azure/blob/939c39c369b0fb58d0d78690919afbdd1fbf26eb/tests/e2e/network/service_annotations.go#L356). The Service [never receives an IP assignment](https://github.com/rainest/cloud-provider-azure/blob/939c39c369b0fb58d0d78690919afbdd1fbf26eb/tests/e2e/utils/service_utils.go#L165). Other tests that rely on the same check do get a Service LB IP and succeed, and AFAIK the PR shouldn't do anything special when a resource group is set.

Is there information elsewhere in the test logs that would show this particular Service's state and/or cloud provider logs related to assigning it an IP? I'm unfortunately not too familiar with test suite output, but it looks like it might only be logs from the test client.

MartinForReal:
<html>
<body>


/test pull-cloud-provider-azure-e2e-ccm-capz
--


</body>
</html>

MartinForReal:
<html>
<body>


/test pull-cloud-provider-azure-e2e-ccm-capz
--


</body>
</html>

rainest:
@MartinForReal thanks for the help with the tests. Do you know roughly when changes here propagate out to being available in Azure for users? I didn't find any scheduled update lifecycle in the SIG FAQ, but it looks like new minor provider releases go out roughly every 6mo and are incorporated into Azure shortly after from reviewing the previous releases.

MartinForReal:
> @MartinForReal thanks for the help with the tests. Do you know roughly when changes here propagate out to being available in Azure for users? I didn't find any scheduled update lifecycle in the SIG FAQ, but it looks like new minor provider releases go out roughly every 6mo and are incorporated into Azure shortly after from reviewing the previous releases.

This pr should be cherrypicked to all of release branches and then will be shipped into next cloud provider release(1 week)

and it will take another 2 weeks before it is deployed to all of the regions.

PR #2524, opened 2022-10-14
fix: check internal lb ip in subnet before keeping it

### LINKED ISSUE #2511: ILB frontend should select a new private IP when load-balancer-subnet is changed
#### What happened:
Currently when service load-balancer-subnet gets changed, service reconcile would fail with error, privateIP is not in the subnet.
When service selects a different lb subnet, a new private IP from that subnet should be used.

Create an internal service. Then update the service with a different load-balancer-subnet.
#### Environment:

- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

jwtty:
#### What type of PR is this?
/kind bug

Check internal lb ip in subnet before keeping it, so cx can change their lb subnet

Fixes #2511 

```release-note
Check internal lb ip in subnet before keeping it, so users can change their lb subnet.
```

```docs

```

### DIFF
--- a/pkg/provider/azure_loadbalancer.go
+++ b/pkg/provider/azure_loadbalancer.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"math"
 	"net"
+	"net/netip"
 	"reflect"
 	"sort"
 	"strconv"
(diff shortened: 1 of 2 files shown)


feiskyer (commented):
/lgtm

PR #2635, opened 2022-10-27
Use VM name instead of Node name for disk operations with Flex VMSS

okushchenko:
#### What type of PR is this?

/kind bug

When using https://github.com/kubernetes-sigs/azuredisk-csi-driver with Flex VMSS, I observed that it fails to attach and detach disks from VMs. This is due to it trying to use k8s Node name instead of VM name when interacting with Azure APIs. This PR updates this logic to use a correct name.

```release-note
Fixed a bug that prevents disks from being attached and detached for VMs in a Flex VMSS
```

### DIFF
--- a/pkg/provider/azure_controller_vmssflex.go
+++ b/pkg/provider/azure_controller_vmssflex.go
@@ -99,14 +99,14 @@ func (fs *FlexScaleSet) AttachDisk(ctx context.Context, nodeName types.NodeName,
 
 	klog.V(2).Infof("azureDisk - update(%s): vm(%s) - attach disk list(%s)", nodeResourceGroup, vmName, diskMap)
 
-	future, rerr := fs.VirtualMachinesClient.UpdateAsync(ctx, nodeResourceGroup, vmName, newVM, "attach_disk")
+	future, rerr := fs.VirtualMachinesClient.UpdateAsync(ctx, nodeResourceGroup, *vm.Name, newVM, "attach_disk")
 	if rerr != nil {
 		klog.Errorf("azureDisk - attach disk list(%s) on rg(%s) vm(%s) failed, err: %v", diskMap, nodeResourceGroup, vmName, rerr)
 		if rerr.HTTPStatusCode == http.StatusNotFound {
 			klog.Errorf("azureDisk - begin to filterNonExistingDisks(%v) on rg(%s) vm(%s)", diskMap, nodeResourceGroup, vmName)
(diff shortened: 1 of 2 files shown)


pkg/provider/azure_controller_vmssflex.go
```diff
@@ -38,7 +38,7 @@ func (fs *FlexScaleSet) AttachDisk(ctx context.Context, nodeName types.NodeName,
 		return nil, err
 	}
 
-	nodeResourceGroup, err := fs.GetNodeResourceGroup(vmName)
+	nodeResourceGroup, err := fs.GetNodeResourceGroup(*vm.Name)
```
zmyzheng, line 41: CSI Driver is  using nodeName as the unique key to identify nodes/VMs. Changing to vm.name might cause issue here if external resource group is used. Can you change back to use vmName?

pkg/provider/azure_controller_vmssflex.go
```diff
@@ -97,20 +97,20 @@ func (fs *FlexScaleSet) AttachDisk(ctx context.Context, nodeName types.NodeName,
 		_ = fs.DeleteCacheForNode(vmName)
 	}()
 
-	klog.V(2).Infof("azureDisk - update(%s): vm(%s) - attach disk list(%s)", nodeResourceGroup, vmName, diskMap)
+	klog.V(2).Infof("azureDisk - update(%s): vm(%s) - attach disk list(%s)", nodeResourceGroup, *vm.Name, diskMap)
```
zmyzheng, line 100: Can you change back to use vmName?

zmyzheng (commented):
Thank you so much for finding this issue. You are right that we should use VM name rather than nodeName (OS hostName) when updating VM, so line 102,109, 182, 189 should be fixed as you proposed in the PR. 

As for the rest lines, since CSI Driver is using nodename as the primary key to identify nodes/VMs, can we keep the way of using vmName? Only changing line 102,109, 182, 189 should be able to fix the issue

zmyzheng (commented):
Can you also update line 222 as well?

okushchenko:
@zmyzheng thanks for the quick review! I updated the PR with your suggestions and updated test cases for the `UpdateVM` method to account for the change that you proposed.

zmyzheng (approved):
/lgtm

okushchenko:
@zmyzheng I rebased and squashed the commits to satisfy the bots, feel free to modify this PR however you think is necessary to get it merged.

zmyzheng (approved):

zmyzheng:
@feiskyer @andyzhangx could you help approve this PR?

PR #2665, opened 2022-10-31
fix: should update dns label if target pip already have one and it needs update

MartinForReal:
#### What type of PR is this?
/kind bug
/kind regression

fix: should update dns label if target pip already have one and it needs update

Fixes #

```release-note
Fix an issue when DNS label of public ip is not updated when service.beta.kubernetes.io/azure-dns-label-name is changed. 
```

```docs

```


MartinForReal:
<html>
<body>


/test pull-cloud-provider-azure-unit
--


</body>
</html>

PR #2671, opened 2022-11-01
VMSS Flex: GetNodeNameByIPConfigurationID should use InterfacesClient to query vmName

### LINKED ISSUE #2670: cloud-provider-azure-e2e-ccm-vmssflex-capz test fails because CCM fails find node from IP Configuration
#### Which jobs are failing:
cloud-provider-azure-e2e-ccm-vmssflex-capz

#### Which test(s) are failing:
should support service annotation 'service.beta.kubernetes.io/azure-dns-label-name' 

#### Since when has it been failing:
Oct 28, 2022

#### Testgrid link:
https://testgrid.k8s.io/provider-azure-cloud-provider-azure#cloud-provider-azure-e2e-ccm-vmssflex-capz

#### Reason for failure:
When creating a VM and attaching it to an existing VMSS Flex, the nic name always has VM name as prefix. However, when adding VMSS Flex VM by increasing the capacity of the VMSS Flex, the new VM's nic name does not have VM name as prefix. In this case, an extra Nic get call is needed to find out which VM this Nic belongs to

#### Anything else we need to know:
NONE

zmyzheng:
#### What type of PR is this?
/kind bug
/kind failing-test

When creating a VM and attaching it to an existing VMSS Flex, the nic name always has VM name as prefix. In this case, CCM can easily finds the VM name by IP config ID. However, when adding VMSS Flex VM by increasing the capacity of the VMSS Flex, the new VM's nic name does not have VM name as prefix. In this case, an extra Nic get call is needed to find out which VM this Nic belongs to.

Fixes #2670

- When using VMSS Unifom, the Nic IP config always contains the VMSS name and intance ID, so we can easily find out which VM this Nic belongs to.
- When using AvailabilitySet, the Nic name always has VM name as prefix, so we can also easily find out which VM this Nic belongs to
- When using VMSS Flex, if the VM is individually created and attached to a VMSS Flex, then the nic name always has VM name as prefix. In this case, CCM can easily finds the VM name by IP config ID.
- When using VMSS Flex, if the VM is added by  increasing the capacity of the VMSS Flex, the new VM's nic name does not have VM name as prefix. In this case, there is no direct way to map nic to VM except send a Get Nic API call.

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/provider/azure_vmssflex.go
+++ b/pkg/provider/azure_vmssflex.go
@@ -394,7 +394,23 @@ func (fs *FlexScaleSet) getNodeInformationByIPConfigurationID(ipConfigurationID 
 		return "", "", "", fmt.Errorf("invalid ip config ID %s", ipConfigurationID)
 	}
 
-	vmName := strings.Replace(nicName, "-nic", "", 1)
+	// get vmName by nic name
+	ctx, cancel := getContextWithCancel()
+	defer cancel()
+	nic, rerr := fs.InterfacesClient.Get(ctx, nicResourceGroup, nicName, "")
+	if rerr != nil {
+		return "", "", "", fmt.Errorf("getNodeInformationByIPConfigurationID(%s): failed to get interface of name %s: %w", ipConfigurationID, nicName, rerr.Error())
+	}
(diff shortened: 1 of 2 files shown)


feiskyer (approved):
/lgtm

PR #2691, opened 2022-11-03
Fix issue: CCM fails to delete LB because updating VMSS Flex network profile fails

### LINKED ISSUE #2690: CCM fails to delete LB because updating VMSS Flex network profile fails
#### Which jobs are failing:
cloud-provider-azure-e2e-ccm-vmssflex-capz

#### Which test(s) are failing:
EnsureLoadBalancer should not update any resources when service config is not changed

#### Since when has it been failing:
11/01/2022

#### Testgrid link:
https://testgrid.k8s.io/provider-azure-cloud-provider-azure#cloud-provider-azure-e2e-ccm-vmssflex-capz

#### Reason for failure:
When deleting backend pool from VMSS, both VMSS Uniform and VMSS Flex are added into the same vmssNameMap: https://github.com/kubernetes-sigs/cloud-provider-azure/blob/master/pkg/provider/azure_vmss.go#L1755. 

However, CCM tries to find those VMSS from the VMSS Uniform cache: https://github.com/kubernetes-sigs/cloud-provider-azure/blob/master/pkg/provider/azure_vmss.go#L2045. Because of that, VMSS Flex cannot be found and returns "Instance not found" error

zmyzheng:
#### What type of PR is this?
/kind bug
/kind failing-test

When deleting backend pool from VMSS, both VMSS Uniform and VMSS Flex are added into the same vmssNameMap: https://github.com/kubernetes-sigs/cloud-provider-azure/blob/master/pkg/provider/azure_vmss.go#L1755.

However, CCM tries to find those VMSS from the VMSS Uniform cache: https://github.com/kubernetes-sigs/cloud-provider-azure/blob/master/pkg/provider/azure_vmss.go#L2045. Because of that, VMSS Flex cannot be found and returns "Instance not found" error

This PR tries to fix this issue.

Fixes #2690 

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/provider/azure_vmss.go
+++ b/pkg/provider/azure_vmss.go
@@ -1682,24 +1682,75 @@ func getScaleSetAndResourceGroupNameByIPConfigurationID(ipConfigurationID string
 }
 
 func (ss *ScaleSet) ensureBackendPoolDeletedFromVMSS(backendPoolID, vmSetName string) error {
-	vmssNamesMap := make(map[string]bool)
+	if !ss.useStandardLoadBalancer() {
+		found := false
 
-	// the standard load balancer supports multiple vmss in its backend while the basic sku doesn't
-	if ss.useStandardLoadBalancer() {
 		cachedUniform, err := ss.vmssCache.Get(consts.VMSSKey, azcache.CacheReadTypeDefault)
 		if err != nil {
 			klog.Errorf("ensureBackendPoolDeletedFromVMSS: failed to get vmss uniform from cache: %v", err)
 			return err
 		}
+		vmssUniformMap := cachedUniform.(*sync.Map)
+
+		vmssUniformMap.Range(func(key, value interface{}) bool {
+			vmssEntry := value.(*VMSSEntry)
+			if to.String(vmssEntry.VMSS.Name) == vmSetName {
+				found = true
+				return false
+			}
+			return true
+		})
+		if found {
+			return ss.ensureBackendPoolDeletedFromVmssUniform(backendPoolID, vmSetName)
+		}
+
 		flexScaleSet := ss.flexScaleSet.(*FlexScaleSet)
 		cachedFlex, err := flexScaleSet.vmssFlexCache.Get(consts.VmssFlexKey, azcache.CacheReadTypeDefault)
 		if err != nil {
 			klog.Errorf("ensureBackendPoolDeletedFromVMSS: failed to get vmss flex from cache: %v", err)
 			return err
 		}
-		vmssUniformMap := cachedUniform.(*sync.Map)
 		vmssFlexMap := cachedFlex.(*sync.Map)
+		vmssFlexMap.Range(func(key, value interface{}) bool {
+			vmssFlex := value.(*compute.VirtualMachineScaleSet)
+			if to.String(vmssFlex.Name) == vmSetName {
+				found = true
+				return false
+			}
+			return true
+		})
 
+		if found {
+			return flexScaleSet.ensureBackendPoolDeletedFromVmssFlex(backendPoolID, vmSetName)
+		}
+
+		return cloudprovider.InstanceNotFound
+
+	}
+
+	err := ss.ensureBackendPoolDeletedFromVmssUniform(backendPoolID, vmSetName)
+	if err != nil {
+		return err
+	}
+	if !ss.DisableAvailabilitySetNodes || ss.EnableVmssFlexNodes {
+		flexScaleSet := ss.flexScaleSet.(*FlexScaleSet)
+		err = flexScaleSet.ensureBackendPoolDeletedFromVmssFlex(backendPoolID, vmSetName)
+	}
+	return err
+}
+
+func (ss *ScaleSet) ensureBackendPoolDeletedFromVmssUniform(backendPoolID, vmSetName string) error {
+	vmssNamesMap := make(map[string]bool)
+
+	// the standard load balancer supports multiple vmss in its backend while the basic sku doesn't
+	if ss.useStandardLoadBalancer() && !ss.EnableMultipleStandardLoadBalancers {
+		cachedUniform, err := ss.vmssCache.Get(consts.VMSSKey, azcache.CacheReadTypeDefault)
+		if err != nil {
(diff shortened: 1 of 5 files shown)


nilo19:
Is it possible to add a flex vmss cache in the scaleSet struct, and query it when the vmss is not found? I think in this way the change will be easier.

zmyzheng:
> Is it possible to add a flex vmss cache in the scaleSet struct, and query it when the vmss is not found? I think in this way the change will be easier.

I had a similar thought at the beginning, but I realize this will not work in the pure vmss flex cluster where tmType is set to vmssflex. In this case, ScaleSet is not initialized, instead, FlexScaleSet is initialized: https://github.com/kubernetes-sigs/cloud-provider-azure/blob/master/pkg/provider/azure.go#L629 , and the checking 0 instnace VMSS logic will not be executed. The way proposed in this PR will handle both cases.

The reason this PR is large is because it actually does two things:
1. fix the issue in your previous PR https://github.com/kubernetes-sigs/cloud-provider-azure/pull/2489 which tries to find vmssflex from vmss uniform cache 
2. Add similar logic to FlexScaleSet implementation as what you did in your PRhttps://github.com/kubernetes-sigs/cloud-provider-azure/pull/2489.

pkg/provider/azure_vmss.go
```diff
@@ -1682,24 +1682,76 @@ func getScaleSetAndResourceGroupNameByIPConfigurationID(ipConfigurationID string
...
 		cachedUniform, err := ss.vmssCache.Get(consts.VMSSKey, azcache.CacheReadTypeDefault)
 		if err != nil {
 			klog.Errorf("ensureBackendPoolDeletedFromVMSS: failed to get vmss uniform from cache: %v", err)
 			return err
 		}
+		vmssUniformMap := cachedUniform.(*sync.Map)
+
+		vmssUniformMap.Range(func(key, value interface{}) bool {
+			vmssEntry := value.(*VMSSEntry)
+			if to.String(vmssEntry.VMSS.Name) == vmSetName {
```
nilo19, line 1697: Why do we need to check the vmSetName here?

pkg/provider/azure_vmss.go
```diff
@@ -1752,15 +1804,17 @@ func (ss *ScaleSet) ensureBackendPoolDeletedFromVMSS(backendPoolID, vmSetName st
...
-		vmssFlexMap.Range(walk)
 		if len(errorList) > 0 {
 			return utilerrors.Flatten(utilerrors.NewAggregate(errorList))
 		}
 	} else {
 		vmssNamesMap[vmSetName] = true
 	}
 
-	return ss.EnsureBackendPoolDeletedFromVMSets(vmssNamesMap, backendPoolID)
+	if len(vmssNamesMap) > 0 {
```
nilo19, line 1814: We don't have to check the length here.

pkg/provider/azure_vmss.go
```diff
@@ -1682,24 +1682,76 @@ func getScaleSetAndResourceGroupNameByIPConfigurationID(ipConfigurationID string
...
+	err := ss.ensureBackendPoolDeletedFromVmssUniform(backendPoolID, vmSetName)
+	if err != nil {
+		return err
+	}
+	if !ss.DisableAvailabilitySetNodes || ss.EnableVmssFlexNodes {
+		flexScaleSet := ss.flexScaleSet.(*FlexScaleSet)
+		err = flexScaleSet.ensureBackendPoolDeletedFromVmssFlex(backendPoolID, vmSetName)
+	}
+	return err
+
```
nilo19, line 1740: nit: remove the line.

pkg/provider/azure_vmss.go
zmyzheng: Removed

pkg/provider/azure_vmss.go
zmyzheng: Done

pkg/provider/azure_vmss.go
zmyzheng: Because at this point, we don't know if this vmSetName is Uniform or Flex yet

nilo19: Why does the name matter? Do you mean there could be vmss flex in vmssUniformMap?

zmyzheng: When the if logic on line 1685 is true, CCM only needs to delete backend Pool from the VMSS of vmSetName. However, CCM does not know if vmSetName is a Flex scale set or Uniform scale set. So, it first goes through the Unifrom VMSS cache and check if any VMSS in the cache has name "vmSetName". If it does not find any, then it continues to go through the VMSS Flex cache.

zmyzheng: The line 1685 means it is using basic LB or using multiple SLB

nilo19: It seems we don't need to check if we use multi-slb or not. We just claim here that we want vmss A, B and C to decouple with the LB. If there is no LB ref on A and B, it will be two no-ops.

zmyzheng: Make sense. I removed checking multi-slb just now.

zmyzheng:
/retest

Kubernetes Prow Robot ***@***.***>于2022年11月4日 周五17:33写道：

nilo19:
@zmyzheng could you please cherry-pick it to release-1.25?

zmyzheng:
> @zmyzheng could you please cherry-pick it to release-1.25?

Hi @nilo19 , do we need to do the cherry pick for this PR? I thought the vmss flex feature will not be avaibable before v1.26 so I did not do any cherry pick for the PRs I sent before. The issue mained in this PR will not happen for vmss uniform cluster. Instead, the issue will only be triggered when the vmss flex is used.

nilo19:
> > @zmyzheng could you please cherry-pick it to release-1.25?
> 
> Hi @nilo19 , do we need to do the cherry pick for this PR? I thought the vmss flex feature will not be avaibable before v1.26 so I did not do any cherry pick for the PRs I sent before. The issue mained in this PR will not happen for vmss uniform cluster. Instead, the issue will only be triggered when the vmss flex is used.

Then we don't have to cherry-pick it.

PR #2707, opened 2022-11-07
fix public ip list call reduction

### LINKED ISSUE #2706: publicIP list calls are not successfully reduced
#### What happened:
PublicIPAddress List calls are not successfully reduced:
I1028 21:04:41.482802       1 azure_armclient.go:145] Send.sendRequest got response with ContentLength -1, StatusCode 200 and responseBody length 103248
I1028 21:04:41.490402       1 azure_backoff.go:338] PublicIPAddressesClient.List(...) success
W1028 21:04:41.490452       1 azure_standard.go:367] serviceOwnsFrontendIP: unexpected error when finding match public IP of the service ... with loadBalancerLP ...: findMatchedPIPByLoadBalancerIP: cannot find public IP with IP address ... in resource group ...
I1028 21:04:41.602206       1 azure_armclient.go:145] Send.sendRequest got response with ContentLength -1, StatusCode 200 and responseBody length 103248
I1028 21:04:41.612513       1 azure_backoff.go:338] PublicIPAddressesClient.List(...) success
W1028 21:04:41.612580       1 azure_standard.go:367] serviceOwnsFrontendIP: unexpected error when finding match public IP of the service ... with loadBalancerLP ...: findMatchedPIPByLoadBalancerIP: cannot find public IP with IP address ... in resource group ...
I1028 21:04:41.786204       1 azure_armclient.go:145] Send.sendRequest got response with ContentLength -1, StatusCode 200 and responseBody length 103248
I1028 21:04:41.792915       1 azure_backoff.go:338] PublicIPAddressesClient.List(...) success
W1028 21:04:41.792956       1 azure_standard.go:367] serviceOwnsFrontendIP: unexpected error when finding match public IP of the service ... with loadBalancerLP ...: findMatchedPIPByLoadBalancerIP: cannot find public IP with IP address ... in resource group ...

#### Environment:

- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

jwtty:
#### What type of PR is this?
/kind bug

Original `pip = &pipList` does not modify the passed slice actually.

Fixes #2706 

```release-note
Fix to reduce PublicIPAddress list arm calls
```

```docs

```

### DIFF
--- a/pkg/provider/azure_loadbalancer.go
+++ b/pkg/provider/azure_loadbalancer.go
@@ -85,7 +85,8 @@ func (az *Cloud) GetLoadBalancer(ctx context.Context, clusterName string, servic
 	// there is a chance that we could orphan public IP resources while we delete the load balancer (kubernetes/kubernetes#80571).
 	// We need to make sure the existence of the load balancer depends on the load balancer resource and public IP resource on Azure.
 	existsPip := func() bool {
-		pipName, _, err := az.determinePublicIPName(clusterName, service, nil)
+		var pips []network.PublicIPAddress
+		pipName, _, err := az.determinePublicIPName(clusterName, service, &pips)
 		if err != nil {
 			return false
 		}
@@ -140,7 +141,8 @@ func (az *Cloud) reconcileService(ctx context.Context, clusterName string, servi
 		return nil, err
 	}
 
-	lbStatus, fipConfig, err := az.getServiceLoadBalancerStatus(service, lb, nil)
+	var pips []network.PublicIPAddress
+	lbStatus, fipConfig, err := az.getServiceLoadBalancerStatus(service, lb, &pips)
 	if err != nil {
 		klog.Errorf("getServiceLoadBalancerStatus(%s) failed: %v", serviceName, err)
 		if !errors.Is(err, ErrorNotVmssInstance) {
@@ -646,7 +648,7 @@ func (az *Cloud) getServiceLoadBalancer(service *v1.Service, clusterName string,
 	}
 
 	// reuse pip list to reduce api call
-	var pips *[]network.PublicIPAddress
+	var pips []network.PublicIPAddress
 
 	// check if the service already has a load balancer
 	for i := range existingLBs {
@@ -684,7 +686,7 @@ func (az *Cloud) getServiceLoadBalancer(service *v1.Service, clusterName string,
 			continue
 		}
 		var fipConfig *network.FrontendIPConfiguration
-		status, fipConfig, err = az.getServiceLoadBalancerStatus(service, &existingLB, pips)
+		status, fipConfig, err = az.getServiceLoadBalancerStatus(service, &existingLB, &pips)
 		if err != nil {
 			return nil, nil, false, err
 		}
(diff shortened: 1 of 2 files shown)


pkg/provider/azure_loadbalancer.go
```diff
@@ -919,11 +919,18 @@ func (az *Cloud) determinePublicIPName(clusterName string, service *v1.Service,
 
 func (az *Cloud) findMatchedPIPByLoadBalancerIP(service *v1.Service, loadBalancerIP, pipResourceGroup string, pips *[]network.PublicIPAddress) (*network.PublicIPAddress, error) {
 	if pips == nil {
+		return nil, fmt.Errorf("findMatchedPIPByLoadBalancerIP: nil pip list passed")
+	} else if cap(*pips) == 0 {
+		// use capacity to check whether pip has been listed
```
lodrem, line 924: Is it an implicit requirement when we call this function? It seems to be error-prone.

feiskyer: why not using `len()` here? (`len()` is used in other places)

feiskyer: do you think it makes sense to save the listed PIP in an in-memory cache instead of passing the pointer to other functions? e.g. something like vmss cache or vm cache?

pkg/provider/azure_loadbalancer.go
```diff
@@ -919,11 +919,18 @@ func (az *Cloud) determinePublicIPName(clusterName string, service *v1.Service,
 
 func (az *Cloud) findMatchedPIPByLoadBalancerIP(service *v1.Service, loadBalancerIP, pipResourceGroup string, pips *[]network.PublicIPAddress) (*network.PublicIPAddress, error) {
 	if pips == nil {
```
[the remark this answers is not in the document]
jwtty: this actually should not happen.

pkg/provider/azure_loadbalancer.go
jwtty: updated, thanks

pkg/provider/azure_loadbalancer.go
```diff
@@ -919,11 +921,14 @@ func (az *Cloud) determinePublicIPName(clusterName string, service *v1.Service,
...
+		// this should not happen
+		return nil, fmt.Errorf("findMatchedPIPByLoadBalancerIP: nil pip list passed")
+	} else if *pips == nil {
```
lodrem, line 926: Should it check the length? Empty slice does not equal `nil`.
```suggestion
	} else if len(*pips) == 0 {
```

jwtty: If length is 0, that means we already listed pip but there isn't any. We check nil to decide whether to list pip.

lodrem: > We check nil to decide whether to list pip.
should it be `if pip == nil`?

lodrem: nil is a valid slice, and by convention we will use `nil` as an `empty slice`. I suggest we better don't distinguish them.

In this case, the type is `*[]T`, so why not:
- `pip == nil`: to list pip
- `len(*pip) == 0`: already list, but it's empty
- else

lodrem: if so, could you please add some comments to the function in case the caller misuse `nil` and `[]T`?

jwtty: Added comments to make it clear. Thanks

feiskyer (approved):
/lgtm

PR #2723, opened 2022-11-09
fix: check private dns zone before creation

ZeroMagic:
#### What type of PR is this?
/kind bug

We should add  get PrivateDNSZone and get VNetLink to make this func idempotent, so that even creation return failed, it should succeeded in next try

Fixes #

```release-note
fix: check private dns zone before creation
```

```docs

```

### DIFF
--- a/pkg/provider/azure_storageaccount.go
+++ b/pkg/provider/azure_storageaccount.go
@@ -205,15 +205,21 @@ func (az *Cloud) EnsureStorageAccount(ctx context.Context, accountOptions *Accou
 	}
 
 	if accountOptions.CreatePrivateEndpoint {
-		// Create DNS zone first, this could make sure driver has write permission on vnetResourceGroup
-		if err := az.createPrivateDNSZone(ctx, vnetResourceGroup); err != nil {
-			return "", "", fmt.Errorf("Failed to create private DNS zone(%s) in resourceGroup(%s), error: %v", PrivateDNSZoneName, vnetResourceGroup, err)
+		if _, err := az.privatednsclient.Get(ctx, vnetResourceGroup, PrivateDNSZoneName); err != nil {
+			klog.V(2).Infof("get private dns zone %s returned with %v", PrivateDNSZoneName, err.Error())
+			// Create DNS zone first, this could make sure driver has write permission on vnetResourceGroup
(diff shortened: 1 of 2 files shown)


pkg/provider/azure_storageaccount.go
```diff
@@ -607,3 +613,13 @@ func isPrivateEndpointAsExpected(account storage.Account, accountOptions *Accoun
 	}
 	return false
 }
+
+func (az *Cloud) getPrivateDNSZone(ctx context.Context, vnetResourceGroup string) (privatedns.PrivateZone, *retry.Error) {
+	klog.V(2).Infof("Getting private dns zone(%s) in resourceGroup (%s)", PrivateDNSZoneName, vnetResourceGroup)
```
andyzhangx, line 618: no need to print `Getting private dns zone` logs

andyzhangx: since there is only one line of code, I think it's better cancel this function, and use `az.privatednsclient.Get` directly.

pkg/provider/azure_storageaccount.go
```diff
@@ -607,3 +613,13 @@ func isPrivateEndpointAsExpected(account storage.Account, accountOptions *Accoun
...
+	return az.privatednsclient.Get(ctx, vnetResourceGroup, PrivateDNSZoneName)
+}
+
+func (az *Cloud) getVNetLink(ctx context.Context, vNetLinkName, vnetResourceGroup, vnetName string) (privatedns.VirtualNetworkLink, *retry.Error) {
+	klog.V(2).Infof("Getting virtual link for vnet(%s) and DNS Zone(%s) in resourceGroup(%s)", vNetLinkName, PrivateDNSZoneName, vnetResourceGroup)
+	return az.virtualNetworkLinksClient.Get(ctx, vnetResourceGroup, PrivateDNSZoneName, vNetLinkName)
```
andyzhangx, line 624: cancel this function and use `az.virtualNetworkLinksClient.Get` in outer call

andyzhangx (commented):
pls submit a candidate PR in azure file csi driver, only modify the vendor change, and then we could verify whether e2e test pass

andyzhangx (approved):
/lgtm
/approve

PR #2741, opened 2022-11-11
[VMSS Flex] refresh vmssflex cache after updating vmss flex backendpool

### LINKED ISSUE #2740: VMSS Flex Cache should be refreshed after updating backend pool
#### What happened:
Assume vmssflex has backend pool [0, 1, 2, 3]
After removing backend pool 1, the actual backend pool left shouild be [0, 2, 3]
However, when getting vmss flex from cache:https://github.com/kubernetes-sigs/cloud-provider-azure/blob/master/pkg/provider/azure_vmssflex_cache.go#L297, the get method is a deep copy: https://github.com/kubernetes-sigs/cloud-provider-azure/blob/master/pkg/cache/azure_cache.go#L128, so updating the vmssflex retrieved from the cache does not update the vmssflex in the cache.
In this case, if a new backend pool 4 is to be added, then the vmss update operation will fail because the deleted backend pool 1 is still in the cache.

The error message is like below
```
ensureVMSSFlexInPool begins to add vmss(k8sflex-mp-0) with new backendPoolID /subscriptions/ccde0aa8-3434-40ee-8442-13c8e4160d62/resourceGroups/mizhe-k8sflex/providers/Microsoft.Network/loadBalancers/flexprofile/backendAddressPools/flexprofile
E1110 22:10:33.204076       1 azure_backoff.go:585] CreateOrUpdateVMSS: error CreateOrUpdate vmss(k8sflex-mp-0): &{false 404 0001-01-01 00:00:00 +0000 UTC {
  "error": {
    "details": [],
    "code": "NotFound",
    "message": "Resource /subscriptions/ccde0aa8-3434-40ee-8442-13c8e4160d62/resourceGroups/MIZHE-K8SFLEX/providers/Microsoft.Network/loadBalancers/FLEXPROFILE-INTERNAL not found."
  }
}}
```

The vmssflex cache should be refreshed after updating backend pool
#### Environment:

- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

zmyzheng:
#### What type of PR is this?
/kind bug
/kind failing-test

Assume vmssflex has backend pool [0, 1, 2, 3]
After removing backend pool 1, the actual backend pool left shouild be [0, 2, 3]
However, when getting vmss flex from cache:https://github.com/kubernetes-sigs/cloud-provider-azure/blob/master/pkg/provider/azure_vmssflex_cache.go#L297, the get method is a deep copy: https://github.com/kubernetes-sigs/cloud-provider-azure/blob/master/pkg/cache/azure_cache.go#L128, so updating the vmssflex retrieved from the cache does not update the vmssflex in the cache.
In this case, if a new backend pool 4 is to be added, then the vmss update operation will fail because the deleted backend pool 1 is still in the cache.

The error message is like below
```
ensureVMSSFlexInPool begins to add vmss(k8sflex-mp-0) with new backendPoolID /subscriptions/ccde0aa8-3434-40ee-8442-13c8e4160d62/resourceGroups/mizhe-k8sflex/providers/Microsoft.Network/loadBalancers/flexprofile/backendAddressPools/flexprofile
E1110 22:10:33.204076       1 azure_backoff.go:585] CreateOrUpdateVMSS: error CreateOrUpdate vmss(k8sflex-mp-0): &{false 404 0001-01-01 00:00:00 +0000 UTC {
  "error": {
    "details": [],
    "code": "NotFound",
    "message": "Resource /subscriptions/ccde0aa8-3434-40ee-8442-13c8e4160d62/resourceGroups/MIZHE-K8SFLEX/providers/Microsoft.Network/loadBalancers/FLEXPROFILE-INTERNAL not found."
  }
}}
```

Fixes #2740 

The vmssflex cache should be refreshed after updating backend pool

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/provider/azure_vmssflex.go
+++ b/pkg/provider/azure_vmssflex.go
@@ -584,7 +584,7 @@ func (fs *FlexScaleSet) EnsureHostInPool(service *v1.Service, nodeName types.Nod
 }
 
 func (fs *FlexScaleSet) ensureVMSSFlexInPool(service *v1.Service, nodes []*v1.Node, backendPoolID string, vmSetNameOfLB string) error {
-	klog.V(2).Infof("ensureVMSSInPool: ensuring VMSS Flex with backendPoolID %s", backendPoolID)
+	klog.V(2).Infof("ensureVMSSFlexInPool: ensuring VMSS Flex with backendPoolID %s", backendPoolID)
 	vmssFlexIDsMap := make(map[string]bool)
 
 	if !fs.useStandardLoadBalancer() {
@@ -612,12 +612,12 @@ func (fs *FlexScaleSet) ensureVMSSFlexInPool(service *v1.Service, nodes []*v1.No
 			// in this scenario the vmSetName is an empty string and the name of vmss should be obtained from the provider IDs of nodes
 			vmssFlexID, err := fs.getNodeVmssFlexID(node.Name)
 			if err != nil {
-				klog.Error("ensureVMSSInPool: failed to get VMSS Flex ID of node: %s, will skip checking and continue", node.Name)
+				klog.Error("ensureVMSSFlexInPool: failed to get VMSS Flex ID of node: %s, will skip checking and continue", node.Name)
 				continue
 			}
 			resourceGroupName, err := fs.GetNodeResourceGroup(node.Name)
 			if err != nil {
-				klog.Error("ensureVMSSInPool: failed to get resource group of node: %s, will skip checking and continue", node.Name)
+				klog.Error("ensureVMSSFlexInPool: failed to get resource group of node: %s, will skip checking and continue", node.Name)
 				continue
 			}
 
@@ -629,13 +629,13 @@ func (fs *FlexScaleSet) ensureVMSSFlexInPool(service *v1.Service, nodes []*v1.No
 	} else {
 		vmssFlexID, err := fs.getVmssFlexIDByName(vmSetNameOfLB)
 		if err != nil {
-			klog.Error("ensureVMSSInPool: failed to get VMSS Flex ID of vmSet: %s, ", vmSetNameOfLB)
+			klog.Error("ensureVMSSFlexInPool: failed to get VMSS Flex ID of vmSet: %s, ", vmSetNameOfLB)
 			return err
 		}
 		vmssFlexIDsMap[vmssFlexID] = true
 	}
 
-	klog.V(2).Infof("ensureVMSSInPool begins to update VMSS list %v with backendPoolID %s", vmssFlexIDsMap, backendPoolID)
+	klog.V(2).Infof("ensureVMSSFlexInPool begins to update VMSS list %v with backendPoolID %s", vmssFlexIDsMap, backendPoolID)
 	for vmssFlexID := range vmssFlexIDsMap {
 		vmssFlex, err := fs.getVmssFlexByVmssFlexID(vmssFlexID, azcache.CacheReadTypeDefault)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/provider/azure_vmssflex.go
```diff
@@ -881,10 +885,14 @@ func (fs *FlexScaleSet) EnsureBackendPoolDeletedFromVMSets(vmssNamesMap map[stri
 				},
 			}
 
-			klog.V(2).Infof("fs.EnsureBackendPoolDeletedFromVMSets begins to update vmss(%s) with backendPoolID %s", vmssName, backendPoolID)
+			defer func() {
+				fs.vmssFlexCache.Get(consts.VmssFlexKey, azcache.CacheReadTypeForceRefresh)
```
nilo19, line 889: Do we need to refresh the cache if failed to update the VMSS?

zmyzheng: Yes, I discovered refreshing the cache when VMSS update fails can help the operation self-healing

nilo19: I think it means we will have an extra API call when the failure is not retriable as well.

zmyzheng: That's true, but this is the best of what we can do. If we don't add this API call, then the VMSS updating will keep failing and this will introduce much more wasted API call. I have saw this happenning in the last few days' E2E testing.

Like Pengfei mentioned before, let's first focus on the functional correctness as top priority. After all the E2E tests pass, we can then further make optimization for the performance

nilo19: Can we only refresh the cache after a successful operation?

nilo19:
@lzhecheng do we have this issue for other cached resources?

pkg/provider/azure_vmssflex.go
zmyzheng: No, I have tried only refreshing the cache after succussful operation, but it does not always work. The special case I discoverd is: if the cache does not reflect the latest status, then the VMSS update will keep failing again and again. 

For example, the cache says backend pool 2 is still with VMSS Flex, but actually backendpool 2 is already removed. At this time, if we want to add backend pool 5 to the VMSS, then the update will fail because it operation will request update VMSS with backend pool 2 and 5. However, since backendpool 2 is already deleted, the update operation will fail and retry will not resolve this issue until the cache expires. But before cache expires, we can see a lot of failed retry in the CCM log. This is why we need to refresh the cache even if the operation fails.

zmyzheng:
> @lzhecheng do we have this issue for other cached resources?

Yes and No.

If there is no special handling, this same issue also applies to VMSS Uniform cache. However, after updating backend pool of VMSS Uniform VM nodes, the node is removed from the cache because the node's NIC info is not up to date
- https://github.com/kubernetes-sigs/cloud-provider-azure/blob/master/pkg/provider/azure_vmss.go#L1898
- https://github.com/kubernetes-sigs/cloud-provider-azure/blob/master/pkg/provider/azure_vmss.go#L1426

Because the node is not in the cache any more, when next time CCM tries to add/delete another backend pool from the node, both **VMSS cahce** and **VMSS VM cache** will be forcely refreshed: https://github.com/kubernetes-sigs/cloud-provider-azure/blob/master/pkg/provider/azure_vmss.go#L834. After this force refresh, the VMSS cache is updated again.

When using VMSS Flex, there is no need to remove the VMSS Flex VM from the cache because the NIC is not a sub-resource of the VM and the VM info itself is still up to date. Because of this, only updating VMSS cache is good enough, which is kind of better than VMSS Uniform.

pkg/provider/azure_vmssflex.go
nilo19: Can we invalidate the cache by removing the entry? This could save some API calls.

zmyzheng: The VMSS Flex Cache stores all VMSS Flex in the cluster. There is only one entry key for this cache: https://github.com/kubernetes-sigs/cloud-provider-azure/blob/master/pkg/consts/consts.go#L149

All the VMSS Flex are queried with one batch API call using VMSS List API, similar to what we are doing for the VMSS Uniform cache

zmyzheng: Sure, I just modified the code to remove the entry as you suggested

nilo19:
LGTM overall. Is it possible to unit test this change?

zmyzheng:
> LGTM overall. Is it possible to unit test this change?

The existing unit test should already cover this change

nilo19:
> > LGTM overall. Is it possible to unit test this change?
> 
> The existing unit test already cover this change: image
> 
> image
> 
> I also manually tested it in my dev box.

Yes, because it is in defer func, so it will be covered all the time. Can you add a test case to cover this behavior? If it's time consuming, we can hold this for now, and you may use your private fork to test what you need.

zmyzheng:
> > > LGTM overall. Is it possible to unit test this change?
> > 
> > 
> > The existing unit test already cover this change: image
> > image
> > I also manually tested it in my dev box.
> 
> Yes, because it is in defer func, so it will be covered all the time. Can you add a test case to cover this behavior? If it's time consuming, we can hold this for now, and you may use your private fork to test what you need.

I might not get your point. Can you clarify which behavior do you want me to add a test case to cover?

zmyzheng:
> > > > LGTM overall. Is it possible to unit test this change?
> > > 
> > > 
> > > The existing unit test already cover this change: image
> > > image
> > > I also manually tested it in my dev box.
> > 
> > 
> > Yes, because it is in defer func, so it will be covered all the time. Can you add a test case to cover this behavior? If it's time consuming, we can hold this for now, and you may use your private fork to test what you need.
> 
> I might not get your point. Can you clarify which behavior do you want me to add a test case to cover?

Update: changed the unit test to verify the cache is empty.

PR #2791, opened 2022-11-17
feat: detect errors that requires vm udpate retries and retry if detected

sunpa93:
#### What type of PR is this?
/kind cleanup

* This PR allows `cloud-provider-azure` to retry vm update on `OperationPreempted` error with http status code of 2xx, where storage config change has been committed but vm update has failed.
* This PR adds `PartialUpdateError` type to implement `error` interface to be returned for other VM update failures with status code of 2xx.
* This PR purports to address a reported customer issue where attaches fail due to `OperationPreempted` error.

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/provider/azure_controller_common.go
+++ b/pkg/provider/azure_controller_common.go
@@ -28,6 +28,7 @@ import (
 	"time"
 
 	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2022-03-01/compute"
+	"github.com/Azure/go-autorest/autorest/azure"
 
 	"k8s.io/apimachinery/pkg/types"
 	kwait "k8s.io/apimachinery/pkg/util/wait"
@@ -38,6 +39,7 @@ import (
 
 	azcache "sigs.k8s.io/cloud-provider-azure/pkg/cache"
 	"sigs.k8s.io/cloud-provider-azure/pkg/consts"
+	"sigs.k8s.io/cloud-provider-azure/pkg/retry"
 )
 
 const (
@@ -53,6 +55,10 @@ const (
 	sourceVolume           = "volume"
 	attachDiskMapKeySuffix = "attachdiskmap"
 	detachDiskMapKeySuffix = "detachdiskmap"
+
+	updateVMRetryDuration = time.Duration(1) * time.Second
+	updateVMRetryFactor   = 3.0
+	updateVMRetrySteps    = 5
 
 	// WriteAcceleratorEnabled support for Azure Write Accelerator on Azure Disks
 	// https://docs.microsoft.com/azure/virtual-machines/windows/how-to-enable-write-accelerator
@@ -72,9 +78,16 @@ var defaultBackOff = kwait.Backoff{
 	Jitter:   0.0,
 }
 
+var updateVMBackoff = kwait.Backoff{
+	Duration: updateVMRetryDuration,
+	Factor:   updateVMRetryFactor,
+	Steps:    updateVMRetrySteps,
+}
+
 var (
 	managedDiskPathRE  = regexp.MustCompile(`.*/subscriptions/(?:.*)/resourceGroups/(?:.*)/providers/Microsoft.Compute/disks/(.+)`)
 	diskSnapshotPathRE = regexp.MustCompile(`.*/subscriptions/(?:.*)/resourceGroups/(?:.*)/providers/Microsoft.Compute/snapshots/(.+)`)
+	errorCodeRE        = regexp.MustCompile(`Code="(.*?)".*`)
 )
 
 type controllerCommon struct {
@@ -254,9 +267,10 @@ func (c *controllerCommon) AttachDisk(ctx context.Context, async bool, diskName,
 	c.diskStateMap.Store(disk, "attaching")
 	defer c.diskStateMap.Delete(disk)
 	future, err := vmset.AttachDisk(ctx, nodeName, diskMap)
-	if err != nil {
-		return -1, err
+	if future == nil {
+		return -1, fmt.Errorf("nil future was returned: %w", err)
 	}
+	// err will be handled by waitForUpdateResult below
 
 	if async && c.diskOpRateLimiter.TryAccept() {
 		// unlock and wait for attach disk complete
@@ -267,7 +281,40 @@ func (c *controllerCommon) AttachDisk(ctx context.Context, async bool, diskName,
 			klog.Warningf("azureDisk - switch to batch operation due to rate limited, QPS: %f", c.diskOpRateLimiter.QPS())
 		}
 	}
-	return lun, vmset.WaitForUpdateResult(ctx, future, nodeName, "attach_disk")
+
+	if err = c.waitForUpdateResult(ctx, vmset, nodeName, future, err); err != nil {
+		return -1, err
+	}
+	return lun, nil
+}
+
+// waitForUpdateResult handles asynchronous VM update operations and retries with backoff if OperationPreempted error is observed
+func (c *controllerCommon) waitForUpdateResult(ctx context.Context, vmset VMSet, nodeName types.NodeName, future *azure.Future, updateErr error) (err error) {
+	err = updateErr
+
+	if err == nil {
+		err = vmset.WaitForUpdateResult(ctx, future, nodeName, "attach_disk")
+	}
+
+	if vmUpdateRequired(future, err) {
+		if derr := kwait.ExponentialBackoffWithContext(ctx, updateVMBackoff, func() (bool, error) {
+			klog.Errorf("Retry VM Update on node (%s) due to error (%v)", nodeName, err)
+			future, err = vmset.UpdateVMAsync(ctx, nodeName)
+			if err == nil {
+				err = vmset.WaitForUpdateResult(ctx, future, nodeName, "attach_disk")
+			}
+			return !vmUpdateRequired(future, err), nil
+		}); derr != nil {
+			err = derr
+			return
+		}
+	}
+
+	if err != nil && configAccepted(future) {
+		err = retry.NewPartialUpdateError(err.Error())
+	}
+
+	return
 }
 
 func (c *controllerCommon) insertAttachDiskRequest(diskURI, nodeName string, options *AttachDiskOptions) error {
@@ -620,6 +667,11 @@ func (c *controllerCommon) checkDiskExists(ctx context.Context, diskURI string) 
 	return true, nil
 }
 
+func vmUpdateRequired(future *azure.Future, err error) bool {
+	errCode := getAzureErrorCode(err)
+	return configAccepted(future) && errCode == consts.OperationPreemptedErrorCode
+}
+
 func getValidCreationData(subscriptionID, resourceGroup, sourceResourceID, sourceType string) (compute.CreationData, error) {
 	if sourceResourceID == "" {
 		return compute.CreationData{
@@ -662,4 +714,22 @@ func isInstanceNotFoundError(err error) bool {
 		return true
 	}
 	return strings.Contains(errMsg, errStatusCode400) && strings.Contains(errMsg, errInvalidParameter) && strings.Contains(errMsg, errTargetInstanceIds)
+}
+
+// getAzureErrorCode uses regex to parse out the error code encapsulated in the error string.
+func getAzureErrorCode(err error) string {
+	if err == nil {
+		return ""
+	}
+	matches := errorCodeRE.FindStringSubmatch(err.Error())
+	if matches == nil {
+		return ""
+	}
+	return matches[1]
+}
+
+// configAccepted returns true if storage profile change had been committed (i.e. HTTP status code == 2xx) and returns false otherwise.
+func configAccepted(future *azure.Future) bool {
+	// if status code indicates success, the storage profile change was committed
+	return future != nil && future.Response() != nil && future.Response().StatusCode/100 == 2
 }
--- a/pkg/provider/azure_controller_vmss.go
+++ b/pkg/provider/azure_controller_vmss.go
@@ -263,20 +263,31 @@ func (ss *ScaleSet) DetachDisk(ctx context.Context, nodeName types.NodeName, dis
 
 // UpdateVM updates a vm
 func (ss *ScaleSet) UpdateVM(ctx context.Context, nodeName types.NodeName) error {
+	future, err := ss.UpdateVMAsync(ctx, nodeName)
+	if err != nil {
+		return err
+	}
+
+	return ss.WaitForUpdateResult(ctx, future, nodeName, "update_vm")
+}
+
+// UpdateVMAsync updates a vm asynchronously
+func (ss *ScaleSet) UpdateVMAsync(ctx context.Context, nodeName types.NodeName) (*azure.Future, error) {
 	vmName := mapNodeNameToVMName(nodeName)
 	vm, err := ss.getVmssVM(vmName, azcache.CacheReadTypeDefault)
 	if err != nil {
-		return err
+		return nil, err
 	}
 
 	nodeResourceGroup, err := ss.GetNodeResourceGroup(vmName)
 	if err != nil {
-		return err
+		return nil, err
 	}
 
-	var updateResult *compute.VirtualMachineScaleSetVM
+	var future *azure.Future
 	var rerr *retry.Error
 
+	// Invalidate the cache right after updating
 	defer func() {
 		// If there is an error with Update operation,
 		// invalidate the cache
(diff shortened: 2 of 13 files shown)


sunpa93:
e2e tests triggered for v1 driver to verify changes (https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/1606)

sunpa93:
@feiskyer Could you review the PR? The change is to address a reported customer issue with `OperationPreempted` error. The change has been tested on both v1 (https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/1606) and v2 driver (https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/1605)

pkg/provider/azure_controller_common.go
```diff
@@ -267,7 +279,38 @@ func (c *controllerCommon) AttachDisk(ctx context.Context, async bool, diskName,
...
-	return lun, vmset.WaitForUpdateResult(ctx, future, nodeName, "attach_disk")
+
+	backoffManager := wait.Backoff{
+		Duration: updateVMUpdateRetryDuration,
+		Factor:   updateVMUpdateRetryFactor,
+		Steps:    updateVMUpdateRetrySteps,
+	}
+
+	err = vmset.WaitForUpdateResult(ctx, future, nodeName, "attach_disk")
+	for c.vmUpdateRequired(future, err) {
```
andyzhangx, line 290: there is already `UpdateVM` func, I think you could use that function directly, and for disk attach/detach, it does not need backoff, if it failed once, then return error directly, CSI driver has backoff mechanism, it would retry next.

https://github.com/kubernetes-sigs/cloud-provider-azure/blob/063aa0f8e2a12f11b354a20a188cae2af45000b1/pkg/provider/azure_controller_common.go#L375-L385

pkg/provider/azure_controller_common.go
```diff
@@ -620,6 +663,11 @@ func (c *controllerCommon) checkDiskExists(ctx context.Context, diskURI string)
 	return true, nil
 }
 
+func (c *controllerCommon) vmUpdateRequired(future *azure.Future, err error) bool {
+	errCode := getErrorCode(err)
```
andyzhangx, line 667: why not do string compare directly with `Code="OperationPreempted"`, that would be quite straightforward than current logic.

andyzhangx: I may get your point, it requires status code = 2xx, but if `Code="OperationPreempted"`, I think status code is always 2xx, and do vm update once does not harm, so only string compare is more straightforward IMO, but that way, you don't need following code:
```
func getErrorCode(err error) *string {
	if err == nil {
		return nil
	}
	matches := errorCodeRE.FindStringSubmatch(err.Error())
	if matches == nil {
		return nil
	}
	return &matches[1]
}

type FutureParser interface {
	ConfigAccepted(future *azure.Future) bool
}

type controllerFutureParser struct{}

func (c *controllerFutureParser) ConfigAccepted(future *azure.Future) bool {
	// if status code indicates success, the storage profile change was committed
	return future != nil && future.Response() != nil && future.Response().StatusCode/100 == 2
}
```

andyzhangx:
> @feiskyer Could you review the PR? The change is to address a reported customer issue with `OperationPreempted` error. The change has been tested on both v1 ([kubernetes-sigs/azuredisk-csi-driver#1606](https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/1606)) and v2 driver ([kubernetes-sigs/azuredisk-csi-driver#1605](https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/1605))

@sunpa93 sorry for the late review, just ping me when you have any other questions, thanks.

pkg/provider/azure_controller_common.go
sunpa93: We have observed other errors with status code of `2xx`, where we were instructed that VM Update needs to be retried until success. However, there were certain errors with code = `2xx`, where we couldn't recover by simply retrying.  The error code and future parsing was for future code extensibility so that we can address issues with other errors with `2xx`, where retry can solve this issue.

pkg/provider/azure_controller_common.go
sunpa93: The purpose of this PR is to address a customer reported issue where a retry triggered by csi driver fails to handle `OperationPreempted` error due to `VM Provisioning` state not being in failed state. After discussing the scenario with the VMSS team, we found it more appropriate to retry on certain errors (in this case, specifically `OperationPreempted` error) in cloud-provider-azure layer. And we needed an async version of `UpdateVM` function as the code change has dependency on `future` value.

sunpa93: The current change only retries if `OperationPreempted` error occurs. Or are you suggesting that we get rid of retry in whole? 

The reason we cannot rely on CSI driver retry and backoff is that from what we have gathered from customer reported incidence discussion, we have no guarantee that VM provisioning state would be in `failed` state on the second `ControllerPublishVolume` call, in which case I think the driver returns success as the disk lun had already been assigned and commited for the disk. That is why we are trying to retry in `cloud-provider-azure` level, for which case we would need a back off implemented as the next VM update can fail with the same error.

Do you have any specific concerns or preferred way to address this issue?

pkg/provider/azure_controller_common.go
```diff
@@ -267,7 +279,38 @@ func (c *controllerCommon) AttachDisk(ctx context.Context, async bool, diskName,
 			klog.Warningf("azureDisk - switch to batch operation due to rate limited, QPS: %f", c.diskOpRateLimiter.QPS())
 		}
 	}
-	return lun, vmset.WaitForUpdateResult(ctx, future, nodeName, "attach_disk")
+
+	backoffManager := wait.Backoff{
```
andyzhangx, line 283: move this var declaration out of this function, could be after :
```
var defaultBackOff = kwait.Backoff{
```

pkg/provider/azure_controller_common.go
```diff
@@ -663,3 +711,25 @@ func isInstanceNotFoundError(err error) bool {
 	}
 	return strings.Contains(errMsg, errStatusCode400) && strings.Contains(errMsg, errInvalidParameter) && strings.Contains(errMsg, errTargetInstanceIds)
 }
+
+func getErrorCode(err error) *string {
```
andyzhangx, line 715: I think it's better returning `string` instead of `*string`, returning pointer has risk of panic.
And write comments and ut for this func.

andyzhangx: `getAzureErrorCode` is better since it's azure specifc

pkg/provider/azure_controller_common.go
```diff
@@ -663,3 +711,25 @@ func isInstanceNotFoundError(err error) bool {
...
+	return &matches[1]
+}
+
+type FutureParser interface {
+	ConfigAccepted(future *azure.Future) bool
+}
+
+type controllerFutureParser struct{}
+
+func (c *controllerFutureParser) ConfigAccepted(future *azure.Future) bool {
```
andyzhangx, line 732: I think write a simple function `ConfigAccepted(future *azure.Future) bool` would be more clear, and add comments for this function.

pkg/provider/azure_controller_vmss.go
```diff
@@ -263,47 +263,40 @@ func (ss *ScaleSet) DetachDisk(ctx context.Context, nodeName types.NodeName, dis
...
-		// invalidate the cache
-		if rerr != nil {
-			_ = ss.DeleteCacheForNode(vmName)
-			return
-		}
-
-		// Update the cache with the updated result only if its not nil
-		// and contains the VirtualMachineScaleSetVMProperties
-		if updateResult != nil && updateResult.VirtualMachineScaleSetVMProperties != nil {
-			if err := ss.updateCache(vmName, nodeResourceGroup, vm.VMSSName, vm.InstanceID, updateResult); err != nil {
```
andyzhangx, line 291: not expected code change, pls rebase to master branch

pkg/provider/azure_controller_common.go
```diff
@@ -254,8 +262,12 @@ func (c *controllerCommon) AttachDisk(ctx context.Context, async bool, diskName,
 	c.diskStateMap.Store(disk, "attaching")
 	defer c.diskStateMap.Delete(disk)
 	future, err := vmset.AttachDisk(ctx, nodeName, diskMap)
-	if err != nil {
```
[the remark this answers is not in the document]
sunpa93: We were trying to create a comprehensive method to handle errors with 200 code. Although we don't think we can get an error with 200 code for `UpdateAsync` function, we couldn't find any supporting documents from VMSS to confirm this. And we believe it is better to be safe here and be able to handle `PartialUpdateError` here as well.,

pkg/provider/azure_controller_common.go
```diff
@@ -267,7 +279,38 @@ func (c *controllerCommon) AttachDisk(ctx context.Context, async bool, diskName,
...
+		Factor:   updateVMUpdateRetryFactor,
+		Steps:    updateVMUpdateRetrySteps,
+	}
+
+	err = vmset.WaitForUpdateResult(ctx, future, nodeName, "attach_disk")
+	for c.vmUpdateRequired(future, err) {
+		select {
+		case <-ctx.Done():
+			return -1, context.DeadlineExceeded
+		case <-time.After(backoffManager.Step()):
```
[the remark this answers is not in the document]
sunpa93: Even if backoff ends, the plan was to loop until context timeout if updates continue failing due to `OperationPreempted` error.  We back off with a pretty aggressive factor here, so by the time backoff exhausts all its steps, it would have a `3^5`   seconds of wait time. As we discussed previously, there is no guarantee of when the operation is going to succeed, so it was hard to put a hard cap on how many update retries we should attempt. Then, we believed it would be reasonable to retry either until success or the given context times out.

Please let me know what you think.

pkg/provider/azure_controller_common.go
```diff
@@ -620,6 +663,11 @@ func (c *controllerCommon) checkDiskExists(ctx context.Context, diskURI string)
 	return true, nil
 }
 
+func (c *controllerCommon) vmUpdateRequired(future *azure.Future, err error) bool {
+	errCode := getAzureErrorCode(err)
+	return configAccepted(future) && errCode != "" && errCode == consts.OperationPreemptedErrorCode
```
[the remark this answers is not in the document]
sunpa93: Good catch :)

pkg/provider/azure_controller_common.go
```diff
@@ -254,8 +262,12 @@ func (c *controllerCommon) AttachDisk(ctx context.Context, async bool, diskName,
 	c.diskStateMap.Store(disk, "attaching")
 	defer c.diskStateMap.Delete(disk)
 	future, err := vmset.AttachDisk(ctx, nodeName, diskMap)
-	if err != nil {
-		return -1, err
+	if future == nil {
+		return -1, fmt.Errorf("nil future was returned: %v", err)
+	} else if err != nil {
+		if !configAccepted(future) {
```
andyzhangx, line 268: if there is 2XX return code with `err != nil`, but it's not `OperationPreemptedErrorCode`, in current PR, it would still call `vmset.WaitForUpdateResult`, is that expected?

pkg/provider/azure_controller_common.go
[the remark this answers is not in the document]
andyzhangx: thanks for the explanation. I think if backoff ends, it should exit, if first few tries failed, the next try will also fail probably per my past experience, you may refer to this, and use defaultBackoff if that's feasible.

https://github.com/kubernetes-sigs/cloud-provider-azure/blob/8fb8dd8d3f23872587607436d02f938f3c6e9725/pkg/provider/azure_managedDiskController.go#L257-L273

pkg/provider/azure_controller_common.go
sunpa93: Given the implementation of polling tracker and how async operations are polled, we expect the `WaitForUpdateResult` to fail with the same error, which we intend to handle in the retry loop.
https://github.com/kubernetes-sigs/cloud-provider-azure/blob/5e5c4af9bee955f0a266f6b8182ca0221044cd1d/pkg/provider/azure_controller_common.go#L290-L301

But I think we could add an additional check to be safe and only `WaitForUpdateResult` after `vmset.AttachDisk` if the operation did not result in error.

pkg/provider/azure_controller_common.go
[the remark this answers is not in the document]
sunpa93: I kept the backoff configurations for attach as is because I felt that factor of `1.5` was too small for handling `OperationPreempted` errors and thought we would hammering the front end. But I updated the loop to use `ExponentialBackoffWithContext`.

pkg/provider/azure_controller_vmss.go
sunpa93: For async operation, I believe the cache update would happen at the `WaitForUpdateResult`

https://github.com/kubernetes-sigs/cloud-provider-azure/blob/e2197aaae8487b0d64530791d05e423e6081cacb/pkg/provider/azure_controller_vmss.go#L144-L153

pkg/provider/azure_controller_common.go
```diff
@@ -72,9 +78,16 @@ var defaultBackOff = kwait.Backoff{
 	Jitter:   0.0,
 }
 
+var attachBackoff = kwait.Backoff{
+	Duration: updateVMUpdateRetryDuration,
```
andyzhangx, line 82: remove `Update` word: `updateVMRetryDuration`

pkg/provider/azure_controller_common.go
```diff
@@ -72,9 +78,16 @@ var defaultBackOff = kwait.Backoff{
 	Jitter:   0.0,
 }
 
+var attachBackoff = kwait.Backoff{
```
andyzhangx, line 81: `updateVMBackoff`

pkg/provider/azure_controller_common.go
```diff
@@ -267,7 +281,46 @@ func (c *controllerCommon) AttachDisk(ctx context.Context, async bool, diskName,
 			klog.Warningf("azureDisk - switch to batch operation due to rate limited, QPS: %f", c.diskOpRateLimiter.QPS())
 		}
 	}
-	return lun, vmset.WaitForUpdateResult(ctx, future, nodeName, "attach_disk")
+
+	err = c.waitForUpdateResult(ctx, vmset, nodeName, future, err)
```
andyzhangx, line 285: `return lun, c.waitForUpdateResult(ctx, vmset, nodeName, future, err)` is ok, it's not necessary to log success since there is already logging outside (in CSI driver)

pkg/provider/azure_controller_common.go
```diff
@@ -267,7 +281,46 @@ func (c *controllerCommon) AttachDisk(ctx context.Context, async bool, diskName,
...
+
+// waitForUpdateResult handles asynchronous VM update operations and retries with backoff if OperationPreempted error is observed
+func (c *controllerCommon) waitForUpdateResult(ctx context.Context, vmset VMSet, nodeName types.NodeName, future *azure.Future, updateErr error) (err error) {
+	err = updateErr
+
+	if err == nil {
+		err = vmset.WaitForUpdateResult(ctx, future, nodeName, "attach_disk")
+	}
+
+	if vmUpdateRequired(future, err) {
```
andyzhangx, line 305: should use `if vmUpdateRequired(future, updateErr)` since `err` is changed in L302?

pkg/provider/azure_controller_common.go
```diff
@@ -267,7 +281,46 @@ func (c *controllerCommon) AttachDisk(ctx context.Context, async bool, diskName,
...
+		if derr := kwait.ExponentialBackoffWithContext(ctx, attachBackoff, func() (bool, error) {
```
andyzhangx, line 306: why not use `err = kwait.ExponentialBackoffWithContext` directly here, then you don't need to `err = derr` in L314

pkg/provider/azure_controller_vmss.go
andyzhangx: there is already a PR (with refactor) to address this: https://github.com/kubernetes-sigs/cloud-provider-azure/pull/2924, pls revert this change and only focus on vm update retries, thanks.

sunpa93: The removal of the cache update in the defer statement is a part of changes that `UpdateAsync` function brings. We can't keep the cache update in the defer here as there won't be any result to update the cache with in this async call.

pkg/provider/azure_controller_common.go
sunpa93: We will be assessing `vmUpdateRequired` based on `WaitForUpdateResult` if `updateErr == nil`. We would be assessing `vmUpdateRequired` with `updateErr` only when `updateErr != nil`, in which case,  we don't issue a `WaitForUpdateResult` call.

pkg/provider/azure_controller_common.go
sunpa93: That's because we can't return `err` directly in the condition func of `kwait.ExponentialBackoffWithContext` as returning a non-nil error would break the loop.

andyzhangx (approved):
thanks, code logic is quite clear now.
Could you update the csi driver vendor change, and make sure e2e test pass, and then I will approve this code change. I need to be very cautious about this change since it's quite fundamental change in csi driver.

pkg/provider/azure_controller_common.go
```diff
@@ -254,9 +267,10 @@ func (c *controllerCommon) AttachDisk(ctx context.Context, async bool, diskName,
 	c.diskStateMap.Store(disk, "attaching")
 	defer c.diskStateMap.Delete(disk)
 	future, err := vmset.AttachDisk(ctx, nodeName, diskMap)
-	if err != nil {
-		return -1, err
+	if future == nil {
+		return -1, fmt.Errorf("nil future was returned: %v", err)
 	}
+	// err will be handled in retry loop below if necessary
```
edreed, line 273: `err will be handled by waitForUpdateResult below`

sunpa93:
tracking e2e tests with the change: https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/1606

pkg/provider/azure_controller_common.go
```diff
@@ -254,9 +267,10 @@ func (c *controllerCommon) AttachDisk(ctx context.Context, async bool, diskName,
 	c.diskStateMap.Store(disk, "attaching")
 	defer c.diskStateMap.Delete(disk)
 	future, err := vmset.AttachDisk(ctx, nodeName, diskMap)
-	if err != nil {
-		return -1, err
+	if future == nil {
+		return -1, fmt.Errorf("nil future was returned: %v", err)
```
andyzhangx, line 271: pls fix golint error, thanks.

andyzhangx (approved):
/lgtm
/approve
/hold cancel

PR #2805, opened 2022-11-21
fix: add disk lun check in AttachDisk to avoid race condition

andyzhangx:
#### What type of PR is this?
/kind bug

fix: add disk lun check in AttachDisk to avoid race condition

Fixes #

There is corner case that second disk attach would get a different LUN num and since disk is already attached to the other LUN num (first disk attach may already time out), second disk attach is returned with a wrong LUN num

```release-note
fix: add disk lun check in AttachDisk to avoid race condition
```

```docs
fix: add disk lun check in AttachDisk to avoid race condition
```

### DIFF
--- a/pkg/provider/azure_controller_standard.go
+++ b/pkg/provider/azure_controller_standard.go
@@ -18,6 +18,7 @@ package provider
 
 import (
 	"context"
+	"fmt"
 	"net/http"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


MartinForReal:
<html>
<body>


/test pull-cloud-provider-azure-check
--


</body>
</html>

ZeroMagic (approved):
/lgtm

PR #2817, opened 2022-11-23
fix: resource group issue in disk attach

andyzhangx:
#### What type of PR is this?
/kind bug

fix: resource group issue in disk attach

Original behavior is using disk resource group as node resource group which is wrong, there would be issue when disk is not in node resource group, this PR fixes the issue.

https://github.com/kubernetes-sigs/cloud-provider-azure/blob/c905fb10cb277916fb580fef75d276f328bc0d7c/pkg/provider/azure_controller_common.go#L270-L274

Fixes #

```release-note
fix: resource group issue in disk attach
```

```docs
fix: resource group issue in disk attach
```

### DIFF
--- a/pkg/provider/azure_controller_common.go
+++ b/pkg/provider/azure_controller_common.go
@@ -267,11 +267,7 @@ func (c *controllerCommon) AttachDisk(ctx context.Context, async bool, diskName,
 			klog.Warningf("azureDisk - switch to batch operation due to rate limited, QPS: %f", c.diskOpRateLimiter.QPS())
 		}
 	}
-	resourceGroup, _, err := getInfoFromDiskURI(diskURI)
-	if err != nil {
-		return -1, err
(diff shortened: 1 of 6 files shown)


mukhoakash:
/lgtm
@andyzhangx I have written an e2e test case with similar changes to fix the resource group fix in the csi driver. I will discard by branch and test with your changes to verify that this issue gets covered in the csi driver e2e test.

mukhoakash (approved):

mukhoakash:
> /lgtm @andyzhangx I have written an e2e test case with similar changes to fix the resource group fix in the csi driver. I will discard by branch and test with your changes to verify that this issue gets covered in the csi driver e2e test.

This is my branch for reference: https://github.com/mukhoakash/cloud-provider-azure/tree/rg_bug_fix

PR #2867, opened 2022-12-06
Revert "Deepcopy results of cache Get()"

edreed:
This reverts commit be10aa29fd9750b05614e534a38799032cf76e62.

#### What type of PR is this?

/kind bug
/kind regression

Reverts #2445. `TimedCache` is used to hold the list of VMs (value) for each VM Scale Set (key). When there a 100s to 1000s of nodes, the entire list ends up copied when a single VM is retrieved from the cache consuming a lot of memory unnecessarily. This can quickly cause OOMKilled conditions when operations (e.g. managed disk attach or detach) are performed on the VMs of the scale set concurrently.

Current usages of the values returned by `TimedCache.Get` appear to be read-only, so the copy seems unnecessary.

Fixes #

```release-note
NONE
```

```docs

```


lzhecheng:
Hello @edreed ,
Deepcopy is for thread safety so even if it is not necessary for VM related resources, it is good for other resource types, e.g. PIP usage in this method:
https://github.com/kubernetes-sigs/cloud-provider-azure/blob/45ab1dc3a1b3f95458136ad757b6b5e58d023736/pkg/provider/azure_loadbalancer.go#L993
Could you try to 1) find a way to optimize VMSS and VM related `TimedCache` or 2) do shallow copy for VM only?

pkg/cache/azure_cache.go
```diff
@@ -122,14 +120,8 @@ func (t *TimedCache) getInternal(key string) (*AzureCacheEntry, error) {
 	return newEntry, nil
 }
 
-// Get returns the requested item by key with deep copy.
+// Get returns the requested item by key.
 func (t *TimedCache) Get(key string, crt AzureCacheReadType) (interface{}, error) {
-	data, err := t.get(key, crt)
```
andyzhangx, line 127: I think we could only revert change in `Get` function, that's minimal revert.

andyzhangx:
> Hello @edreed , Deepcopy is for thread safety so even if it is not necessary for VM related resources, it is good for other resource types, e.g. PIP usage in this method:
> 
> https://github.com/kubernetes-sigs/cloud-provider-azure/blob/45ab1dc3a1b3f95458136ad757b6b5e58d023736/pkg/provider/azure_loadbalancer.go#L993
> 
> 
> Could you try to 1) find a way to optimize VMSS and VM related `TimedCache` or 2) do shallow copy for VM only?

@lzhecheng if there is deep copy requirement, I think it should be called out side of `func (t *TimedCache) Get(key string, crt AzureCacheReadType)`, currently it's deep copying every caching `data`, and if size of `data` is big, and `TimedCache.Get` is called frequently, memory would be exhausted quickly. For this `func (t *TimedCache) Get`, I think it's better not deep copy by default, and if there is deep copy requirement, just call it out side of `func (t *TimedCache) Get`, that would be more flexiable.

lzhecheng:
@andyzhangx I think we can do shallow copy and deep copy for different resource types but it should depend more on if the resource is read-only rather than data size. I think it is better to find a way to reduce data size instead of simply ignoring thread safefy. So here VM can have shallow copy but not for PIP. @feiskyer 
As for the code change, actually I have considered such design when I created my  PR:
Current:
https://github.com/kubernetes-sigs/cloud-provider-azure/blob/6e788a44e57393c2721fa7adaedcfc2ddbb70ab8/pkg/cache/azure_cache.go#L126-L132

Can be:
```
func (t *TimedCache) GetWithDeepCopy(key string, crt AzureCacheReadType) (interface{}, error) {
	data, err := t.get(key, crt)
	copied := deepcopy.Copy(data)
	return copied, err
}

func (t *TimedCache) Get(key string, crt AzureCacheReadType) (interface{}, error) {
	return t.get(key, crt)
}

func (t *TimedCache) get(key string, crt AzureCacheReadType) (interface{}, error) {
...
}
```

lzhecheng:
I will create a PR

lzhecheng:
@edreed BTW, how much memory is used in your scale test, like for 1000 nodes?

edreed:
> I will create a PR

I will close this PR once #2869 merges.

andyzhangx:
close this PR since https://github.com/kubernetes-sigs/cloud-provider-azure/pull/2949 has been merged

PR #2875, opened 2022-12-08
cleanup: set context as parameter in cloud provider functions

umagnus:
#### What type of PR is this?

/kind cleanup

cleanup: set context as parameter in cloud provider functions

Fixes #

```release-note
none
```

```docs
none
```

### DIFF
--- a/pkg/provider/azure_vmssflex.go
+++ b/pkg/provider/azure_vmssflex.go
@@ -17,6 +17,7 @@ limitations under the License.
 package provider
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"strconv"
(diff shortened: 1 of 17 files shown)


pkg/provider/azure_vmssflex.go
```diff
@@ -65,11 +66,11 @@ func newFlexScaleSet(az *Cloud) (VMSet, error) {
 	}
 
 	var err error
-	fs.vmssFlexCache, err = fs.newVmssFlexCache()
+	fs.vmssFlexCache, err = fs.newVmssFlexCache(context.Background())
```
andyzhangx, line 69: should add ctx parameter in `newFlexScaleSet`, the intention is that the outside consumer, .e.g. csi driver would pass the ctx in outside

andyzhangx (approved):
/lgtm
@nilo19 could you also take a look? this PR tries to set context in external facing api, by this PR, ccm and csi driver could set the correct context outside.

andyzhangx:
@umagnus there are many places are using `getContextWithCancel()` func, I think we could also do similar change to set context in external functions, make context really useful.

PR #2951, opened 2022-12-15
improvement: Flex - Reduce throttling on Attach/Detach disk operations

gcampbell12:
#### What type of PR is this?

/kind bug

Currently refreshing the cache for VMSS flex in `getNodeVmssFlexID` is very expensive as each scale set is refreshed despite only needing to get the scale set for the specified node - this often leads to hitting ARM rate limits in even small sized environments. 
- To reduce calls to ARM APIs here I try and work out the correct scaleset from the node name by stripping the last 6 characters (similar to how the uniform version of this works) - this works for VMs which are created in flex scalesets using scaling as the vmss name is used as the computer name prefix, for vm's which are added to a Flexible scale set this won't be useful. The results are either appended or prepended to a slice to provide an ordered list of vmss to check
- Using the above slice we continue to refresh the cache as before but after each `vmssFlexVMCache.Get` we'll check the vm cache to see if the desired node now exists and return immediately if so, with the above ordered list we'll potentially only need to have one iteration of the vmss range

This leads to vastly reduced ARM API usage - especially on Attach and Detach disk operations and also makes those operations faster as we're not looping though however many scalesets you happen to have 

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/provider/azure_vmssflex_cache.go
+++ b/pkg/provider/azure_vmssflex_cache.go
@@ -25,6 +25,7 @@ import (
 	"time"
 
 	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2022-03-01/compute"
+	"github.com/Azure/go-autorest/autorest/to"
 
 	cloudprovider "k8s.io/cloud-provider"
 	"k8s.io/klog/v2"
@@ -181,18 +182,35 @@ func (fs *FlexScaleSet) getNodeVmssFlexID(nodeName string) (string, error) {
 		}
 		vmssFlexes := cached.(*sync.Map)
 
+		var vmssFlexIDs []string
 		vmssFlexes.Range(func(key, value interface{}) bool {
 			vmssFlexID := key.(string)
-			_, err := fs.vmssFlexVMCache.Get(vmssFlexID, azcache.CacheReadTypeForceRefresh)
-			if err != nil {
-				klog.Errorf("failed to refresh vmss flex VM cache for vmssFlexID %s", vmssFlexID)
+			vmssFlex := value.(*compute.VirtualMachineScaleSet)
+			vmssPrefix := to.String(vmssFlex.Name)
+			if vmssFlex.VirtualMachineProfile != nil &&
+				vmssFlex.VirtualMachineProfile.OsProfile != nil &&
+				vmssFlex.VirtualMachineProfile.OsProfile.ComputerNamePrefix != nil {
+				vmssPrefix = to.String(vmssFlex.VirtualMachineProfile.OsProfile.ComputerNamePrefix)
+			}
+			if strings.EqualFold(vmssPrefix, nodeName[:len(nodeName)-6]) {
+				// we should check this vmss first since nodeName and vmssFlex.Name or
+				// ComputerNamePrefix belongs to same vmss, so prepend here
+				vmssFlexIDs = append([]string{vmssFlexID}, vmssFlexIDs...)
+			} else {
+				vmssFlexIDs = append(vmssFlexIDs, vmssFlexID)
 			}
 			return true
 		})
 
-		cachedVmssFlexID, isCached = fs.vmssFlexVMNameToVmssID.Load(nodeName)
-		if isCached {
-			return fmt.Sprintf("%v", cachedVmssFlexID), nil
+		for _, vmssID := range vmssFlexIDs {
+			if _, err := fs.vmssFlexVMCache.Get(vmssID, azcache.CacheReadTypeForceRefresh); err != nil {
+				klog.Errorf("failed to refresh vmss flex VM cache for vmssFlexID %s", vmssID)
+			}
+			// if the vm is cached stop refreshing
+			cachedVmssFlexID, isCached = fs.vmssFlexVMNameToVmssID.Load(nodeName)
+			if isCached {
+				return fmt.Sprintf("%v", cachedVmssFlexID), nil
+			}
 		}
 		return "", cloudprovider.InstanceNotFound
 	}


andyzhangx:
cc @zmyzheng for code review

pkg/provider/azure_vmssflex_cache.go
```diff
@@ -181,18 +181,28 @@ func (fs *FlexScaleSet) getNodeVmssFlexID(nodeName string) (string, error) {
...
 
+		var vmssFlexIDs []string
 		vmssFlexes.Range(func(key, value interface{}) bool {
 			vmssFlexID := key.(string)
-			_, err := fs.vmssFlexVMCache.Get(vmssFlexID, azcache.CacheReadTypeForceRefresh)
-			if err != nil {
-				klog.Errorf("failed to refresh vmss flex VM cache for vmssFlexID %s", vmssFlexID)
+			vmssFlex := value.(*compute.VirtualMachineScaleSet)
+			if vmssFlex.Name != nil && strings.HasPrefix(*vmssFlex.Name, nodeName[:len(nodeName)-6]) {
+				// we should check this vmss first so prepend
```
andyzhangx, line 189: comment like this is better :
```
// we should check this vmss first since nodeName and vmssFlex.Name belongs to same vmss, so prepend here
```

andyzhangx (commented):
/ok-to-test

zmyzheng:
Hi @gcampbell12 , thanks for providing this PR. We did considered similar approach before. However, we figured out there is no guarantee that the VMSS Flex VM has VMSS name as prefix, especially when the VM is created independently and attached to an existing VMSS Flex. Because of this, refreshing the cache of all the VMSS Flex is all what we can do for now. We have raised this issue to the VMSS Flex and they are thinking about how to optimize it.

The throttling issue you observed in Disk attach/detach should not be because of refreshing all the VMSS Flex's caches. Cloud Controller Manager uses this cache a lot more often than CSI Driver and we don't see throttling issue in Cloud Controller Manager. @andyzhangx , I remember you had a PR to fix the throttling in CSI: https://github.com/kubernetes-sigs/cloud-provider-azure/pull/2816. Is this PR fix the cache refreshing issue?

andyzhangx:
> Hi @gcampbell12 , thanks for providing this PR. We did considered similar approach before. However, we figured out there is no guarantee that the VMSS Flex VM has VMSS name as prefix, especially when the VM is created independently and attached to an existing VMSS Flex. Because of this, refreshing the cache of all the VMSS Flex is all what we can do for now. We have raised this issue to the VMSS Flex and they are thinking about how to optimize it.
> 
> The throttling issue you observed in Disk attach/detach should not be because of refreshing all the VMSS Flex's caches. Cloud Controller Manager uses this cache a lot more often than CSI Driver and we don't see throttling issue in Cloud Controller Manager. @andyzhangx , I remember you had a PR to fix the throttling in CSI: #2816. Is this PR fix the cache refreshing issue?

@zmyzheng I think this PR just changed the list sequence to make cache hit earlier, it did not change the behavior, I think that's an optimization? And this PR(https://github.com/kubernetes-sigs/cloud-provider-azure/pull/2950) would fix the VM **list** throttling issue, since would update `vmssFlexVMCache` by VM property returned from VM update operation, so cache is always hit unless it's expired after 10min.

zmyzheng:
> > Hi @gcampbell12 , thanks for providing this PR. We did considered similar approach before. However, we figured out there is no guarantee that the VMSS Flex VM has VMSS name as prefix, especially when the VM is created independently and attached to an existing VMSS Flex. Because of this, refreshing the cache of all the VMSS Flex is all what we can do for now. We have raised this issue to the VMSS Flex and they are thinking about how to optimize it.
> > The throttling issue you observed in Disk attach/detach should not be because of refreshing all the VMSS Flex's caches. Cloud Controller Manager uses this cache a lot more often than CSI Driver and we don't see throttling issue in Cloud Controller Manager. @andyzhangx , I remember you had a PR to fix the throttling in CSI: #2816. Is this PR fix the cache refreshing issue?
> 
> @zmyzheng I think this PR just changed the list sequence to make cache hit earlier, it did not change the behavior, that's an optimization? And this PR(#2950) would fix the VM **list** throttling issue, since it would update `vmssFlexVMCache` by VM property returned from VM update operation, so cache is always hit unless it's expired after 10min.

The issue of this PR is, it tries to only refresh the cache of the VMSS  this VM belongs to by using the name prefix. However, this Prefix is not guaranteed

pkg/provider/azure_vmssflex_cache.go
```diff
@@ -181,18 +181,29 @@ func (fs *FlexScaleSet) getNodeVmssFlexID(nodeName string) (string, error) {
...
 		vmssFlexes := cached.(*sync.Map)
 
+		var vmssFlexIDs []string
 		vmssFlexes.Range(func(key, value interface{}) bool {
 			vmssFlexID := key.(string)
-			_, err := fs.vmssFlexVMCache.Get(vmssFlexID, azcache.CacheReadTypeForceRefresh)
-			if err != nil {
-				klog.Errorf("failed to refresh vmss flex VM cache for vmssFlexID %s", vmssFlexID)
+			vmssFlex := value.(*compute.VirtualMachineScaleSet)
+			if vmssFlex.Name != nil && strings.HasPrefix(*vmssFlex.Name, nodeName[:len(nodeName)-6]) {
```
zmyzheng, line 188: A small suggestion: It is possible the vmssflex vm prefix is different from vmss name as the user can define a customized prefix. A more secure way is as below:
vmssPrefix := *v.VMSS.Name
if v.VMSS.VirtualMachineProfile != nil &&
v.VMSS.VirtualMachineProfile.OsProfile != nil &&
v.VMSS.VirtualMachineProfile.OsProfile.ComputerNamePrefix != nil {
vmssPrefix = *v.VMSS.VirtualMachineProfile.OsProfile.ComputerNamePrefix
}
if strings.EqualFold(vmssPrefix, nodeName[:len(nodeName)-6]) { 
...
}

reference: https://github.com/kubernetes-sigs/cloud-provider-azure/blob/master/pkg/provider/azure_vmss.go#L800~L807

gcampbell12:
@andyzhangx the main behaviour change here is that we return early once we find a matching VM and the if we're using VMs with the vmss prefix the VMSS list that is ranged over is sorted accordingly. 

If for instance you have a stateful set with 10 replicas each using a PVC and 30 VMSS currently on attach you'd end up calling `ListVmssFlexVMsWithoutInstanceView` and `ListVmssFlexVMsWithOnlyInstanceView` 300 times which is likely to exhaust the `Microsoft.Compute/HighCostGet3Min` limit. This change makes it so if your vm is using the vmss prefix you'd only make 10 calls as you should just be refreshing the first vmss ID in the `vmssFlexIDs` slice.

pkg/provider/azure_vmssflex_cache.go
```diff
@@ -181,18 +181,35 @@ func (fs *FlexScaleSet) getNodeVmssFlexID(nodeName string) (string, error) {
...
 		vmssFlexes := cached.(*sync.Map)
 
+		var vmssFlexIDs []string
 		vmssFlexes.Range(func(key, value interface{}) bool {
 			vmssFlexID := key.(string)
-			_, err := fs.vmssFlexVMCache.Get(vmssFlexID, azcache.CacheReadTypeForceRefresh)
-			if err != nil {
-				klog.Errorf("failed to refresh vmss flex VM cache for vmssFlexID %s", vmssFlexID)
+			vmssFlex := value.(*compute.VirtualMachineScaleSet)
+			vmssPrefix := *vmssFlex.Name
```
andyzhangx, line 188: use `to.String(vmssFlex.Name)` that could avoid panic

pkg/provider/azure_vmssflex_cache.go
```diff
@@ -181,18 +181,35 @@ func (fs *FlexScaleSet) getNodeVmssFlexID(nodeName string) (string, error) {
...
+			if vmssFlex.VirtualMachineProfile != nil &&
+				vmssFlex.VirtualMachineProfile.OsProfile != nil &&
+				vmssFlex.VirtualMachineProfile.OsProfile.ComputerNamePrefix != nil {
+				vmssPrefix = *vmssFlex.VirtualMachineProfile.OsProfile.ComputerNamePrefix
```
andyzhangx, line 192: use `to.String(vmssFlex.VirtualMachineProfile.OsProfile.ComputerNamePrefix)` and you don't need `vmssFlex.VirtualMachineProfile.OsProfile.ComputerNamePrefix != nil` check again.

andyzhangx (commented):
pls also squash all commits when your PR is ready, thanks.

pkg/provider/azure_vmssflex_cache.go
gcampbell12: Done

pkg/provider/azure_vmssflex_cache.go
gcampbell12: Not entirely sure this is what you meant but updated to `vmssPrefix = to.String(vmssFlex.VirtualMachineProfile.OsProfile.ComputerNamePrefix)`, we still have to nil check through the struct pointers

andyzhangx: I mean only this `vmssFlex.VirtualMachineProfile.OsProfile.ComputerNamePrefix != nil ` is not necessary if you use `to.String`

gcampbell12: You still need to nil check that though because if the vmss has an OS profile but is not setting the computer name `to.String` would return an empty string here - in this case you still want to check the VMSS name. Try this
```go
func TestVMSSPrefix(t *testing.T) {
	vmssFlex := &compute.VirtualMachineScaleSet{
		Name: to.StringPtr("foo"),
		VirtualMachineScaleSetProperties: &compute.VirtualMachineScaleSetProperties{
			VirtualMachineProfile: &compute.VirtualMachineScaleSetVMProfile{
				OsProfile: &compute.VirtualMachineScaleSetOSProfile{AdminUsername: to.StringPtr("admin")},
			},
		},
	}
	vmssPrefix := to.String(vmssFlex.Name)
	if vmssFlex.VirtualMachineProfile != nil && vmssFlex.VirtualMachineProfile.OsProfile != nil {
		vmssPrefix = to.String(vmssFlex.VirtualMachineProfile.OsProfile.ComputerNamePrefix)
	}
	assert.NotEmpty(t, vmssPrefix)
}
```

andyzhangx (approved):
/lgtm
/approve

zmyzheng (approved):

PR #2987, opened 2022-12-22
fix: support custom storage endpoint suffix

cvvz:

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/pkg/provider/azure_storageaccount.go
+++ b/pkg/provider/azure_storageaccount.go
@@ -37,7 +37,8 @@ import (
 const SkipMatchingTag = "skip-matching"
 const LocationGlobal = "global"
 const GroupIDFile = "file"
-const privateDNSZonePrefix = "privatelink.file."
+
+var privateDNSZoneName = "privatelink.file.%s"
 
 // AccountOptions contains the fields which are used to create storage account.
 type AccountOptions struct {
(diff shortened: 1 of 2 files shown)


pkg/provider/azure_storageaccount.go
```diff
@@ -160,6 +162,14 @@ func (az *Cloud) EnsureStorageAccount(ctx context.Context, accountOptions *Accou
 		subsID = accountOptions.SubscriptionID
 	}
 
+	if len(accountOptions.StorageEndpointSuffix) == 0 {
+		accountOptions.StorageEndpointSuffix = az.cloud.Environment.StorageEndpointSuffix
+	}
+
+	if accountOptions.CreatePrivateEndpoint {
+		privateDNSZoneName = fmt.Sprintf(privateDNSZoneName, accountOptions.StorageEndpointSuffix)
```
andyzhangx, line 170: don't use same var, otherwise `privateDNSZoneName` would be appened again and again

andyzhangx (commented):
/ok-to-test

andyzhangx (approved):
/lgtm

andyzhangx (commented):
/hold

andyzhangx (commented):
I think it's better generate `privateDNSZoneName` when it's needed, use a global env var may not be a good idea since file and blob use different privateDNSZoneName.

cvvz:
> I think it's better generate `privateDNSZoneName` when it's needed, use a global env var may not be a good idea since file and blob use different privateDNSZoneName.

I have removed the global env var, I agree that we should avoid using global variables. `privateDNSZoneName` can be assembled at the beginning of `EnsureStorageAccout` then pass it as function parameter to avoid assembling it everytime when using it.

I will introduce `StorageType` in this [PR](https://github.com/kubernetes-sigs/cloud-provider-azure/pull/2992/files) to support blob csi driver.

andyzhangx (approved):
/lgtm
/hold canel

PR #2992, opened 2022-12-22
feat: Support privateendpoint for blob-csi-driver

cvvz:

Fixes #

```release-note
feat: Support private endpoint for blob-csi-driver
```

```docs

```

### DIFF
--- a/pkg/provider/azure_storageaccount.go
+++ b/pkg/provider/azure_storageaccount.go
@@ -37,7 +37,15 @@ import (
 const SkipMatchingTag = "skip-matching"
 const LocationGlobal = "global"
 const GroupIDFile = "file"
-const privateDNSZoneNameFmt = "privatelink.file.%s"
+const GroupIDBlob = "blob"
+const privateDNSZoneNameFmt = "privatelink.%s.%s"
+
+type StorageType string
+
+const (
+	StorageTypeBlob StorageType = "blob"
+	StorageTypeFile StorageType = "file"
+)
 
 // AccountOptions contains the fields which are used to create storage account.
 type AccountOptions struct {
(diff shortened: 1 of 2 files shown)


pkg/provider/azure_storageaccount.go
```diff
@@ -160,6 +170,17 @@ func (az *Cloud) EnsureStorageAccount(ctx context.Context, accountOptions *Accou
 		subsID = accountOptions.SubscriptionID
 	}
 
+	if len(accountOptions.StorageEndpointSuffix) == 0 {
+		accountOptions.StorageEndpointSuffix = az.cloud.Environment.StorageEndpointSuffix
+	}
+
+	if accountOptions.CreatePrivateEndpoint {
+		if accountOptions.StorageType == "" {
```
andyzhangx, line 178: to make it backward compatible, if `StorageType == ""`, it should be `file` by default.

pkg/provider/azure_storageaccount.go
```diff
@@ -405,7 +425,7 @@ func (az *Cloud) createPrivateEndpoint(ctx context.Context, accountName string,
 	privateLinkServiceConnection := network.PrivateLinkServiceConnection{
 		Name: &privateLinkServiceConnectionName,
 		PrivateLinkServiceConnectionProperties: &network.PrivateLinkServiceConnectionProperties{
-			GroupIds:             &[]string{GroupIDFile},
```
andyzhangx, line 408: what about setting `GroupIds` as both `file`, `blob`, by that way, this PrivateLinkServiceConnection could be used for both  blob and file? that may be more straightforward

andyzhangx (approved):
/lgtm
/kind feat
pls also add release note in the PR description.

PR #2998, opened 2022-12-26
feat: support private endpoint for blob and file csi driver

cvvz:

Fixes #

```release-note
feat: support private endpoint for blob and file csi driver
```

```docs

```


pkg/provider/azure_storageaccount.go
```diff
@@ -271,9 +273,6 @@ func (az *Cloud) EnsureStorageAccount(ctx context.Context, accountOptions *Accou
 			}
 		}
 
-		if location == "" {
```
andyzhangx, line 274: why remove this?

andyzhangx (commented):
/ok-to-test

andyzhangx (commented):
pls also add ut for this PR

pkg/provider/azure_storageaccount.go
cvvz: move it to [here](https://github.com/kubernetes-sigs/cloud-provider-azure/pull/2998/files#diff-43399bb105b7d123e4c10ca28097886143c41c198bdd8189be09eba0a5d70fdaR171) since we need to use it at [here](https://github.com/kubernetes-sigs/cloud-provider-azure/pull/2998/files#diff-43399bb105b7d123e4c10ca28097886143c41c198bdd8189be09eba0a5d70fdaR390)

pkg/provider/azure_storageaccount.go
```diff
@@ -462,7 +470,7 @@ func (az *Cloud) createVNetLink(ctx context.Context, vNetLinkName, vnetResourceG
 		Location: &location,
 		VirtualNetworkLinkProperties: &privatedns.VirtualNetworkLinkProperties{
 			VirtualNetwork:      &privatedns.SubResource{ID: &vnetID},
-			RegistrationEnabled: to.BoolPtr(true)},
+			RegistrationEnabled: to.BoolPtr(false)},
```
andyzhangx, line 473: why make this change?

cvvz: According to the [document](https://learn.microsoft.com/en-us/azure/dns/private-dns-virtual-network-links#registration-virtual-network), every virtual network can only have one registration zone associated with it. Since we need to create two   virtual network link, we cannot set `RegistrationEnabled` as true for both of them. 

And auto-registration is used for virtual machine: `A DNS record gets automatically created for any virtual machines you deploy in the virtual network.` So no need to set it as true.

andyzhangx (approved):
/lgtm

PR #3070, opened 2023-01-06
Set --skip-nodes-with-system-pods to false when running autoscaling tests

lzhecheng:

#### What type of PR is this?
/kind testing

Set --skip-nodes-with-system-pods to false when running autoscaling tests
Currently without this option (default: true), it is possible that some deployments like metrics-server will be allocated to the new Node. In this case, when scaling down, no Nodes can be deleted.
https://github.com/Azure/AKS/issues/875#issuecomment-475486313

Fixes #

```release-note
NONE
```

```docs

```


feiskyer (commented):
/lgtm

PR #3084, opened 2023-01-09
fix: update rather than replace existing cache entry

edreed:
#### What type of PR is this?

/kind bug

Since `TimedCache` relies on a per-entry `sync.Mutex` to serialize calls to the refresh function, `TimedCache.Update` **must** update the existing entry rather than replace it to ensure proper synchronization. Otherwise, it is possible for multiple concurrent calls to the refresh function to occurr.

Fixes #

```release-note
Fixes an issue in the `TimedCache` implementation that could result in multiple concurrent calls to the getter function.
```

```docs

```

### DIFF
--- a/pkg/cache/azure_cache.go
+++ b/pkg/cache/azure_cache.go
@@ -189,9 +189,16 @@ func (t *TimedCache) Set(key string, data interface{}) {
 
 // Update updates the data cache for the key.
 func (t *TimedCache) Update(key string, data interface{}) {
-	_ = t.Store.Update(&AzureCacheEntry{
-		Key:       key,
-		Data:      data,
-		CreatedOn: time.Now().UTC(),
-	})
+	if entry, err := t.getInternal(key); err == nil {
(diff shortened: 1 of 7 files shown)


pkg/cache/azure_cache.go
```diff
@@ -189,9 +189,16 @@ func (t *TimedCache) Set(key string, data interface{}) {
 
 // Update updates the data cache for the key.
 func (t *TimedCache) Update(key string, data interface{}) {
-	_ = t.Store.Update(&AzureCacheEntry{
```
feiskyer, line 192: Could you add some unit tests to catch this condition?

edreed: Done.

andyzhangx:
@edreed could you fix the lint issue? after this PR merged, I will publish a new release late this week, thanks.

andyzhangx (approved):
/lgtm

PR #3115, opened 2023-01-13
feat: support data protection for blob csi driver

cvvz:
#### What type of PR is this?
/kind feature

support data protection for blob csi driver

Fixes #

```release-note
support data protection for blob csi driver
```

```docs

```

### DIFF
--- a/pkg/provider/azure_storageaccount.go
+++ b/pkg/provider/azure_storageaccount.go
@@ -364,6 +364,8 @@ func (az *Cloud) EnsureStorageAccount(ctx context.Context, accountOptions *Accou
 			accountOptions.SoftDeleteBlobs > 0 ||
 			accountOptions.SoftDeleteContainers > 0 {
 			var blobPolicy, containerPolicy *storage.DeleteRetentionPolicy
+			var enableBlobVersioning *bool
+
 			if accountOptions.SoftDeleteContainers > 0 {
 				containerPolicy = &storage.DeleteRetentionPolicy{
 					Enabled: pointer.Bool(accountOptions.SoftDeleteContainers > 0),
@@ -377,9 +379,13 @@ func (az *Cloud) EnsureStorageAccount(ctx context.Context, accountOptions *Accou
 				}
 			}
 
+			if accountOptions.EnableBlobVersioning != nil {
+				enableBlobVersioning = pointer.Bool(*accountOptions.EnableBlobVersioning)
+			}
+
 			property := storage.BlobServiceProperties{
 				BlobServicePropertiesProperties: &storage.BlobServicePropertiesProperties{
-					IsVersioningEnabled:            pointer.Bool(pointer.BoolDeref(accountOptions.EnableBlobVersioning, false)),
+					IsVersioningEnabled:            enableBlobVersioning,
 					ContainerDeleteRetentionPolicy: containerPolicy,
 					DeleteRetentionPolicy:          blobPolicy,
 				},
--- a/pkg/provider/azure_storageaccount.go
+++ b/pkg/provider/azure_storageaccount.go
@@ -360,7 +360,9 @@ func (az *Cloud) EnsureStorageAccount(ctx context.Context, accountOptions *Accou
 			return "", "", fmt.Errorf("failed to create storage account %s, error: %v", accountName, rerr)
 		}
 
-		if pointer.BoolDeref(accountOptions.EnableBlobVersioning, false) || accountOptions.SoftDeleteBlobs > 0 || accountOptions.SoftDeleteContainers > 0 {
+		if pointer.BoolDeref(accountOptions.EnableBlobVersioning, false) ||
+			accountOptions.SoftDeleteBlobs > 0 ||
+			accountOptions.SoftDeleteContainers > 0 {
 			var blobPolicy, containerPolicy *storage.DeleteRetentionPolicy
 			if accountOptions.SoftDeleteContainers > 0 {
 				containerPolicy = &storage.DeleteRetentionPolicy{
@@ -377,7 +379,7 @@ func (az *Cloud) EnsureStorageAccount(ctx context.Context, accountOptions *Accou
 
 			property := storage.BlobServiceProperties{
 				BlobServicePropertiesProperties: &storage.BlobServicePropertiesProperties{
-					IsVersioningEnabled:            accountOptions.EnableBlobVersioning,
+					IsVersioningEnabled:            pointer.Bool(pointer.BoolDeref(accountOptions.EnableBlobVersioning, false)),
 					ContainerDeleteRetentionPolicy: containerPolicy,
 					DeleteRetentionPolicy:          blobPolicy,
 				},
--- a/pkg/provider/azure_storageaccount.go
+++ b/pkg/provider/azure_storageaccount.go
@@ -72,6 +72,9 @@ type AccountOptions struct {
 	SubnetName                              string
 	AccessTier                              string
 	MatchTags                               bool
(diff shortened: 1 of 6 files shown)


cvvz:
@andyzhangx PTAL

pkg/provider/azure_storageaccount.go
```diff
@@ -72,6 +72,9 @@ type AccountOptions struct {
 	SubnetName                              string
 	AccessTier                              string
 	MatchTags                               bool
+	EnableBlobVersioning                    bool
```
andyzhangx, line 75: it's better using `*bool`, then it would have 3 values: true, false, nil, otherwise you cannot distinguish false and nil condition

andyzhangx:
> @cvvz: Adding label `do-not-merge/contains-merge-commits` because PR contains merge commits, which are not allowed in this repository. Use `git rebase` to reapply your commits on top of the target branch. Detailed instructions for doing so can be found [here](https://git.k8s.io/community/contributors/guide/github-workflow.md#4-keep-your-branch-in-sync).

@cvvz you could use `git rebase upstream/master` to rebase to master branch, then you won't have this warning message.

pkg/provider/azure_storageaccount.go
```diff
@@ -355,6 +369,34 @@ func (az *Cloud) EnsureStorageAccount(ctx context.Context, accountOptions *Accou
...
+
+			property := storage.BlobServiceProperties{
+				BlobServicePropertiesProperties: &storage.BlobServicePropertiesProperties{
+					IsVersioningEnabled:            accountOptions.EnableBlobVersioning,
+					ContainerDeleteRetentionPolicy: containerPolicy,
+					DeleteRetentionPolicy:          blobPolicy,
+				},
+			}
+
+			if _, err := az.BlobClient.SetServiceProperties(ctx, subsID, resourceGroup, accountName, property); err != nil {
```
andyzhangx, line 395: if `SetServiceProperties` failed, then `StorageAccountClient.Delete` account, that could avoid creating lots of empty storage accounts.

pkg/provider/azure_storageaccount.go
```diff
@@ -109,6 +112,17 @@ func (az *Cloud) getStorageAccounts(ctx context.Context, accountOptions *Account
 				isPrivateEndpointAsExpected(acct, accountOptions)) {
 				continue
 			}
+
+			property, err := az.BlobClient.GetServiceProperties(ctx, accountOptions.SubscriptionID, accountOptions.ResourceGroup, *acct.Name)
```
andyzhangx, line 116: this is calling `GetServiceProperties` for every account search, that's expensive, it's better if only BlobVersioning is enabled, then call `GetServiceProperties`

pkg/provider/azure_storageaccount.go
```diff
@@ -354,6 +368,42 @@ func (az *Cloud) EnsureStorageAccount(ctx context.Context, accountOptions *Accou
 		if rerr := az.StorageAccountClient.Create(ctx, subsID, resourceGroup, accountName, cp); rerr != nil {
 			return "", "", fmt.Errorf("failed to create storage account %s, error: %v", accountName, rerr)
 		}
+		defer func() {
+			if retErr != nil {
+				klog.V(2).Infof("delete storage account %s because error happened after storage account created: %v", accountName, retErr)
```
andyzhangx, line 373: `klog.Errorf`

pkg/provider/azure_storageaccount.go
```diff
@@ -354,6 +368,42 @@ func (az *Cloud) EnsureStorageAccount(ctx context.Context, accountOptions *Accou
 		if rerr := az.StorageAccountClient.Create(ctx, subsID, resourceGroup, accountName, cp); rerr != nil {
 			return "", "", fmt.Errorf("failed to create storage account %s, error: %v", accountName, rerr)
 		}
+		defer func() {
+			if retErr != nil {
```
andyzhangx, line 372: not quite sure whether `retErr` would be equal to `fmt.Errorf("failed to set blob service properties for storage account %s, error: %w", accountName, err)`

pkg/provider/azure_storageaccount.go
cvvz: here we need to search an existed storageaccount which has the same property value of EnableBlobVersion(enable/disable) as specified by storageclass.

pkg/provider/azure_storageaccount.go
cvvz: let me do some manual test

cvvz: image

image

It worked, there was not large amount of storageaccounts in the resource group.

pkg/provider/azure_storageaccount.go
andyzhangx: since `ServiceProperties` is only used in these 3 functions, I think it would be better to only call if user enables `SoftDelete` functions, otherwise that's two expensive call in every account matching.

pkg/provider/azure_storageaccount.go
```diff
@@ -72,6 +72,9 @@ type AccountOptions struct {
 	SubnetName                              string
 	AccessTier                              string
 	MatchTags                               bool
+	EnableBlobVersioning                    *bool
```
andyzhangx, line 75: since we already have `SoftDeleteBlobs`, `SoftDeleteContainers`, do we need this `EnableBlobVersioning` field?

cvvz: I think so. Blob Versioning is similar to snapshot which makes customer able to rollback to a specific version.

pkg/provider/azure_storageaccount.go
```diff
@@ -355,6 +360,34 @@ func (az *Cloud) EnsureStorageAccount(ctx context.Context, accountOptions *Accou
...
+			if accountOptions.SoftDeleteBlobs > 0 {
+				blobPolicy = &storage.DeleteRetentionPolicy{
+					Enabled: pointer.Bool(accountOptions.SoftDeleteBlobs > 0),
+					Days:    pointer.Int32(accountOptions.SoftDeleteBlobs),
+				}
+			}
+
+			property := storage.BlobServiceProperties{
+				BlobServicePropertiesProperties: &storage.BlobServicePropertiesProperties{
+					IsVersioningEnabled:            accountOptions.EnableBlobVersioning,
```
andyzhangx, line 380: what if only `accountOptions.EnableBlobVersioning` is set while other two values are nil? it's still supported?

cvvz: Yes, they are different features and these three properties can be set seperately.
image

andyzhangx, line 380: avoid using `*bool` directly since next time if `accountOptions.EnableBlobVersioning` changes, it may change, define a new bool var here is safer.

cvvz: Do you mean we should copy value from `accountOptions.EnableBlobVersioning` rather then pass the pointer itself directly?

andyzhangx: like this, it's safer:
```
value := *accountOptions.EnableBlobVersioning
IsVersioningEnabled: &value
```

andyzhangx:
```
var value *bool
if accountOptions.EnableBlobVersioning != nil {
  value = *accountOptions.EnableBlobVersioning
}
...
IsVersioningEnabled: &value
```

cvvz: ok

andyzhangx (approved):
/lgtm

PR #3281, opened 2023-02-03
helm chart 1.26.4 with latest releases

jackfrancis:
#### What this PR does / why we need it:

This PR updates the helm chart with the latest releases.

#### Which issue(s) this PR fixes:

Fixes #

#### Does this PR introduce a user-facing change?

```release-note

```

#### Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.:

```docs

```

PR #3283, opened 2023-02-03
Use hostprocess for Windows node manager in Helm chart

CecileRobertMichon:
#### What type of PR is this?

/kind bug

Currently, external cloud-provider does not work with Windows nodes when using the Helm chart. The Daemonset for cloud-node-manager-windows is not able to reach the metadata endpoint and therefore windows nodes always have the Taint node.cloudprovider.kubernetes.io/uninitialized=true:NoSchedule.

This PR changes the cloud-node-manager-windows Daemonset to use `hostProcess: true` and adds a temporary workaround for not being able to use `inClusterConfig()` (until k8s 1.26 + containerd v1.7).

All credits go to @mweibel and @marosset for discovering the issue and suggesting a fix.

Fixes https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/2591

```release-note
NONE
```

```docs

```


helm/cloud-provider-azure/templates/cloud-provider-azure.yaml
```diff
@@ -3,13 +3,13 @@
...
+  {{- $_ := set . "cloudProviderAzureVersion" "v1.23.27" }}
 {{- else if eq .Capabilities.KubeVersion.Minor "24" -}}
-  {{- $_ := set . "cloudProviderAzureVersion" "v1.24.11" }}
+  {{- $_ := set . "cloudProviderAzureVersion" "v1.24.14" }}
 {{- else if eq .Capabilities.KubeVersion.Minor "25" -}}
-  {{- $_ := set . "cloudProviderAzureVersion" "v1.25.5" }}
+  {{- $_ := set . "cloudProviderAzureVersion" "v1.25.8" }}
 {{- else if eq .Capabilities.KubeVersion.Minor "26" -}}
-  {{- $_ := set . "cloudProviderAzureVersion" "v1.26.0" }}
+  {{- $_ := set . "cloudProviderAzureVersion" "v1.26.4" }}
```
CecileRobertMichon, line 12: will rebase once #3281 merges

CecileRobertMichon: done

feiskyer:
> the Daemonset for cloud-node-manager-windows is not able to reach the metadata endpoint 

Could you elaborate why IMDS is not reachable on capz?

helm/cloud-provider-azure/templates/cloud-provider-azure.yaml
```diff
@@ -428,6 +428,21 @@ spec:
...
+metadata:
+  name: cloud-provider-azure-scripts
+  namespace: kube-system
+  labels:
+    app: cloud-provider-azure-windows
+data:
+  start.ps1: |
+    $cm = "$env:CONTAINER_SANDBOX_MOUNT_POINT/cloud-node-manager.exe"
+
+    ((Get-Content -path $env:CONTAINER_SANDBOX_MOUNT_POINT/var/lib/cpaw/kubeconfig.conf -Raw) -replace '/var',"$($env:CONTAINER_SANDBOX_MOUNT_POINT)/var") | Set-Content -Path $env:CONTAINER_SANDBOX_MOUNT_POINT/var/lib/cpaw/kubeconfig.conf
```
feiskyer, line 442: Could the in-cluster authz with service account work here? If so, there is no need of the script

marosset: In-cluster auth with service accounts works for hostprocess containers with containerd v1.7 only (not v1.6.x).
The way volume mounts get set up was changed between containerd 1.6 and 1.7

CecileRobertMichon: yes this is a workaround until containerd 1.7 is released, after that we can remove the script

CecileRobertMichon:
> Could you elaborate why IMDS is not reachable on capz?

@marosset can weigh in but I believe it's not a capz limitation, it's a calico overlay networking issue: https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/2132

Azure CSI driver helm charts also had to do this for this reason: https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/1201

marosset (approved):
/lgtm

This should unblock deploying the Windows cloud-node-manager!
We can update the deployment to no longer need the powershell / kube-config after containerd 1.7 releases (but the current changes should work on both containerd 1.6 and 1.7)

helm/cloud-provider-azure/templates/cloud-provider-azure.yaml
feiskyer: then let's wait for 1.7?

CecileRobertMichon: it's blocking https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3105 

we want to move all CAPZ templates and tests to external cloud-provider to get as much test coverage as possible before in-tree cloud-providers are completely removed

feiskyer: I see. Could you add a TODO comment with containerd issue link?

marosset: Here is the issue https://github.com/kubernetes/kubernetes/issues/104562

CecileRobertMichon: Added

feiskyer (approved):
/lgtm

PR #3289, opened 2023-02-07
Fix how getSecurityRuleName() handles IPv6 addr prefix

lzhecheng:

#### What type of PR is this?
/kind bug

Fix how getSecurityRuleName() handles IPv6 addr prefix
Error msg is like: `Resource name a29b3b265d9f84776b6c877408e7a519-TCP-80-2001:1234:5678:9abd::4_128 is invalid. The name can be up to 80 characters long. It must begin with a word character, and it must end with a word character or with '_'. The name may contain word characters or '.', '-', '_'."`
* Bug fix
* Add UT and adjust e2e

Fixes #

```release-note
Fix incorrect security rule name with IPv6 address prefix
```

```docs

```

### DIFF
--- a/pkg/provider/azure_standard.go
+++ b/pkg/provider/azure_standard.go
@@ -292,11 +292,11 @@ func (az *Cloud) getloadbalancerHAmodeRuleName(service *v1.Service) string {
 }
 
 func (az *Cloud) getSecurityRuleName(service *v1.Service, port v1.ServicePort, sourceAddrPrefix string) string {
+	safePrefix := strings.Replace(sourceAddrPrefix, "/", "_", -1)
+	safePrefix = strings.Replace(safePrefix, ":", ".", -1) // Consider IPv6 address
 	if useSharedSecurityRule(service) {
-		safePrefix := strings.Replace(sourceAddrPrefix, "/", "_", -1)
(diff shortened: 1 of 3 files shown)


pkg/provider/azure_standard.go
```diff
@@ -292,11 +292,11 @@ func (az *Cloud) getloadbalancerHAmodeRuleName(service *v1.Service) string {
 }
 
 func (az *Cloud) getSecurityRuleName(service *v1.Service, port v1.ServicePort, sourceAddrPrefix string) string {
+	safePrefix := strings.Replace(sourceAddrPrefix, "/", "_", -1)
+	safePrefix = strings.Replace(safePrefix, ":", "_", -1) // Consider IPv6 address
```
jwtty, line 296: Is it possible to get same rule name with different IPs:
e.g. both "fe80::1/32" and "fe80::1:32" could be translated to "fe80__1_32"

lzhecheng: Currently, such problem won't happen because `sourceAddrPrefix` is an IPNet and it has a mask.
However, `.` can be used instead because it does no harm and it is possible someday in the future, someone changes `sourceAddrPrefix` but forget to update here. Thank you.

feiskyer:
Good catch and thanks for the fix

/lgtm

PR #3482, opened 2023-03-03
fix: detach disk failure when there is throttling

andyzhangx:
#### What type of PR is this?
/kind bug

fix: detach disk failure when there is throttling error

This PR would always check disk lun after disk detach complete, that could avoid detach disk failure when there is throttling error in batch operation, and I have run 100+ 1K disk attach/detach tests with throttling errors, all disk detach completes successfully!

Fixes #

```release-note
fix: detach disk failure when there is throttling
```

```docs
fix: detach disk failure when there is throttling
```


ZeroMagic (approved):
/lgtm

ZeroMagic (approved):
/lgtm

PR #3517, opened 2023-03-09
Change MachinePool to use extenral cloud-provider

zmyzheng:
#### What type of PR is this?
/kind bug

When setting "cloud-provider: azure", the CAPZ will use in-tree kube-controller-manager rather than out-of-tree CCM and CNM. This will cause the ndoe providerID is set with wrong format for VMSS Flex VMs

Fixes #

```release-note
NONE
```

```docs

```


lzhecheng:
The checks don't check changes in templates and it will fail because of current incorrect template in master branch.

PR #3575, opened 2023-03-16
[e2e] Delete ns and client if not nil to avoid panic

lzhecheng:
#### What type of PR is this?
/kind testing

[e2e] Delete ns and client if not nil to avoid panic

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/tests/e2e/network/private_link_service.go
+++ b/tests/e2e/network/private_link_service.go
@@ -80,11 +80,13 @@ var _ = Describe("Private link service", Label(utils.TestSuiteLabelPrivateLinkSe
 	})
 
 	AfterEach(func() {
-		err := cs.AppsV1().Deployments(ns.Name).Delete(context.TODO(), serviceName, metav1.DeleteOptions{})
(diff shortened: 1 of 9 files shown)


tests/e2e/network/private_link_service.go
```diff
@@ -59,27 +59,36 @@ var _ = Describe("Private link service", Label(utils.TestSuiteLabelPrivateLinkSe
...
 	})
 
 	AfterEach(func() {
+		if !strings.EqualFold(os.Getenv(utils.LoadBalancerSkuEnv), string(network.PublicIPAddressSkuNameStandard)) {
+			Skip("private link service only works with standard load balancer")
+		}
+
+		if tc.IPFamily != utils.IPv4 {
+			tc = nil
+			return
```
jwtty, line 89: Why return instead of skip here?

lzhecheng: `Skip` is better. Updated.

PR #3643, opened 2023-03-24
[WIP] Fix incorrect order of Node internal address issue

### LINKED ISSUE #3401: ipv6 + Calico + out of tree cloud-provider is broken
While moving all CAPZ templates to out-of-tree CCM, I was not able to get a reliably passing test for ipv6. This [test](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/3105/pull-cluster-api-provider-azure-e2e/1627421299045306368) is very flaky (it fails most of the time). Looking into it I found that the error is that kubelet can't patch the status of one of the kube-system pods (sometimes kube-apiserver, sometimes kube-scheduler, sometimes kube-controller-manager). One example of the error can be seen [here](https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/3105/pull-cluster-api-provider-azure-e2e/1627421299045306368/artifacts/clusters/capz-e2e-4laer3-ipv6/machines/capz-e2e-4laer3-ipv6-control-plane-sj49g/kubelet.log). The error is 
```
	Pod \"kube-controller-manager-capz-e2e-4laer3-ipv6-control-plane-6v7dn\" is invalid: [status.podIPs: Invalid value: []core.PodIP{core.PodIP{IP:\"2001:1234:5678:9abc::5\"}, core.PodIP{IP:\"2001:1234:5678:9abc::5\"}}: may specify no more than one IP for each IP family, status.podIPs[1]: Duplicate value: core.PodIP{IP:\"2001:1234:5678:9abc::5\"}]"
```

The same error does not repro with the in-tree cloud-provider (see passing ipv6 test with CAPZ and in-tree cloud-provider [here](https://testgrid.k8s.io/sig-cluster-lifecycle-cluster-api-provider-azure#capz-periodic-e2e-main)).

The cloud-provider-azure test for ipv6 is also failing (although the failure is different: https://testgrid.k8s.io/provider-azure-cloud-provider-azure#cloud-provider-azure-master-ipv6-capz)

lzhecheng:
#### What type of PR is this?
/kind bug

Fix incorrect order of Node internal address issue. getLocalInstanceNodeAddresses() always returns IPv4 then IPv6 if Node is dual-stack. The order of 2 internal IPs should be consistent as when first IP is added to the Node.

Fixes #3401

```release-note
Fix incorrect order of Node internal address issue. getLocalInstanceNodeAddresses() always returns IPv4 then IPv6 if Node is dual-stack. The order of 2 internal IPs should be consistent as when first IP is added to the Node.
```

```docs

```

### DIFF
--- a/pkg/nodemanager/nodemanager.go
+++ b/pkg/nodemanager/nodemanager.go
@@ -280,13 +280,52 @@ func (cnc *CloudNodeController) updateNodeAddress(ctx context.Context, node *v1.
 	}
 
 	newNode := node.DeepCopy()
-	newNode.Status.Addresses = nodeAddresses
+	newNode.Status.Addresses = ensureNodeInternalAddressesOrder(node.Status.Addresses, nodeAddresses)
 	_, _, err = PatchNodeStatus(cnc.kubeClient.CoreV1(), types.NodeName(node.Name), node, newNode)
 	if err != nil {
 		return fmt.Errorf("patching node with cloud ip addresses: %w", err)
 	}
 
 	return nil
+}
+
+// ensureNodeInternalAddressesOrder ensures Node internal addresses are consistent in terms of order.
+func ensureNodeInternalAddressesOrder(old, new []v1.NodeAddress) []v1.NodeAddress {
+	ordered := []v1.NodeAddress{}
+	ipv6First := false
+	for _, addr := range old {
+		if addr.Type != v1.NodeInternalIP {
+			continue
+		}
+		if net.ParseIP(addr.Address).To4() == nil {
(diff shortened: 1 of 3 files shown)


lzhecheng:
/close
This PR is not needed. IPv4 first and then IPv6 is the expected behavior from CAPZ side.
Detail: https://github.com/kubernetes-sigs/cloud-provider-azure/issues/3401#issuecomment-1483650755

Issue #3519, opened 2023-03-09
[Refactor] Consider not using suffix for Service annotation azure-load-balancer-ip with IPv4


lzhecheng:
#### What would you like to be added:
Currently, Service annotations for azure LB IP for IPv4 and IPv6 have `-ipv4` and `-ipv6` suffixes, while azure-pip-name, azure-pip-prefix-id and backend pool don't. It is more consistent if suffix can be remove for azure LB IP with IPv4.

#### Why is this needed:
Consistency

nikzayn:
@lzhecheng I would like to take this up!

Could you confirm with me once, here I need to make changes right in consts.go file. right?
```

	ServiceAnnotationLoadBalancerIPDualStack = map[bool]string{
		false: "service.beta.kubernetes.io/azure-load-balancer-ipv4",
		true:  "service.beta.kubernetes.io/azure-load-balancer-ipv6",
	}
	ServiceAnnotationPIPNameDualStack = map[bool]string{
		false: "service.beta.kubernetes.io/azure-pip-name-ipv4",
		true:  "service.beta.kubernetes.io/azure-pip-name-ipv6",
	}
	ServiceAnnotationPIPPrefixIDDualStack = map[bool]string{
		false: "service.beta.kubernetes.io/azure-pip-prefix-id-ipv4",
		true:  "service.beta.kubernetes.io/azure-pip-prefix-id-ipv6",
	}
)
```

/assign

lzhecheng:
@nikzayn thank you! Actually only the LB IP one and the other 2 don't have the issue now. We should make sure the old annotation still works after your change.

nikzayn:
Sorry, I still didn't get it clearly!

lzhecheng:
> Sorry, I still didn't get it clearly!

Sorry for the late reply, I mean:
1. `ServiceAnnotationPIPNameDualStack` and `ServiceAnnotationPIPPrefixIDDualStack` don't need change any more.
2. Since we have already put `service.beta.kubernetes.io/azure-load-balancer-ipv4` into the doc, please make sure it still works when supporting `service.beta.kubernetes.io/azure-load-balancer`.

PR #3826, opened 2023-04-28
[3519] - config refactored for service annotation azure-load-balancer-ip

nikzayn:
#### What type of PR is this?
/kind feature

It maintains consistency with other service annotations

[3519](https://github.com/kubernetes-sigs/cloud-provider-azure/issues/3519)

Please just do sanity check

NONE


lzhecheng:
This is not what we expected. The old annotation should be supported as well.
/hold

PR #3854, opened 2023-05-05
[Refactor] Drop RG param in getXXXID()

lzhecheng:
#### What type of PR is this?
/kind cleanup

[Refactor] Drop RG param in getXXXID(). Resource group name can be got from az.getLoadBalancerResourceGroup().

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/provider/azure_standard.go
+++ b/pkg/provider/azure_standard.go
@@ -69,7 +69,11 @@ func (az *Cloud) getAvailabilitySetID(resourceGroup, availabilitySetName string)
 }
 
 // returns the full identifier of a loadbalancer frontendipconfiguration.
-func (az *Cloud) getFrontendIPConfigID(lbName, rgName, fipConfigName string) string {
+func (az *Cloud) getFrontendIPConfigID(lbName, fipConfigName string) string {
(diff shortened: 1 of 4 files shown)


pkg/provider/azure_standard.go
```diff
@@ -69,31 +69,31 @@ func (az *Cloud) getAvailabilitySetID(resourceGroup, availabilitySetName string)
 }
 
 // returns the full identifier of a loadbalancer frontendipconfiguration.
-func (az *Cloud) getFrontendIPConfigID(lbName, rgName, fipConfigName string) string {
+func (az *Cloud) getFrontendIPConfigID(lbName, fipConfigName string) string {
 	return fmt.Sprintf(
 		consts.FrontendIPConfigIDTemplate,
 		az.getNetworkResourceSubscriptionID(),
-		rgName,
+		az.getLoadBalancerResourceGroup(),
```
MartinForReal, line 76: getFrontendIPConfigID will rely on az config, the function will return different value if the config is changed.

lzhecheng: RG comes from `az.getLoadBalancerResourceGroup()` in all current usages. Added a new feature `XXXWithRG()` to allow using specific RG.
Updated.

MartinForReal: LGTM

PR #3886, opened 2023-05-16
track2 sdk: return error when client received 429 toomanyrequest error

MartinForReal:
#### What type of PR is this?
/kind feature

track2 sdk: return error when client received 429 toomanyrequest error
add tracing utils

Fixes #

```release-note
NONE
```

```docs

```


MartinForReal:
/override  "Header rules"  "Mixed content" "Pages changed" "Redirect rules" "
deploy/netlify"

MartinForReal:
<html>
<body>


/test pull-cloud-provider-azure-e2e-ccm-vmss-capz
--


</body>
</html>

pkg/azclient/policy/retry.go
```diff
@@ -35,7 +35,6 @@ import (
 func GetRetriableStatusCode() []int {
 	return []int{
 		http.StatusRequestTimeout,      // 408
-		http.StatusTooManyRequests,     // 429
```
nilo19, line 38: Do we still retry when hitting 429?

pkg/azclient/utils/tracing.go
```diff
@@ -0,0 +1,134 @@
...
+
+	return ctx, tracing.NewSpan(
+		tracing.SpanImpl{
+			End: func() {
+				span.End()
+			},
+			SetAttributes: func(attributes ...tracing.Attribute) {
+				var traceAttributes []attribute.KeyValue
+				for _, entry := range attributes {
+					traceAttributes = append(traceAttributes, attribute.String(entry.Key, fmt.Sprintf("%v", entry.Value)))
```
nilo19, line 66: Is it possible to avoid using %v in our code?

pkg/azclient/policy/retry.go
MartinForReal: no. the client will return error and let caller decide when to retry. throttle policy will read retry after header and reject any request if the client is throttled.

pkg/azclient/utils/tracing.go
MartinForReal: attribute. Value is type of interface{}. It will throw panic if the object doesn't implement String() func. I agree we should avoid using %v, but here we don't have better option.

MartinForReal:
<html>
<body>


/test pull-cloud-provider-azure-e2e-capz
--


</body>
</html>

PR #4018, opened 2023-06-01
Add concurrent-node-syncs option

vitaliy-sn:
#### What type of PR is this?
/kind feature
/kind bug

This PR:
- add concurrent-node-syncs option

Without this, ConcurrentNodeSyncs is always 0, as a result of which the node controller does not work.

Fixes #

feat: add concurrent-node-syncs option

```release-note

```

```docs

```


feiskyer:
> Without this, ConcurrentNodeSyncs is always 0, as a result of which the node controller does not work.

@vitaliy-sn  could you share more details of the issue? would you like to share the cloud-controller-manager config options and its logs?

PR #4067, opened 2023-06-08
feat: add GetLatestAccountKey in account key fetch

andyzhangx:
#### What type of PR is this?
/kind feature

feat: add GetLatestAccountKey in account key fetch
related to requirement: https://github.com/kubernetes-sigs/azurefile-csi-driver/issues/1218
This PR does not change the default behavior.

Fixes #

```release-note
feat: add GetLatestAccountKey in account key fetch
```

```docs
feat: add GetLatestAccountKey in account key fetch
```

### DIFF
--- a/pkg/provider/azure_storageaccount.go
+++ b/pkg/provider/azure_storageaccount.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"math/big"
 	"strings"
+	"time"
 
 	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2022-07-01/network"
 	"github.com/Azure/azure-sdk-for-go/services/privatedns/mgmt/2018-09-01/privatedns"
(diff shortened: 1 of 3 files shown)


pkg/provider/azure_storageaccount.go
```diff
@@ -138,16 +141,40 @@ func (az *Cloud) GetStorageAccesskey(ctx context.Context, subsID, account, resou
...
+				// get first key
+				return v, nil
+			}
+			// get account key with latest CreationTime
+			if key == "" {
+				key = v
+				if k.CreationTime != nil {
+					creationTime = k.CreationTime.ToTime()
+				}
+				klog.V(2).Infof("got storage account key with creation time: %v", creationTime)
```
lzhecheng, line 163: nit: Do we need to print the creation time for every key? How about printing once in the end, before returning the key?

andyzhangx: there are actually only two keys: primary and secondary account keys, and the log could make sure if secondary key is rotated later, the secondary key should be retrieved, easier for troubleshooting. (by default, only one log is printed since the creation time is the same for both keys)

PR #4074, opened 2023-06-08
Fix vmssflex ensureBackendPoolDeletedFromNode

lzhecheng:
#### What type of PR is this?
/kind bug

Fix vmssflex ensureBackendPoolDeletedFromNode
* Fix loop pointer issue
* Use lock on nicUpdated var
* Fix log format

Fixes #

```release-note
Fix vmssflex ensureBackendPoolDeletedFromNode
* Fix loop pointer issue
* Use lock on nicUpdated var
* Fix log format
```

```docs

```

### DIFF
--- a/pkg/provider/azure_vmss.go
+++ b/pkg/provider/azure_vmss.go
@@ -24,6 +24,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"sync/atomic"
 
 	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2022-08-01/compute"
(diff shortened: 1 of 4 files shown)


pkg/provider/azure_vmss.go
```diff
@@ -1906,7 +1907,9 @@ func (ss *ScaleSet) ensureBackendPoolDeleted(service *v1.Service, backendPoolIDs
 				return rerr.Error()
 			}
 
+			ss.lockMap.LockEntry(updatedVMLock)
```
jwtty, line 1910: Can we use atomic.Store() instead of lock?

lzhecheng: For what purpose? atomic is never used.

PR #4393, opened 2023-08-02
InstanceExists should return true when node is unmanaged nodes

MartinForReal:
#### What type of PR is this?
/kind bug

InstanceExists should return true when node is unmanaged nodes. These nodes are created by kwok/virtual kubelet.

Fixes #

```release-note
cloudprovider will not delete virtual node.
```

```docs

```


feiskyer (commented):
thanks for the fix!

/lgtm

fuweid (approved):
LGTM

Thanks

PR #4422, opened 2023-08-07
fix: add StorageAccountCache to avoid querying storage account frequently

andyzhangx:
#### What type of PR is this?
/kind bug

fix: add StorageAccountCache to avoid querying storage account frequently

when one storage acount is exceeding quota limit(100TB), AddStorageAccountTags and RemoveStorageAccountTag are called frequently, that would lead to querying storage account frequently, thus finally lead to account operation throttling, this PR adds a timed cache to avoid this issue, and since `StorageAccountCache` is only for managing tags, it's ok that the data could be outdated due to introducing cache.

Fixes #

```release-note
fix: add StorageAccountCache to avoid querying storage account frequently
```

```docs
fix: add StorageAccountCache to avoid querying storage account frequently
```

### DIFF
--- a/pkg/provider/azure.go
+++ b/pkg/provider/azure.go
@@ -417,6 +417,8 @@ type Cloud struct {
 	pipCache azcache.Resource
 	// use LB frontEndIpConfiguration ID as the key and search for PLS attached to the frontEnd
 	plsCache azcache.Resource
+	// a timed cache storing storage account properties to avoid querying storage account frequently
+	storageAccountCache azcache.Resource
 
 	// Add service lister to always get latest service
 	serviceLister corelisters.ServiceLister
@@ -811,6 +813,9 @@ func (az *Cloud) initCaches() (err error) {
 		return err
 	}
 
+	if az.storageAccountCache, err = az.newStorageAccountCache(); err != nil {
+		return err
+	}
 	return nil
 }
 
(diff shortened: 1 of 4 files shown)


pkg/provider/azure.go
```diff
@@ -417,6 +417,8 @@ type Cloud struct {
 	pipCache azcache.Resource
 	// use LB frontEndIpConfiguration ID as the key and search for PLS attached to the frontEnd
 	plsCache azcache.Resource
+	// a timed cache storing storage account properties to avoid querying storage account frequently
+	StorageAccountCache azcache.Resource
```
lzhecheng, line 421: Is it necessary to make the variable external?

pkg/provider/azure.go
```diff
@@ -811,6 +813,10 @@ func (az *Cloud) initCaches() (err error) {
 		return err
 	}
 
+	getter := func(key string) (interface{}, error) { return nil, nil }
+	if az.StorageAccountCache, err = azcache.NewTimedCache(time.Minute, getter, false); err != nil {
```
lzhecheng, line 817: Storage account will be supported by ARG so could you pass `az.Config.DisableAPICallCache` instead of hard coding it as false?

lzhecheng: How about creating a `newStorageAccountCache()` method so it can be reused?

andyzhangx: fixed

pkg/provider/azure.go
andyzhangx: it's internal now

PR #4508, opened 2023-08-25
add node out-of-service taint when node is shutdown

### LINKED ISSUE #3269: implement Non Graceful node shutdown feature in CCM
from k8s 1.26, [Non Graceful node shutdown feature](https://kubernetes.io/docs/concepts/architecture/nodes/#non-graceful-node-shutdown) is beta, we could enable the feature gate `NodeOutOfServiceVolumeDetach`, and when a node is shutdown, give that node a taint `node.kubernetes.io/out-of-service`, and then pod with volume attachments could be successfully detached and terminated from the shutdown node, you could find all details here: https://kubernetes.io/docs/concepts/architecture/nodes/#non-graceful-node-shutdown

jwtty:
#### What type of PR is this?
/kind feature

Add `node.kubernetes.io/out-of-service` when instance is shutdown so pods can be forcefully deleted. Implement the [node non-graceful shutdown feature](https://kubernetes.io/docs/concepts/architecture/nodes/#non-graceful-node-shutdown).

Fixes #3269 

```release-note
Implement node non-graceful shut down feature. Add "node.kubernetes.io/out-of-service" taint to node when cloud provider determines the node is shutdown.
```

```docs

```


pkg/provider/azure.go
```diff
@@ -1318,6 +1332,32 @@ func (az *Cloud) updateNodeCaches(prevNode, newNode *v1.Node) {
...
+	}
+	if isNodeReady(node) {
+		if az.KubeClient != nil {
+			if err := cloudnodeutil.RemoveTaintOffNode(az.KubeClient, node.Name, node, nodeOutOfServiceTaint); err != nil {
+				klog.Errorf("failed to remove taint %s from the node %s", v1.TaintNodeOutOfService, node.Name)
+			}
+		}
+	} else {
+		// node shutdown taint is added when cloud provider determines instance is shutdown
+		if !taints.TaintExists(node.Spec.Taints, nodeOutOfServiceTaint) &&
```
andyzhangx, line 1348: when node is not ready, we cannot tell whether node is really shutdown or not

pkg/provider/azure.go
```diff
@@ -1318,6 +1332,32 @@ func (az *Cloud) updateNodeCaches(prevNode, newNode *v1.Node) {
 	}
 }
 
+// updateNodeTaint updates node out-of-service taint
+func (az *Cloud) updateNodeTaint(node *v1.Node) {
+	if node == nil {
+		return
```
andyzhangx, line 1338: provide a warning msg for `node == nil`

pkg/provider/azure.go
```diff
@@ -1318,6 +1332,32 @@ func (az *Cloud) updateNodeCaches(prevNode, newNode *v1.Node) {
...
+	}
```
andyzhangx, line 1339: let's add `if az.KubeClient != nil` in the beginning, that would make the logic more clear since `if az.KubeClient == nil`, remove taint is impossible.

```
if az.KubeClient == nil {
  print warning msg
  return
}

File: pkg/provider/azure.go [hunk 4]
```diff
@@ -1318,6 +1332,32 @@ func (az *Cloud) updateNodeCaches(prevNode, newNode *v1.Node) {
 	}
 }
 
+// updateNodeTaint updates node out-of-service taint
+func (az *Cloud) updateNodeTaint(node *v1.Node) {
+	if node == nil {
```
feiskyer, line 1337: `|| az.KubeClient == nil` could be merged together here.

pkg/provider/azure.go
jwtty: Updated, thanks

pkg/provider/azure.go
jwtty: Update, thanks

pkg/provider/azure.go
jwtty: Updated. I made it separate to add detailed warning msg. Thanks

pkg/provider/azure.go
cvvz: Should we handle this scenario? I think we should use out-of-service taint carefully since it will force delete pod and detach volume.

PR #4519, opened 2023-08-28
feat: add support for disable tcp reset

feiskyer:
#### What type of PR is this?

/kind feature

Add Service annotation `service.beta.kubernetes.io/azure-load-balancer-disable-tcp-reset` to support disable tcp reset.

Fixes #

```release-note
LoadBalancer tcp reset could be disabled with Service annotation "service.beta.kubernetes.io/azure-load-balancer-disable-tcp-reset=true"
```

```docs

```

### DIFF
--- a/pkg/consts/consts.go
+++ b/pkg/consts/consts.go
@@ -333,6 +333,9 @@ const (
 	// The list is separated by comma. It will be omitted if multi-slb is not used.
 	ServiceAnnotationLoadBalancerConfigurations = "service.beta.kubernetes.io/azure-load-balancer-configurations"
 
+	// ServiceAnnotationDisableTCPReset is the annotation used on the service to disable TCP reset on the load balancer.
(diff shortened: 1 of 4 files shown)


pkg/consts/consts.go
```diff
@@ -333,6 +333,9 @@ const (
 	// The list is separated by comma. It will be omitted if multi-slb is not used.
 	ServiceAnnotationLoadBalancerConfigurations = "service.beta.kubernetes.io/azure-load-balancer-configurations"
 
+	// ServiceAnnotationDisableTCPReset is the annotation used on the service to disable TCP reset on the load balancer.
+	ServiceAnnotationDisableTCPReset = "service.beta.kubernetes.io/azure-load-balancer-disable-tcp-reset"
```
MartinForReal, line 337: I'm wondering if we need to add annotation key for port in line 457

feiskyer: No need for specific ports. It applies to all the ports on the service.

Issue #2880, opened 2022-12-09
Nodes getting deleted after a rate-limited cache refresh


okushchenko:
#### What happened:

After adding a bunch of new nodes to our Kubernetes cluster, I observed that some of them failed to get registered. Based on the log line below, it seems like cloud-controller-manager failed to refresh the VM cache due to a rate limit, and deleted a k8s node afterwards, because it wasn't in the cache.

```
INFO  [2022-12-09T16:39:57.990980639Z] sigs.k8s.io/cloud-provider-azure/azure_vmssflex_cache.go:204: Could not find node (node-foo) in the existing cache. Forcely freshing the cache to check again... (pid: 10)
ERROR [2022-12-09T16:39:58.2093317Z] sigs.k8s.io/cloud-provider-azure/util.go:137: Send.sendRequest: error in parsing response body string "\"{\\r\\n  \\\"error\\\": {\\r\\n    \\\"details\\\": [\\r\\n      {\\r\\n        \\\"code\\\": \\\"TooManyRequests\\\",\\r\\n        \\\"message\\\": \\\"{\\\\\\\"operationGroup\\\\\\\":\\\\\\\"HighCostGet3Min\\\\\\\",\\\\\\\"startTime\\\\\\\":\\\\\\\"2022-12-09T16:39:58.2035018+00:00\\\\\\\",\\\\\\\"endTime\\\\\\\":\\\\\\\"2022-12-09T16:41:00+00:00\\\\\\\",\\\\\\\"allowedRequestCount\\\\\\\":140,\\\\\\\"measuredRequestCount\\\\\\\":866}\\\",\\r\\n        \\\"target\\\": \\\"HighCostGet3Min\\\"\\r\\n      }\\r\\n    ],\\r\\n    \\\"innererror\\\": {\\r\\n      \\\"internalErrorCode\\\": \\\"TooManyRequestsReceived\\\"\\r\\n    },\\r\\n    \\\"code\\\": \\\"OperationNotAllowed\\\",\\r\\n    \\\"message\\\": \\\"The server rejected the request because too many requests have been received for this subscription.\\\"\\r\\n  }\\r\\n}\"": json: cannot unmarshal string into Go value of type map[string]interface {}, Skip retrying regional host (pid: 10)
[...]

PR #4535, opened 2023-08-30
Improvements to reduce rate limiting with flex scalesets

gcampbell12:
#### What type of PR is this?

/kind bug

The implementation of Azure flexible scalesets in cloud controller manager is causing a high rate of API rate limiting when used at scale due to the volume of calls made to retrieve instances, this has the knock on affect of causing instances to be deleted from kubernetes because the error is passed down and the instance is handled as not found:
https://github.com/kubernetes-sigs/cloud-provider-azure/blob/371c1509418bdad8cbe23ff2d093f762e9abcf60/pkg/provider/azure_vmssflex_cache.go#L146-L157

This PR implements a new `NodeExistsByProviderID` method on the scaleset and for uniform and standard implementations this should behave the same as currently, for flex we extract the VM name from the provider ID and get the VM from the vm cache (which will call `GetVirtualMachine` if uncached) this saves us pulling the scaleset cache and listing every VM when `InstanceExists` is invoked by the node lifecycle controller. Also changes the power status and provisioning status checks to use a provider ID instead of node names for the same reasons so we can easily retrieve the desired VM just by taking the VM name from the provider ID in flex

Fixes https://github.com/kubernetes-sigs/cloud-provider-azure/issues/2880

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/cache/azure_cache.go
+++ b/pkg/cache/azure_cache.go
@@ -41,6 +41,10 @@ const (
 	// CacheReadTypeForceRefresh force refreshes the cache even if the cache entry
 	// is not expired
 	CacheReadTypeForceRefresh
+	// CacheReadTypeNoRefresh returns data from cache even if the cache entry is
+	// active/expired. If entry doesn't exist in cache, then nil is returned.
+	// Data must be nil checked before inference.
+	CacheReadTypeNoRefresh
 )
 
 // GetFunc defines a getter function for timedCache.
@@ -178,6 +182,10 @@ func (t *TimedCache) get(key string, crt AzureCacheReadType) (interface{}, error
 
 	entry.Lock.Lock()
 	defer entry.Lock.Unlock()
+
+	if crt == CacheReadTypeNoRefresh {
+		return entry.Data, nil
+	}
 
 	// entry exists and if cache is not force refreshed
 	if entry.Data != nil && crt != CacheReadTypeForceRefresh {
--- a/pkg/cache/azure_cache_test.go
+++ b/pkg/cache/azure_cache_test.go
@@ -230,6 +230,31 @@ func TestCacheAllowUnsafeRead(t *testing.T) {
 	assert.Equal(t, val, v, "cache should return expired as allow unsafe read is allowed")
 }
 
+func TestCacheAllowNoRefreshRead(t *testing.T) {
+	val := &fakeDataObj{}
+	data := map[string]*fakeDataObj{
+		testKey: val,
+	}
+	dataSource, cache := newFakeCache(t)
+	dataSource.set(data)
+
+	v, err := cache.GetWithDeepCopy(testKey, CacheReadTypeDefault)
+	assert.NoError(t, err)
+	assert.Equal(t, 1, dataSource.called)
+	assert.Equal(t, val, v, "cache should get correct data")
+
+	time.Sleep(fakeCacheTTL)
+	v, err = cache.GetWithDeepCopy(testKey, CacheReadTypeNoRefresh)
+	assert.NoError(t, err)
+	assert.Equal(t, 1, dataSource.called)
+	assert.Equal(t, val, v, "cache should return expired as allow unsafe read is allowed")
+
+	v, err = cache.GetWithDeepCopy("doesNotExist", CacheReadTypeNoRefresh)
+	assert.NoError(t, err)
+	assert.Equal(t, 1, dataSource.called)
+	assert.Equal(t, nil, v, "cache should return nil as entry does not exist")
+}
+
 func TestCacheNoConcurrentGet(t *testing.T) {
 	val := &fakeDataObj{}
 	data := map[string]*fakeDataObj{
--- a/pkg/provider/azure_controller_vmssflex.go
+++ b/pkg/provider/azure_controller_vmssflex.go
@@ -279,23 +279,24 @@ func (fs *FlexScaleSet) updateCache(nodeName string, vm *compute.VirtualMachine)
 	if vm.OsProfile == nil || vm.OsProfile.ComputerName == nil {
 		return fmt.Errorf("vm.OsProfile.ComputerName is nil")
 	}
-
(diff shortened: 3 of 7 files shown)


feiskyer:
Thanks for the contribution! But I'm a little concerned on this PR as it would impact all of the VM types. 

Could you share the logs of CCM on VMSS flex nodes, especially why the Nodes were deleted? When cache refresh failed, the controller should continue to refresh until it succeeds.

gcampbell12:
@feiskyer I'll get some logs to you later but to explain what we are seeing:

If we look at `GetNodeNameByProviderID` this calls `fs.getNodeNameByVMName`, this will retrieve a list of **every** VMSS  and begin range through them calling `ListVmssFlexVMsWithoutInstanceView` and `ListVmssFlexVMsWithOnlyInstanceView` for each scaleset, if you have say 50 scalesets and a couple of hundred nodes you soon get rate limited and azure returns 429's (I believe the number of VMs returned in these responses also contributes to rate limits being used up), this error is propagated from the vmclient back to here https://github.com/kubernetes-sigs/cloud-provider-azure/blob/d9bb39d2e81d766416a5cee932403d246927de32/pkg/provider/azure_vmssflex_cache.go#L84-L87

And here it just ends up being logged and we carry on looping over the rest of VMSS's (further using up rate limits)
https://github.com/kubernetes-sigs/cloud-provider-azure/blob/d9bb39d2e81d766416a5cee932403d246927de32/pkg/provider/azure_vmssflex_cache.go#L146-L149
Eventually the getter fails and is ran again with a forced cache refresh which also fails due to the volume of calls and we return `cloudprovider.InstanceNotFound` https://github.com/kubernetes-sigs/cloud-provider-azure/blob/d9bb39d2e81d766416a5cee932403d246927de32/pkg/provider/azure_vmssflex_cache.go#L157
Here we return false to the node lifecycle controller https://github.com/kubernetes-sigs/cloud-provider-azure/blob/d9bb39d2e81d766416a5cee932403d246927de32/pkg/provider/azure_instances.go#L205-L207 which causes the node lifecycle controller to delete the node from k8s https://github.com/kubernetes-sigs/cloud-provider-azure/blob/d9bb39d2e81d766416a5cee932403d246927de32/vendor/k8s.io/cloud-provider/controllers/nodelifecycle/node_lifecycle_controller.go#L161-L178

Since I've made some improvements to the underlying cache methods in flex I might be able to revert the change to add new  Interfaces for all the scalesets and just fix the flex ones in place, I'll look at that later.

gcampbell12:
@feiskyer It's actually pretty easy to replicate this if you client side rate limit yourself to, e.g.
```
ERROR [2023-09-07T15:46:55.990557043Z] sigs.k8s.io/cloud-provider-azure/azure_vmssflex_cache.go:148: failed to refresh vmss flex VM cache for vmssFlexID /subscriptions/[subscription-id]/resourceGroups/[resource-group-name]/providers/Microsoft.Compute/virtualMachineScaleSets/[vmss-name-1] (pid: 10)
ERROR [2023-09-07T15:46:55.990567562Z] sigs.k8s.io/cloud-provider-azure/azure_vmssflex_cache.go:85: ListVmssFlexVMsWithoutInstanceView failed: &{true 0 0001-01-01 00:00:00 +0000 UTC azure cloud provider rate limited(read) for operation "VMList"} (pid: 10)
ERROR [2023-09-07T15:46:55.990585285Z] sigs.k8s.io/cloud-provider-azure/azure_vmssflex_cache.go:148: failed to refresh vmss flex VM cache for vmssFlexID /subscriptions/[subscription-id]/resourceGroups/[resource-group-name]/providers/Microsoft.Compute/virtualMachineScaleSets/[vmss-name-2] (pid: 10)
ERROR [2023-09-07T15:46:55.990602967Z] sigs.k8s.io/cloud-provider-azure/azure_vmssflex_cache.go:85: ListVmssFlexVMsWithoutInstanceView failed: &{true 0 0001-01-01 00:00:00 +0000 UTC azure cloud provider rate limited(read) for operation "VMList"} (pid: 10)
ERROR [2023-09-07T15:46:55.990630277Z] sigs.k8s.io/cloud-provider-azure/azure_vmssflex_cache.go:148: failed to refresh vmss flex VM cache for vmssFlexID /subscriptions/[subscription-id]/resourceGroups/[resource-group-name]/providers/Microsoft.Compute/virtualMachineScaleSets/[vmss-name-3] (pid: 10)
ERROR [2023-09-07T15:46:55.990645964Z] sigs.k8s.io/cloud-provider-azure/azure_vmssflex_cache.go:85: ListVmssFlexVMsWithoutInstanceView failed: &{true 0 0001-01-01 00:00:00 +0000 UTC azure cloud provider rate limited(read) for operation "VMList"} (pid: 10)
ERROR [2023-09-07T15:46:55.990661875Z] sigs.k8s.io/cloud-provider-azure/azure_vmssflex_cache.go:148: failed to refresh vmss flex VM cache for vmssFlexID /subscriptions/[subscription-id]/resourceGroups/[resource-group-name]/providers/Microsoft.Compute/virtualMachineScaleSets/[vmss-name-4] (pid: 10)
ERROR [2023-09-07T15:46:55.990700825Z] sigs.k8s.io/cloud-provider-azure/azure_vmssflex_cache.go:85: ListVmssFlexVMsWithoutInstanceView failed: &{true 0 0001-01-01 00:00:00 +0000 UTC azure cloud provider rate limited(read) for operation "VMList"} (pid: 10)
ERROR [2023-09-07T15:46:55.990714065Z] sigs.k8s.io/cloud-provider-azure/azure_vmssflex_cache.go:148: failed to refresh vmss flex VM cache for vmssFlexID /subscriptions/[subscription-id]/resourceGroups/[resource-group-name]/providers/Microsoft.Compute/virtualMachineScaleSets/[vmss-name-5] (pid: 10)
INFO  [2023-09-07T15:46:55.990728857Z] sigs.k8s.io/cloud-provider-azure/node_lifecycle_controller.go:164: deleting node since it is no longer present in cloud provider: [k8s-node-name] (pid: 10)
```

PR #4537, opened 2023-08-31
fix: truncate pls frontendIPConfig name if it's too long

jwtty:
#### What type of PR is this?
/kind bug

In case cx's subnet name or pls name is too long, pls creation would fail because frontendIPConfiguration name is too long with error `InvalidResourceName`. This PR fixes the issue by truncating the frontend name if its length exceeds Azure's limit, 80.

Fixes #

```release-note
Fix privateLinkService creation failure in case customer's subnet name or privateLinkService name is too long.
```

```docs

```


MartinForReal:
I'm wondering if it would be better to generate an event for this, but it is not important.

PR #4822, opened 2023-10-20
Revert "Use CAAPH for credential-provider-capz template"

lzhecheng:
Reverts kubernetes-sigs/cloud-provider-azure#4812


lzhecheng:
Reverting this change before CAPZ root cause the issue with their new feature CAAPH.

Skipping tests because the change is not covered.

PR #5125, opened 2023-12-06
Allow parsing of certificate chains and password-less certificates

handsomejack-42:
#### What type of PR is this?

/kind bug

**Certificate parsing function isn't capable of parsing certificate chains**

The `decodePksc12` function fails to parse my PKCS12 certificate chain with the message

```
decoding the client certificate: decoding the PKCS#12 client certificate: pkcs12: expected exactly two safe bags in the PFX PDU
```

because, internally, it uses the [pkcs12.Decode](https://github.com/SSLMate/go-pkcs12/blob/a23dd40d71e2f5498281f7f86bec59c39447b1bc/pkcs12.go#L398) function, which expects only the key and a main cert, not a whole chain of certs. It is mentioned in the comment that

> Since PKCS#12 files often contain more than one certificate, you probably want to use [DecodeChain] instead.

The option is either switching to that `.DecodeChain` function, or, there is a [cert parser implemented in azidentity](https://github.com/Azure/azure-sdk-for-go/blob/f8be4a3c0bf9cc64feff4d5502b03e0e6aeba6d2/sdk/azidentity/client_certificate_credential.go#L91), that provides even more (PEM) parsing capabilities.

I propose using the parsing function from `azidentity` instead, as that is the 'native' identity provider for azure anyway.

**Password-less certificates**

When parsing the config into credential providers in `NewAuthProvider`, the only way to trigger the certificate credential path is to have password protected certificate. Certificates can exist without password protection.

Remove this condition to allow even password-less certificates.

x

x

```release-note
Added certificate parsing capability for password-less certificates and certificate chains
```

```docs

```


MartinForReal:
Thanks for the fix!
/lgtm
/approve

PR #5126, opened 2023-12-06
chores: Extract GetStorageAccesskey function from provider package

MartinForReal:
#### What type of PR is this?
/kind cleanup

chores: Extract GetStorageAccesskey function from provider package

Fixes #

```release-note
NONE
```

```docs

```


pkg/provider/storage/storage_account.go
```diff
@@ -0,0 +1,79 @@
...
+	"time"
+
+	"k8s.io/klog/v2"
+
+	"sigs.k8s.io/cloud-provider-azure/pkg/azclient/accountclient"
+)
+
+// GetStorageAccesskey gets the storage account access key
+// getLatestAccountKey: get the latest account key per CreationTime if true, otherwise get the first account key
+func GetStorageAccesskey(ctx context.Context, saClient accountclient.Interface, account, resourceGroup string, getLatestAccountKey bool) (string, error) {
```
andyzhangx, line 32: are you missing `subsID` parameter?
https://github.com/kubernetes-sigs/cloud-provider-azure/blob/cc2003e00479e8186b9edd9f2b91cf48dc577711/pkg/provider/azure_storageaccount.go#L134

andyzhangx: it's better keeping the same interface, here is the caller in azure file csi driver: https://github.com/kubernetes-sigs/azurefile-csi-driver/blob/f35d4633debddcc0f470ff4853e6402ebc9d0fe0/pkg/azurefile/azurefile.go#L787

MartinForReal: sub is part of track2 client https://github.com/Azure/azure-sdk-for-go/blob/e4a2a1fed7d240c87aafd6f5f7c1d82f23db50c6/sdk/resourcemanager/storage/armstorage/accounts_client.go#L27

MartinForReal: I'm wondering if we need to create client object from scratch.

MartinForReal: How about tenantid? Does csi driver support cross tenant scenario?

andyzhangx: No, cross tenant is not supported yet, let's focus on support existing cross-subs first.

MartinForReal: then the csi driver should create client before invoke GetStorageAccesskey.

MartinForReal: the client is reused if it is created by ClientFactory

andyzhangx: why not pass subs in the `GetStorageAccesskey` func? otherwise the csi driver needs to create client in every GetStorageAccesskey call

MartinForReal: we need to create a client for each request. And most of the clients can be shared

cvvz: Looks good to me, we not only need to generate a new client when cross sub but also when we need to use a new client with new credentials: https://github.com/kubernetes-sigs/cloud-provider-azure/pull/4996/files#diff-43399bb105b7d123e4c10ca28097886143c41c198bdd8189be09eba0a5d70fdaR160-R170

MartinForReal: https://github.com/kubernetes-sigs/cloud-provider-azure/pull/5138 reuse existing client config and create new client

PR #5219, opened 2024-01-05
Truncate lengthy PIP name

### LINKED ISSUE #5194: Lengthy PIP Prefix Name not supported - causes invalid PIP address name
#### What happened:

While using the `azure-pip-prefix-id`/`azure-pip-prefix-id-ipv6` annotations to provide a PIP Prefix, if the PIP Prefix name is too long (>31 chars), the auto-generated name for the PIP Address becomes unsupported by Azure, causing the LB not being synced correctly.

```
  Warning  SyncLoadBalancerFailed  46s (x2 over 47s)  service-controller  Error syncing load balancer: failed to ensure load balancer: Retriable: false, RetryAfter: 0s, HTTPStatusCode: 400, RawError: {
  "error": {
    "code": "InvalidResourceName",
    "message": "Resource name kubernetes-aec772416db7f4e228004265424d3059-ls-pme-tst-us-il001a-eus-ipv6-in-0-IPv6 is invalid. The name can be up to 80 characters long. It must begin with a word character, and it must end with a word character or with '_'. The name may contain word characters or '.', '-', '_'.",
    "details": []
  }
}
```

```
service.beta.kubernetes.io/azure-pip-prefix-id-ipv6: /subscriptions/xxxxxxxxxxxx/resourceGroups/xxxxxxxxxxxx/providers/Microsoft.Network/publicIPPrefixes/ls-pme-tst-us-il001a-eus-ipv6-in-0
```

The auto-generated name should always be 80 chars max, and the PIP Prefix name could either be hashed, or truncated (e.g. `kubernetes-aec772416db7f4e228004265424d3059-<hash>-IPv6`)
This would ensure maximum compatibility, although compromising on the readability of the PIP name.

Alternatively, an explicit validation of the input PIP Prefix should be enforced, and the max PIP Prefix name length advertised in the doc.

https://cloud-provider-azure.sigs.k8s.io/topics/loadbalancer/
image

Create an external LB service and provide a `service.beta.kubernetes.io/azure-pip-prefix-id` annotation with a PIP Prefix that has a lengthy name (the exact size might depend on the the usage of `azure-pip-prefix-id` vs `azure-pip-prefix-id-ipv6`, but 40 chars or more will cause this to happen).

Might be related to https://github.com/kubernetes-sigs/cloud-provider-azure/issues/1443 in terms of possible implementation.
Possibly this function needs some change https://github.com/kubernetes-sigs/cloud-provider-azure/blob/7c0a8afb84df58905abe03583929f0f70ad48236/pkg/provider/azure_standard.go#L336

#### Environment:

- Kubernetes version (use `kubectl version`): v1.27.3
- Cloud provider or hardware configuration: Azure, AKS
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

lzhecheng:
#### What type of PR is this?
/kind bug

Truncate lengthy PIP name. If PIP prefix is lengthy, the PIP name may be longer than 80. If so, the PIP name needs truncation.

Fixes #5194

```release-note
Truncate lengthy PIP name. If PIP prefix is lengthy, the PIP name may be longer than 80. If so, the PIP name needs truncation.
```

```docs

```

### DIFF
--- a/pkg/consts/consts.go
+++ b/pkg/consts/consts.go
@@ -377,6 +377,8 @@ const (
 	FrontendIPConfigNameMaxLength = 80
 	// LoadBalancerRuleNameMaxLength is the max length of the load balancing rule
 	LoadBalancerRuleNameMaxLength = 80
+	// PIPPrefixNameMaxLength is the max length of the PIP prefix name
+	PIPPrefixNameMaxLength = 80
 	// IPFamilySuffixLength is the length of suffix length of IP family ("-IPv4", "-IPv6")
 	IPFamilySuffixLength = 5
 
--- a/pkg/provider/azure_standard.go
+++ b/pkg/provider/azure_standard.go
@@ -342,7 +342,14 @@ func (az *Cloud) getPublicIPName(clusterName string, service *v1.Service, isIPv6
 			pipName = fmt.Sprintf("%s-%s", pipName, id)
 		}
 	}
-	return getResourceByIPFamily(pipName, isDualStack, isIPv6), nil
+
+	pipNameSegment := pipName
+	maxLength := consts.PIPPrefixNameMaxLength - consts.IPFamilySuffixLength
+	if len(pipName) > maxLength {
+		pipNameSegment = pipNameSegment[:maxLength]
+		klog.V(6).Infof("original PIP name is lengthy %q, truncate it to %q", pipName, pipNameSegment)
+	}
+	return getResourceByIPFamily(pipNameSegment, isDualStack, isIPv6), nil
 }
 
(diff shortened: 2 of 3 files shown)


lodrem:
Should it emit a warning event to inform the user?

lzhecheng:
> Should it emit a warning event to inform the user?

I think no warning is fine because users don't have to pay attention to the name?

lodrem:
Just curious, why not using hash or returning error if the value gets too long?
And what happens when two names clash, like `aaaa..aaaa` and `aaaa..aaaabbbb` (more than 80), and they both generate the same PIP name `aaaa..aaaa` (==80)?

feiskyer (commented):
/lgtm

feiskyer:
Please cherry-pick this to release branches once merged.

PR #5325, opened 2024-01-22
fix: Minor bugs in deploy-cluster-capz.sh

ab63:
#### What type of PR is this?

/kind bug

"kind-" is repeated in the cluster name as MGMT_CLUSTER_CONTEXT already includes the prefix. Namespace name is missing while getting secret

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/hack/deploy-cluster-capz.sh
+++ b/hack/deploy-cluster-capz.sh
@@ -129,7 +129,7 @@ function create_workload_cluster() {
   fi
 
   echo "Waiting for the kubeconfig to become available"
-  timeout --foreground 1000 bash -c "while ! kubectl get secrets | grep ${CLUSTER_NAME}-kubeconfig; do sleep 1; done"
+  timeout --foreground 1000 bash -c "while ! kubectl get secrets -n "${CLUSTER_NAME}" | grep ${CLUSTER_NAME}-kubeconfig; do sleep 1; done"
   if [ "$?" == 124 ]; then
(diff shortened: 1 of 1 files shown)


hack/deploy-cluster-capz.sh
```diff
@@ -129,7 +129,7 @@ function create_workload_cluster() {
   fi
 
   echo "Waiting for the kubeconfig to become available"
-  timeout --foreground 1000 bash -c "while ! kubectl get secrets | grep ${CLUSTER_NAME}-kubeconfig; do sleep 1; done"
+  timeout --foreground 1000 bash -c "while ! kubectl get secrets -n "${CLUSTER_NAME}" | grep ${CLUSTER_NAME}-kubeconfig; do sleep 1; done"
```
lzhecheng, line 132: IRRC, the secret is in default namespace. Have you check it?

ab63: This was the case earlier but it was updated with this change https://github.com/kubernetes-sigs/cloud-provider-azure/commit/cd295b03129579e5834a73cb41aa78a8d1aa1898

lzhecheng: Got it. I was not aware of the change.

ab63:
The failure should be fixed with https://github.com/kubernetes-sigs/cloud-provider-azure/commit/79d3cc48c13b4fdb21b0f9ce76b3460ae8192cba

lzhecheng:
The failure is not related to the change. Merging this PR.

PR #5347, opened 2024-01-31
Updates Makefile to allow override of $(PKG_CONFIG)

theobarberbany:
This change updates the makefile to allow an override of the $(PKG_CONFIG) target.

This allows for overriding of the generated ldflags from `hack/pkg-config.sh`.

### What type of PR is this?
/kind feature

#### Why?

We are trying to ship the acr-credential-provider with openshift. To do this we are building an RPM. The build system expects the version(s) to be settable from env vars so they are consistent across openshift (e.g 4.15). We don't want to use the git commit / version, or just pin to 'latest'. 

#### What?

This allows setting ldflags without using the `hack/pkg-config.sh` script.

Currently, the options for setting version `ldflags` are:

-  Set everything to `latest` 
-  Rely on `git` being present in the build environment, and use the pkg-config script.

This change allows you to override the invocation of the pkg-config script, and pass your own flags in the `.pkg_config` file. 

NONE

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -372,7 +372,11 @@ clean: ## Cleanup local builds.
 	rm -rf $(BIN_DIR) $(PKG_CONFIG) $(TEST_RESULTS_DIR)
 
 $(PKG_CONFIG):
+ifeq ($(OVERRIDE_PKG_CONFIG), true)
+	@echo "Overriding pkg-config script. Use an alternate method to generate .pkg_config"
+else
 	ENABLE_GIT_COMMAND=$(ENABLE_GIT_COMMAND) hack/pkg-config.sh > $@
+endif
 
 ## --------------------------------------
 ##@ Release


Makefile
```diff
@@ -372,7 +372,11 @@ clean: ## Cleanup local builds.
 	rm -rf $(BIN_DIR) $(PKG_CONFIG) $(TEST_RESULTS_DIR)
 
 $(PKG_CONFIG):
+ifeq ($(OVERRIDE_PKG_CONFIG), true)
+	@echo "Overriding pkg-config script. Use an alternate method to generate .pkg_config"
+else
```
elmiko, line 377: do we need to set the `ENABLED_GIT_COMMAND` for other targets that might come after this one?

theobarberbany: I'm not 100% following: are you suggesting we want to add

```bash
ENABLE_GIT_COMMAND=$(ENABLE_GIT_COMMAND)
```
to the override part of the target?

lzhecheng:
This change LGTM. Do you have any concern? @feiskyer

PR #5360, opened 2024-02-02
Refactor: remove async ops in vmset

MartinForReal:
#### What type of PR is this?
/kind cleanup

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/provider/azure_controller_standard.go
+++ b/pkg/provider/azure_controller_standard.go
@@ -36,16 +36,16 @@ import (
 )
 
 // AttachDisk attaches a disk to vm
-func (as *availabilitySet) AttachDisk(ctx context.Context, nodeName types.NodeName, diskMap map[string]*AttachDiskOptions) (*azure.Future, error) {
+func (as *availabilitySet) AttachDisk(ctx context.Context, nodeName types.NodeName, diskMap map[string]*AttachDiskOptions) error {
 	vm, err := as.getVirtualMachine(nodeName, azcache.CacheReadTypeDefault)
 	if err != nil {
-		return nil, err
+		return err
 	}
 
 	vmName := mapNodeNameToVMName(nodeName)
 	nodeResourceGroup, err := as.GetNodeResourceGroup(vmName)
 	if err != nil {
-		return nil, err
+		return err
 	}
 
 	disks := make([]compute.DataDisk, len(*vm.StorageProfile.DataDisks))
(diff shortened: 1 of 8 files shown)


pkg/provider/azure_controller_standard.go
```diff
@@ -36,16 +36,16 @@ import (
 )
 
 // AttachDisk attaches a disk to vm
-func (as *availabilitySet) AttachDisk(ctx context.Context, nodeName types.NodeName, diskMap map[string]*AttachDiskOptions) (*azure.Future, error) {
```
andyzhangx, line 39: how would you handle such case if you don't expose future in the return value of this func?

https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/5b0b17a8dd10fabd552c9d509ac1f7cae3a93c55/pkg/azuredisk/azure_controller_common.go#L275-L299

MartinForReal: Please see line 120, the result has been fetched before the function returns.

andyzhangx: I mean following logic in azure_controller_common.go could not be handled if future not returned:

```
	if vmUpdateRequired(future, err) {
		if derr := kwait.ExponentialBackoffWithContext(ctx, updateVMBackoff, func(ctx context.Context) (bool, error) {
			klog.Errorf("Retry VM Update on node (%s) due to error (%v)", nodeName, err)
			future, err = vmset.UpdateVMAsync(ctx, nodeName)
			if err == nil {
				err = vmset.WaitForUpdateResult(ctx, future, nodeName, "attach_disk")
			}
			return !vmUpdateRequired(future, err), nil
		}); derr != nil {
			err = derr
			return
		}
	}

	if err != nil && configAccepted(future) {
		err = retry.NewPartialUpdateError(err.Error())
	}
```

MartinForReal: no need to invoke err = vmset.WaitForUpdateResult(ctx, future, nodeName, "attach_disk") because it has been invoked in UpdateVM

andyzhangx: but how is this logic handled?
```
if vmUpdateRequired(future, err) {
		if derr := kwait.ExponentialBackoffWithContext(ctx, updateVMBackoff, func(ctx context.Context) (bool, error) {
			klog.Errorf("Retry VM Update on node (%s) due to error (%v)", nodeName, err)
			future, err = vmset.UpdateVMAsync(ctx, nodeName)
...

```

MartinForReal: the request should be resent when status code is ok and error code is OperationPreemptedErrorCode. the client will return error when the status is not healthy. we can check the error and compare the errorcode.

andyzhangx (approved):
/lgtm
pls make sure the logic handled in csi driver would be correctly migrated using sdk track 2, thanks.

Issue #5558, opened 2024-02-28
Failure to initialize on Azure Stack for v1.29


theobarberbany:
#### What happened:
On Azure StackHub initialization of the cloud provider fails: 
```
I0126 09:40:31.744654       1 leaderelection.go:281] successfully renewed lease openshift-cloud-controller-manager/cloud-controller-manager
I0126 09:40:36.766965       1 azure.go:931] Azure cloudprovider using try backoff: retries=6, exponent=1.500000, duration=6, jitter=1.000000
I0126 09:40:36.767030       1 azure_auth.go:142] azure: using client_id+client_secret to retrieve access token
I0126 09:40:36.767115       1 azure_diskclient.go:68] Azure DisksClient using API version: 2019-03-01
I0126 09:40:36.767196       1 azure_blobclient.go:73] Azure BlobClient using API version: 2019-06-01
W0126 09:40:36.767226       1 azure_privateendpointclient.go:66] Azure Stack is not supported for Private Endpoint API
W0126 09:40:36.767239       1 azure_privatednszonegroupclient.go:66] Azure Stack is not supported for Private DNS Zone Group API
F0126 09:40:36.777509       1 controllermanager.go:314] Cloud provider azure could not be initialized: could not init cloud provider azure: provided Cloud field is missing Azure Resource Manager configuration
```

#### What you expected to happen:
The cloud-provider to succesfully initialize without an error

#### How to reproduce it (as minimally and precisely as possible):
Start a cluster on Azure Stack Hub using the 1.29 release code for the cloud provider.

cloud.conf (exposed `CLOUD_CONFIG`) : 
```json
{
	"cloud": "AzureStackCloud",
	"tenantId": "edd13307-68dc-48da-9ee2-2791182e571f",
	"aadClientId": "",
	"aadClientSecret": "",
	"aadClientCertPath": "",
	"aadClientCertPassword": "",
	"useManagedIdentityExtension": false,
	"userAssignedIdentityID": "",
	"subscriptionId": "de7e09c3-b59a-4c7d-9c77-439c11b92879",
	"resourceManagerEndpoint": "https://management.mtcazs.wwtatc.com/",
	"resourceGroup": "zhsun-ash-bv56k-rg",
	"location": "mtcazs",
	"vnetName": "zhsun-ash-bv56k-vnet",
	"vnetResourceGroup": "zhsun-ash-bv56k-rg",
[...]

PR #5643, opened 2024-03-11
Handle endpoint not being rerturned in AzureCloudConfigFromURL

theobarberbany:
#### What type of PR is this?
/kind bug

This change updates AzureCloudConfigFromURL to check that the URL (ResourceManagerEndpoint) that is used to build the config has been returned. If it has not, it sets it using the URL provided.

This avoids the case where the endpoint is used to build config, but does not return it's self - causing the created configuration to fail .

Fixes  https://github.com/kubernetes-sigs/cloud-provider-azure/issues/5558

```release-note
NONE
```

```docs
NONE
```


MartinForReal:
Thank you for the pr! nice catch! @theobarberbany

PR #5650, opened 2024-03-12
feat: skip invalid LoadBalancerSourceRanges network CIDRs when provisioning nsg rules

### LINKED ISSUE #5599: Validation doesn't catch CIDR blocks that Azure NSG won't accept
#### What happened:
An entry in `loadbalancerSourceRanges` was added that contained bits set in the host portion of the address (example: 10.0.0.1/24 instead of 10.0.0.0/24). Kubernetes accepts this because the validation on the service object is just checking for a properly formatted CIDR block. Azure NSG rules will not accept this because it's a host address rather than a network address, and this breaks the reconcile loop for the service. This prevents any further updates from occurring until the bad address entry is removed.

An example error in the event is as follows:

```
Error syncing load balancer: failed to ensure load balancer: Retriable: false, RetryAfter: 0s, HTTPStatusCode: 400, RawError: {
  "error": {
    "code": "SecurityRuleInvalidAddressPrefix",
    "message": "Security rule /subscriptions/SUBSCRIPTION/resourceGroups/RESOURCEGROUP/providers/Microsoft.Network/networkSecurityGroups/aks-agentpool-16933646-nsg/securityRules/a94f2d0ba146146b2be9e9b3d753d449-TCP-80-10.0.0.1_24 has invalid Address prefix. Value provided: 10.0.0.1/24",
    "details": []
  }
}
```

cloud-provider-azure should have validation to make sure that the host portion of the address is valid and skip that one entry from `loadBalancerSourceRanges` when creating the NSG rules (with an error event logged to Kubernetes), so that other updates are not blocked. All other updates should be applied.

EDIT: The default deny rule should still be created even if all of the entries in `loadBalancerSourceRanges` fail validation - this ensures that we fail-secure by not leaving access open the customer wanted to close off. This addresses @jwtty's [point here](https://github.com/kubernetes-sigs/cloud-provider-azure/issues/5599#issuecomment-1996587236) about making sure that we don't leave something unrestricted when it shouldn't be.

Add the following to a LoadBalancer service:
```yaml
loadBalancerSourceRanges:
- 10.0.0.1/24
```

Here is some quick go sample code to check if the CIDR block will be accepted by Azure networking:

```go
package main

import (
  "fmt"
  "net"
  "os"
  "strings"
)

func main() {
  for i := 1; i < len(os.Args); i++ {
    ipAddr, ipNet, err := net.ParseCIDR(os.Args[i])
    if err != nil {
      fmt.Printf("%s failed to parse as a valid CIDR block\n", os.Args[i])
    }

    ipNetworkAddress := net.ParseIP(strings.Split(ipNet.String(), "/")[0])
    if ipAddr.Equal(ipNetworkAddress) {
      fmt.Printf("%s is a valid network CIDR block\n", os.Args[i])
    } else {
      fmt.Printf("%s is not a valid network CIDR block\n", os.Args[i])
    }
  }
}
```

```shell
$ go build ip.go && ./ip 10.0.0.0/24 10.0.0.1/24 fe80::/64 fe80::1/64
10.0.0.0/24 is a valid network CIDR block
10.0.0.1/24 is not a valid network CIDR block
fe80::/64 is a valid network CIDR block
fe80::1/64 is not a valid network CIDR block
```

#### Environment:

- Kubernetes version (use `kubectl version`): checked in AKS 1.27 and 1.28
- Install tools: n/a
- Network plugin and version (if this is a network-related bug): n/a
- Others: n/a

jwtty:
#### What type of PR is this?
/kind feature

When user-provided `LoadBalancerSourceRanges` or `azure-allowed-ip-ranges` are not valid network prefixes, cloud-controller-manager skips these invalid ranges, emits a warning event, and adds a deny-All rule in nsg.

Fixes #5599 

```release-note
When user-provided `LoadBalancerSourceRanges` or `azure-allowed-ip-ranges` are not valid network prefixes, cloud-controller-manager skips these invalid ranges, emits a warning event, and adds a deny-All rule in nsg.
```

```docs

```

### DIFF
--- a/pkg/provider/azure_loadbalancer.go
+++ b/pkg/provider/azure_loadbalancer.go
@@ -2856,6 +2856,13 @@ func (az *Cloud) reconcileSecurityGroup(
 				consts.ServiceAnnotationAllowedIPRanges, consts.ServiceAnnotationAllowedServiceTags,
 			))
 		}
+
+		if len(accessControl.InvalidRanges) > 0 {
+			az.Event(service, v1.EventTypeWarning, "InvalidConfiguration", fmt.Sprintf(
+				"Found invalid LoadBalancerSourceRanges %v, ignoring and adding a default DenyAll rule in security group.",
+				accessControl.InvalidRanges,
+			))
+		}
 	}
 
 	var (
--- a/pkg/provider/loadbalancer/accesscontrol.go
+++ b/pkg/provider/loadbalancer/accesscontrol.go
@@ -46,6 +46,7 @@ type AccessControl struct {
 	// immutable pre-compute states.
 	SourceRanges                           []netip.Prefix
 	AllowedIPRanges                        []netip.Prefix
+	InvalidRanges                          []string
 	AllowedServiceTags                     []string
 	securityRuleDestinationPortsByProtocol map[network.SecurityRuleProtocol][]int32
 }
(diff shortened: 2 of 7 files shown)


lodrem:
LGTM

feiskyer (approved):
/lgtm

PR #5652, opened 2024-03-12
Add timeout setting for default transport

MartinForReal:
#### What type of PR is this?
/kind cleanup

Fixes #

```release-note
NONE
```

```docs

```


MartinForReal:
I think a better approach would be adding timeout in in reconcile loop or retry loop.

feiskyer:
> I think using specific timeouts throughout the codebase would be appropriate as it would allow you to asses what timeout you need for each specific use case, rather than setting a blanket rule, WDYT?

Thanks for the reviewing and yes we should add those context timeouts. Let's add those in a following PR.

/lgtm

PR #5763, opened 2024-03-21
Reorder security rules to ensure 'DenyAll' rule appears after 'Allow' rule

### LINKED ISSUE #5762: The NSG DenyAll rule appears before the Allow rule.

After updating `spec.loadBalancerSourceRanges`, the `DenyAll` rule should be repositioned after `Allow` rule.

``` bash
kubectl apply -f - <<EOF
apiVersion: v1
kind: Service
metadata:
  name: foo
  annotations:
    service.beta.kubernetes.io/azure-deny-all-except-load-balancer-source-ranges: "true"
spec:
  type: LoadBalancer
  selector:
    app: foo
  loadBalancerSourceRanges:
  - 10.10.0.0/16
  ports:
  - name: tcp
    port: 4000
    protocol: TCP
    targetPort: 4000
EOF

# Update loadBalancerSourceRanges
kubectl apply -f - <<EOF
apiVersion: v1
kind: Service
metadata:
  name: foo
  annotations:
    service.beta.kubernetes.io/azure-deny-all-except-load-balancer-source-ranges: "true"
spec:
  type: LoadBalancer
  selector:
    app: foo
  loadBalancerSourceRanges:
  - 10.10.0.0/16
  - 192.168.0.0/16 # Added
  ports:
  - name: tcp
    port: 4000
    protocol: TCP
    targetPort: 4000
EOF
```

image

The issue is only encountered with Kubernetes version 1.28 or lower.

#### Environment:

- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

lodrem:
#### What type of PR is this?

/kind bug

Fixes #5762

```release-note
Fix NSG DenyAll rule priority
```

```docs

```


pkg/provider/azure_loadbalancer.go
```diff
@@ -3069,13 +3069,108 @@ func (az *Cloud) reconcileSecurityRules(sg network.SecurityGroup,
...
 }
 
+// shouldTidySecurityRules returns true if the priorities of rules should be re-ordered.
+// The rules should be tidied if there are any allow rules after deny rules which will not work.
+func shouldTidySecurityRules(rules []network.SecurityRule) bool {
+	if len(rules) <= 1 {
+		return false
+	}
+	sort.Slice(rules, func(i, j int) bool {
+		return *rules[i].Priority < *rules[j].Priority
```
jwtty, line 3091: What if rules to be created do not have Priority yet?

lodrem: It should and will only be called after the priority is assigned to it. I assume the assignment is in https://github.com/kubernetes-sigs/cloud-provider-azure/blob/2a6e42fd6d00071509f4d506c7ba884b1b252795/pkg/provider/azure_loadbalancer.go#L3064

jwtty: oh I see. Thanks for explanation.

PR #5886, opened 2024-04-09
Fix cleaning up stale security rules

lodrem:
#### What type of PR is this?

/kind bug

Fix cleaning legacy security rules while upgrading from versions that are 1.28 or older.

Fixes #

```release-note
Fix cleaning legacy security rules while upgrading from versions that are 1.28 or older.
```

```docs

```

### DIFF
--- a/pkg/provider/loadbalancer/securitygroup/securitygroup.go
+++ b/pkg/provider/loadbalancer/securitygroup/securitygroup.go
@@ -313,10 +313,11 @@ func (helper *RuleHelper) RemoveDestinationPrefixesFromRules(prefixes []string) 
 	}
 
 	for _, rule := range helper.rules {
+		if rule.DestinationAddressPrefix != nil && index[*rule.DestinationAddressPrefix] {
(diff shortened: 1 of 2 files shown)


pkg/provider/loadbalancer/securitygroup/securitygroup.go
```diff
@@ -313,10 +313,11 @@ func (helper *RuleHelper) RemoveDestinationPrefixesFromRules(prefixes []string)
 	}
 
 	for _, rule := range helper.rules {
+		if rule.DestinationAddressPrefix != nil && index[*rule.DestinationAddressPrefix] {
+			rule.DestinationAddressPrefix = nil
+			continue
+		}
 		if rule.DestinationAddressPrefixes == nil {
```
jwtty, line 320: The original code does not work because `rule.DestinationAddressPrefixes != nil`? Is it an empty slice?

lodrem: yep

PR #5948, opened 2024-04-18
Fix getVMManagementTypeByIPConfigurationID()

lzhecheng:
#### What type of PR is this?
/kind bug

Force cache refresh for getVMManagementTypeByIPConfigurationID(). The reason is that when a new standalone VM is included in the cluster, CCM cannot tell its VM type with stale cache.

Fixes #

```release-note
Force cache refresh for getVMManagementTypeByIPConfigurationID(). The reason is that when a new standalone VM is included in the cluster, CCM cannot tell its VM type with stale cache.
```

```docs

```


feiskyer (commented):
Could you fix the unit test failures?

lzhecheng:
> Could you fix the unit test failures?

Thank you for reminding. This failure comes from the log change in this PR and it is fixed.

tallaxes (approved):
LGTM

feiskyer (commented):
/lgtm

PR #6166, opened 2024-05-08
Add ListWithInstanceView to Azure vm client

wenxuan0923:
#### What type of PR is this?
/kind feature
/kind api-change

We are currently working on the autoscaling feature for single instance pool (vms pool) - CAS will scale down vms pool with Deallocate/Delete mode based on the agentpool's setting. We will need the power state of a VM to know whether it's running or being deallocated, which is located in the instance view of VM.

CAS (Cluster-Autoscaler) relies on cloud provider to list VMs in the node resource group. This PR add a new ListWithInstanceView method to list all VMs in a resource group with expanded instance view.

Fixes #

```release-note
No breaking change on existing behavior.
```

```docs

```


feiskyer (commented):
Thanks adding the new client

/lgtm
/approve

PR #6186, opened 2024-05-13
Fix generating NSG rules while using shared BYO public IP

zarvd:
#### What type of PR is this?

/kind bug

bug: When creating LoadBalancer services with a shared BYO public IP, the NSG rules are not generated correctly.

how to reproduce:
1. create 1 BYO public IP
2. creating 2 services with `service.beta.kubernetes.io/azure-pip-name: {the PIP name}`
3. check NSG rules

This PR would list all services and then filter by IP address instead of using PIP tags when checking for services with a shared PIP.

```release-note
Fix generating NSG rules while using shared BYO public IP
```

```docs

```

### DIFF
--- a/tests/e2e/network/network_security_group.go
+++ b/tests/e2e/network/network_security_group.go
@@ -830,6 +830,153 @@ var _ = Describe("Network security group", Label(utils.TestSuiteLabelNSG), func(
 			})
 		})
 	})
+
+	When("creating 2 LoadBalancer services with shared BYO public IP", func() {
+		It("should add rules independently", func() {
+
+			const (
+				Deployment1Name = "app-01"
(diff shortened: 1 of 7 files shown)


tests/e2e/network/network_security_group.go
```diff
@@ -830,6 +830,153 @@ var _ = Describe("Network security group", Label(utils.TestSuiteLabelNSG), func(
 			})
 		})
 	})
+
+	When("creating 2 LoadBalancer services with shared BYO public IP", func() {
```
feiskyer, line 834: Thanks adding a set of new tests

/ok-to-test

feiskyer (commented):
/lgtm
/approve

PR #6601, opened 2024-07-15
chore: support PEM format cert for service principal + cert auth

bowen5:
#### What type of PR is this?

/kind feature

Currently SP+Cert auth only support PFX format. This PR is to add:
- Change to use `azidentity.ParseCertificates` instead of ``adal.DecodePfxCertificateData`, to parse both PEM and PFX format certs.
- Add logic to find expected public cert file when multi public cert bag provided, by referencing https://github.com/Azure/go-autorest/blob/main/autorest/adal/persist.go#L110.

Fixes #

```release-note
Service principal with cert auth supports PEM format cert.
```

```docs

```


feiskyer (commented):
/lgtm
/approve

PR #6846, opened 2024-08-19
[credential provider] Add a flag mirrorMapping

### LINKED ISSUE #6845: [Feat][Credential Provider] Registry mirror mapping

Need a new feature to support registry mirror mapping for credential provider. For example, the image URL is `mcr.microsoft.com/test/acronly:dev`, and credential provider fetches credential from `xxx.azurecr.io` ACR.

Steps:

1. docker push an image to an ACR `xxx.azurecr.io/test/acronly:dev` and set annonymous pull to false
2. Set up [containerd config](https://github.com/containerd/containerd/blob/main/docs/hosts.md#simple-default-host-config-for-docker) on Node 
```/etc/containerd/certs.d/mcr.microsoft.com# cat hosts.toml
server = "https://mcr.microsoft.com"

[host."https://xxx.azurecr.io"]
  capabilities = ["pull", "resolve"]
```
3. Set up `/var/lib/kubelet/credential-provider-config.yaml`
```
apiVersion: kubelet.config.k8s.io/v1
kind: CredentialProviderConfig
providers:
  - name: acr-credential-provider
    matchImages:
      - "*.azurecr.io"
      - "*.azurecr.cn"
      - "*.azurecr.de"
      - "*.azurecr.us"
      - "mcr.microsoft.com"
    defaultCacheDuration: "10m"
    apiVersion: credentialprovider.kubelet.k8s.io/v1
    args:
      - /etc/kubernetes/azure.json
      - --mirror-mapping=mcr.microsoft.com:xxx.azurecr.io
```
5. Create a Pod using image `mcr.microsoft.com/test/acronly:dev` and image pull is successful

#### Why is this needed:

Credential provider supports registry host namespace in containerd config

lzhecheng:
#### What type of PR is this?
/kind feature

[credential provider] Add a flag mirrorMapping. This flag is to mirror registry A to B when fetching credential.

Fixes #6845

```release-note
[credential provider] Add a flag mirrorMapping. This flag is to mirror registry A to B when fetching credential.
```

```docs

```

### DIFF
--- a/cmd/acr-credential-provider/main.go
+++ b/cmd/acr-credential-provider/main.go
@@ -33,6 +33,9 @@ import (
 
 func main() {
 	rand.Seed(time.Now().UnixNano())
+
+	var RegistryMirrorStr string
+
 	command := &cobra.Command{
 		Use:   "acr-credential-provider configFile",
 		Short: "Acr credential provider for Kubelet",
@@ -44,7 +47,7 @@ func main() {
 				os.Exit(1)
 			}
 
-			acrProvider, err := credentialprovider.NewAcrProviderFromConfig(args[0])
+			acrProvider, err := credentialprovider.NewAcrProviderFromConfig(args[0], RegistryMirrorStr)
 			if err != nil {
 				klog.Errorf("Failed to initialize ACR provider: %v", err)
 				os.Exit(1)
@@ -59,6 +62,10 @@ func main() {
 
 	logs.InitLogs()
 	defer logs.FlushLogs()
+
+	// Flags
+	command.Flags().StringVarP(&RegistryMirrorStr, "registry-mirror", "r", "",
+		"Mirror a source registry host to a target registry host, and image pull credential will be requested to the target registry host when the image is from source registry host")
 
 	if err := command.Execute(); err != nil {
 		os.Exit(1)
--- a/examples/out-of-tree/credential-provider-config.yaml
+++ b/examples/out-of-tree/credential-provider-config.yaml
@@ -9,5 +9,7 @@ providers:
   - "*.azurecr.cn"
   - "*.azurecr.de"
   - "*.azurecr.us"
+  - "mcr.microsoft.com"
   args:
   - /etc/kubernetes/azure.json
+  - --registry-mirror=mcr.microsoft.com:xxx.azurecr.io
--- a/pkg/credentialprovider/azure_credentials.go
+++ b/pkg/credentialprovider/azure_credentials.go
@@ -60,9 +60,10 @@ type CredentialProvider interface {
 
 // acrProvider implements the credential provider interface for Azure Container Registry.
 type acrProvider struct {
-	config      *providerconfig.AzureAuthConfig
-	environment *azclient.Environment
-	credential  azcore.TokenCredential
+	config         *providerconfig.AzureAuthConfig
+	environment    *azclient.Environment
+	credential     azcore.TokenCredential
+	registryMirror map[string]string // Registry mirror relation: source registry -> target registry
 }
 
 func NewAcrProvider(config *providerconfig.AzureAuthConfig, environment *azclient.Environment, credential azcore.TokenCredential) CredentialProvider {
@@ -74,7 +75,7 @@ func NewAcrProvider(config *providerconfig.AzureAuthConfig, environment *azclien
 }
 
 // NewAcrProvider creates a new instance of the ACR provider.
-func NewAcrProviderFromConfig(configFile string) (CredentialProvider, error) {
+func NewAcrProviderFromConfig(configFile string, registryMirrorStr string) (CredentialProvider, error) {
 	if len(configFile) == 0 {
 		return nil, errors.New("no azure credential file is provided")
 	}
@@ -120,15 +121,16 @@ func NewAcrProviderFromConfig(configFile string) (CredentialProvider, error) {
 	}
 
 	return &acrProvider{
-		config:      config,
-		credential:  managedIdentityCredential,
-		environment: &envConfig,
+		config:         config,
+		credential:     managedIdentityCredential,
+		environment:    &envConfig,
+		registryMirror: parseRegistryMirror(registryMirrorStr),
 	}, nil
 }
 
 func (a *acrProvider) GetCredentials(ctx context.Context, image string, _ []string) (*v1.CredentialProviderResponse, error) {
-	loginServer := a.parseACRLoginServerFromImage(image)
-	if loginServer == "" {
+	targetloginServer, sourceloginServer := a.parseACRLoginServerFromImage(image)
+	if targetloginServer == "" {
 		klog.V(2).Infof("image(%s) is not from ACR, return empty authentication", image)
 		return &v1.CredentialProviderResponse{
 			CacheKeyType:  v1.RegistryPluginCacheKeyType,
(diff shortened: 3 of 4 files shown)


examples/out-of-tree/credential-provider-config.yaml
```diff
@@ -9,5 +9,7 @@ providers:
   - "*.azurecr.cn"
   - "*.azurecr.de"
   - "*.azurecr.us"
+  - "mcr.microsoft.com"
```
bingosummer, line 12: is it required?

pkg/credentialprovider/azure_credentials.go
```diff
@@ -232,24 +240,55 @@ func (a *acrProvider) getFromACR(ctx context.Context, loginServer string) (strin
...
+func (a *acrProvider) processImageWithMirrorMapping(image string) (string, string) {
+	for sourceRegistry, targetRegistry := range a.mirrorMapping {
+		if strings.HasPrefix(image, sourceRegistry) {
+			return strings.Replace(image, sourceRegistry, targetRegistry, 1), sourceRegistry
+		}
+	}
+	return image, ""
+}
+
+// processMirrorMapping input format: "--mirror-mapping=aaa:bbb,ccc:ddd"
```
bingosummer, line 279: how to handle "aaa:bbb,aaa:ccc"?

pkg/credentialprovider/azure_credentials.go
```diff
@@ -232,24 +240,55 @@ func (a *acrProvider) getFromACR(ctx context.Context, loginServer string) (strin
 // parseACRLoginServerFromImage takes image as parameter and returns login server of it.
```
bingosummer, line 240: Update the comment to describe what the function returns.

pkg/credentialprovider/azure_credentials.go
```diff
@@ -232,24 +240,55 @@ func (a *acrProvider) getFromACR(ctx context.Context, loginServer string) (strin
...
 // Parameter `image` is expected in following format: foo.azurecr.io/bar/imageName:version
 // If the provided image is not an acr image, this function will return an empty string.
-func (a *acrProvider) parseACRLoginServerFromImage(image string) string {
-	match := acrRE.FindAllString(image, -1)
+func (a *acrProvider) parseACRLoginServerFromImage(image string) (string, string) {
+	processedImage, mirrorRegistry := a.processImageWithMirrorMapping(image)
```
bingosummer, line 244: the second returned value is actually the source registry or the original registry. Right? I feel that `mirrorRegistry` is a little confusing.

examples/out-of-tree/credential-provider-config.yaml
lzhecheng: It is required when testing on my local setup.

pkg/credentialprovider/azure_credentials.go
lzhecheng: Current logic takes `aaa:ccc`

pkg/credentialprovider/azure_credentials.go
lzhecheng: done

pkg/credentialprovider/azure_credentials.go
lzhecheng: Good idea. Now I use `source` and `target` for all registry/image var name.

cmd/acr-credential-provider/main.go
```diff
@@ -60,6 +63,9 @@ func main() {
 	logs.InitLogs()
 	defer logs.FlushLogs()
 
+	// Flags
+	command.Flags().StringVarP(&mirrorMappingStr, "mirror-mapping", "m", "", "mirror mapping to use")
```
andyzhangx, line 67: this flag value is defined after it's used in L50, then how could external user use defined flag to override this?

pkg/credentialprovider/azure_credentials.go
andyzhangx: so this PR only supports one mirror mapping?

andyzhangx (commented):
pls also add detailed description in the PR for `mirror-mapping=mcr.microsoft.com:xxx.azurecr.io` mapping

pkg/credentialprovider/azure_credentials.go
```diff
@@ -150,16 +152,22 @@ func (a *acrProvider) GetCredentials(ctx context.Context, image string, _ []stri
...
 			return nil, err
 		}
 
-		response.Auth[loginServer] = v1.AuthConfig{
+		response.Auth[targetloginServer] = v1.AuthConfig{
 			Username: username,
 			Password: password,
 		}
+		if sourceloginServer != "" {
+			response.Auth[sourceloginServer] = v1.AuthConfig{
```
andyzhangx, line 166: you will add auth for both `sourceloginServer` and `targetloginServer`, is that necessary?

pkg/credentialprovider/azure_credentials.go
lzhecheng: No, it supports multiple.
Bin's question is how to handle two mappings with the same source registry. The logic is that the latter one takes effect.

cmd/acr-credential-provider/main.go
lzhecheng: I think as long as it is before `command.Execute()` then it is ok.

Please check the CCM code. You can see cmd is defined first, then flags.
https://github.com/kubernetes-sigs/cloud-provider-azure/blob/9eb8173ac1c320f227b51c333e37d3b7e60c899e/cmd/cloud-controller-manager/app/controllermanager.go#L79-L159

pkg/credentialprovider/azure_credentials.go
lzhecheng: Yes. I think it does no harm, right.

andyzhangx: then you will use password pull the mcr image, not sure whether that has side effect, in this case, only `targetloginServer` needs password?

lzhecheng: When mirror mapping is used, in this case, `sourceLoginServer` is like MCR and `targetloginServer` is like ACR. So, it is the `sourceloginServer` needs password.

mainred: mcr + acr auth is a MUST, additional auth should bring no hard.
kubelet will send mcr image + acr auth to containerd to pull mcr image, and containerd will mirror the mcr image to acr through [registry host](https://github.com/containerd/containerd/blob/main/docs/hosts.md#registry-host-namespace), and finally pull mcr image from acr server.

pkg/credentialprovider/azure_credentials.go
```diff
@@ -229,27 +237,59 @@ func (a *acrProvider) getFromACR(ctx context.Context, loginServer string) (strin
...
+		if strings.HasPrefix(image, sourceRegistry) {
+			return strings.Replace(image, sourceRegistry, targetRegistry, 1), sourceRegistry
+		}
+	}
+	return image, ""
+}
+
+// processMirrorMapping input format: "--mirror-mapping=aaa:bbb,ccc:ddd"
+// output format: map[string]string{"aaa": "bbb", "ccc": "ddd"}
+func processMirrorMapping(mirrorMappingStr string) map[string]string {
```
mainred, line 282: Maybe parseMirrorMapping to be specific, and to not conflict with other processXXX func.

cmd/acr-credential-provider/main.go
mainred, line 67: Please let's elaborate the flags. I did not see the detailed description of this flag anywhere else.

```suggestion
	command.Flags().StringVarP(&mirrorMappingStr, "mirror-mapping", "m", "", "mirror mapping to mirror a source registry host to a target registry host, and image pull credentail will be requested to the target registry host when the image is from source registry host")
```

mainred: The name may be registry-mirror, since mirror is also the term used in containerd host namespace feature https://github.com/containerd/containerd/blob/main/docs/hosts.md#registry-host-namespace

pkg/credentialprovider/azure_credentials.go
```diff
@@ -229,27 +237,59 @@ func (a *acrProvider) getFromACR(ctx context.Context, loginServer string) (strin
...
+	mirrorMapping := map[string]string{}
+
+	mirrorMappingStr = strings.Trim(mirrorMappingStr, " ")
+	for _, mapping := range strings.Split(mirrorMappingStr, ",") {
+		parts := strings.Split(mapping, ":")
```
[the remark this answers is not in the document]
lzhecheng: Good catch. Updated.

pkg/credentialprovider/azure_credentials.go
lzhecheng: Done.

cmd/acr-credential-provider/main.go
lzhecheng: Done

lzhecheng:
Verified locally. SP and MSI work.

andyzhangx (approved):
/lgtm

PR #7084, opened 2024-09-18
fix cloud node manager bin path in dockerfile

mainred:
#### What type of PR is this?
/kind bug

The fix is straightfoward, and correct the binary path in the windows dockerfile entrypoint.
So far there's no case reporting this issue as the entrypoint is defined explicitly in the container manifest.

Fixes #

```release-note
NONE
```

```docs

```


andyzhangx (commented):
/ok-to-test

PR #7138, opened 2024-09-26
Service LB SKU support(reconcile path)

kartickmsft:

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/pkg/provider/azure_loadbalancer.go
+++ b/pkg/provider/azure_loadbalancer.go
@@ -41,6 +41,7 @@ import (
 	"k8s.io/klog/v2"
 	"k8s.io/utils/ptr"
 	"k8s.io/utils/strings/slices"
+	"k8s.io/apimachinery/pkg/util/intstr"
 
 	azcache "sigs.k8s.io/cloud-provider-azure/pkg/cache"
 	"sigs.k8s.io/cloud-provider-azure/pkg/consts"
@@ -2685,7 +2686,8 @@ func (az *Cloud) getExpectedLBRules(
 	// https://github.com/kubernetes/kubernetes/blob/7c013c3f64db33cf19f38bb2fc8d9182e42b0b7b/pkg/proxy/healthcheck/service_health.go#L236
 	var nodeEndpointHealthprobe *network.Probe
 	var nodeEndpointHealthprobeAdded bool
-	if servicehelpers.NeedsHealthCheck(service) && !(consts.IsPLSEnabled(service.Annotations) && consts.IsPLSProxyProtocolEnabled(service.Annotations)) {
+	if servicehelpers.NeedsHealthCheck(service) && !(consts.IsPLSEnabled(service.Annotations) && consts.IsPLSProxyProtocolEnabled(service.Annotations)) &&
+	   !az.useServiceLoadBalancer() {
 		podPresencePath, podPresencePort := servicehelpers.GetServiceHealthCheckPathPort(service)
 		lbRuleName := az.getLoadBalancerRuleName(service, v1.ProtocolTCP, podPresencePort, isIPv6)
 		probeInterval, numberOfProbes, err := az.getHealthProbeConfigProbeIntervalAndNumOfProbe(service, podPresencePort)
@@ -2706,7 +2708,8 @@ func (az *Cloud) getExpectedLBRules(
 
 	var useSharedProbe bool
 	if az.useSharedLoadBalancerHealthProbeMode() &&
-		!strings.EqualFold(string(service.Spec.ExternalTrafficPolicy), string(v1.ServiceExternalTrafficPolicyLocal)) {
+		!strings.EqualFold(string(service.Spec.ExternalTrafficPolicy), string(v1.ServiceExternalTrafficPolicyLocal)) &&
+		!az.useServiceLoadBalancer() {
 		nodeEndpointHealthprobe = az.buildClusterServiceSharedProbe()
 		useSharedProbe = true
 	}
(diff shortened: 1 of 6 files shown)


pkg/provider/azure_loadbalancer_backendpool.go
```diff
@@ -911,3 +912,203 @@ func removeNodeIPAddressesFromBackendPool(
...
+		klog.Errorf("bpi.EnsureHostsInPool: failed to get endpoint slice list for service %q, error: %s", service.Name, err.Error())
+		return err
+	}
+
+	lbBackendPoolName := bpi.getBackendPoolNameForService(service, clusterName, isIPv6)
+
+	/* Remove all addresses from the backend pool and add the addresses from all the 
+	   endpoint-slices pertaining to a service.*/
+	if strings.EqualFold(ptr.Deref(backendPool.Name, ""), lbBackendPoolName) &&
+		backendPool.BackendAddressPoolPropertiesFormat != nil {
```
david-kow, line 952: Can we escape early to avoid nesting? Also, should we handle the error case that "if" checks here?

pkg/provider/azure_loadbalancer_backendpool.go
```diff
@@ -911,3 +912,203 @@ func removeNodeIPAddressesFromBackendPool(
...
+}
+
+func (bpi *backendPoolTypePodIP) EnsureHostsInPool(service *v1.Service, _ []*v1.Node, _, _, clusterName, lbName string, backendPool network.BackendAddressPool) error {
+	isIPv6 := isBackendPoolIPv6(ptr.Deref(backendPool.Name, ""))
+
+	var (
+		changed               bool
+		err                   error
+		podIPsToBeAdded []string
+		endpointsliceName string
```
david-kow, line 935: nit:

```suggestion
		endpointSliceName string
```

pkg/provider/azure_loadbalancer_backendpool.go
```diff
@@ -911,3 +912,203 @@ func removeNodeIPAddressesFromBackendPool(
...
+				if endpoint.Conditions.Ready != nil && !*endpoint.Conditions.Ready {
+					// Skip endpoints that are not ready
+					continue
+				}
+
+				for _, address := range endpoint.Addresses {
+					klog.V(6).Infof("bpi.EnsureHostsInPool: adding ip address %s", address)
+					podIPsToBeAdded = append(podIPsToBeAdded, address)
+					endpointsliceName = strings.ToLower(fmt.Sprintf("%s/%s", ES.Namespace, ES.Name))
+					endpointSliceNames = append(endpointSliceNames,endpointsliceName)
```
david-kow, line 981: Should this be added for every address?

pkg/provider/azure_loadbalancer_backendpool.go
```diff
@@ -911,3 +912,203 @@ func removeNodeIPAddressesFromBackendPool(
...
+
+	if backendPool.LoadBalancerBackendAddresses == nil {
+		lbBackendPoolAddresses := make([]network.LoadBalancerBackendAddress, 0)
+		backendPool.LoadBalancerBackendAddresses = &lbBackendPoolAddresses
+	}
+
+	var changed bool
+	addresses := *backendPool.LoadBalancerBackendAddresses
+	for _, ipAddress := range podIPAddresses {
+		if !hasIPAddressInBackendPool(backendPool, ipAddress) {
```
david-kow, line 1100: Can we do a duplicate check earlier instead of scanning the pool with each add?

pkg/provider/azure_loadbalancer_backendpool.go
```diff
@@ -911,3 +912,203 @@ func removeNodeIPAddressesFromBackendPool(
...
+	serviceName := getServiceName(service)
+
+	lbBackendPoolNames := bpi.getBackendPoolNamesForService(service, clusterName)
+
+	if lb.LoadBalancerPropertiesFormat == nil || lb.LoadBalancerPropertiesFormat.BackendAddressPools == nil {
+		return nil, nil
+	}
+
+	backendPrivateIPv4s, backendPrivateIPv6s := utilsets.NewString(), utilsets.NewString()
+	for _, bp := range *lb.BackendAddressPools {
```
david-kow, line 1010: Can we reduce nesting a little in this func?

pkg/provider/azure_loadbalancer_backendpool.go
```diff
@@ -911,3 +912,203 @@ func removeNodeIPAddressesFromBackendPool(
...
+		return err
+	}
+
+	lbBackendPoolName := bpi.getBackendPoolNameForService(service, clusterName, isIPv6)
+
+	/* Remove all addresses from the backend pool and add the addresses from all the 
+	   endpoint-slices pertaining to a service.*/
+	if strings.EqualFold(ptr.Deref(backendPool.Name, ""), lbBackendPoolName) &&
+		backendPool.BackendAddressPoolPropertiesFormat != nil {
+		if backendPool.LoadBalancerBackendAddresses == nil {
```
david-kow, line 953: Do we need these checks? `removeNodeIPAddressesFromBackendPool` checks already and `addPodIPAddressesToBackendPool` inits it as well.

david-kow (commented):
I'd opt for more return statements and less nesting, but LGTM overall

pkg/provider/azure_loadbalancer.go
```diff
@@ -2815,6 +2823,18 @@ func (az *Cloud) getExpectedLBRules(
 				props.BackendPort = ptr.To(port.NodePort)
 				props.EnableFloatingIP = ptr.To(false)
 			}
+
+			if az.useServiceLoadBalancer() {
+				//If Interger value of TargetPort is present in the service, use it as the backend port.
+				//We current don't support string(named Port).
```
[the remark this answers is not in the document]
kartickmsft: I think better to error out as we are overridding the customer intent.

pkg/provider/azure_loadbalancer_backendpool.go
kartickmsft: A new property is used in the Address for the slice/ipgroup name. I think we don't need to populate this field if we the updateBackendIpGroup API.

pkg/provider/azure_loadbalancer_backendpool.go
kartickmsft: sure

PR #7194, opened 2024-10-03
fix: k8s 1.31 condition

mweibel:
#### What type of PR is this?
/kind bug

cloud-provider-azure helm chart doesn't work for 1.31 based clusters because of a wrong condition.

```release-note
NONE
```


mweibel:
FYI @mboersma - original change was in #6938

mboersma (approved):
/lgtm

🤦🏻

mweibel:
> /lgtm
> 
> 🤦🏻

:smile: happens :)

mboersma:
Actually I spoke too soon. Any change to the Helm chart requires a new chart release.

So we need to bump `version` to `v1.31.1` in Chart.yaml, update `cloudProviderAzureVersion`s in cloud-provider-azure.yaml to the current patches, and run `make update-helm`.

Let me know if I can help with any of that. Thanks for noticing this! Sorry for the braindead copy-paste. 😄

mweibel:
thanks for the pointers - just updated! Let me know if there's anything else I need to check.

mweibel:
found a few more version updates - hope that's all now.

mweibel:
I think the /ok-to-test is already applied but would need an /approve now if all is fine.

PR #7243, opened 2024-10-09
Ensure context is passed through in cloud provider functions

MartinForReal:
#### What type of PR is this?
/kind cleanup

Ensure context is passed through in cloud provider functions

Fixes #

```release-note
NONE
```

```docs

```


pkg/cache/azure_cache.go
```diff
@@ -64,8 +65,8 @@ func cacheKeyFunc(obj interface{}) (string, error) {
 
 // Resource operations
 type Resource interface {
-	Get(key string, crt AzureCacheReadType) (interface{}, error)
-	GetWithDeepCopy(key string, crt AzureCacheReadType) (interface{}, error)
+	Get(ctx context.Context, key string, crt AzureCacheReadType) (interface{}, error)
+	GetWithDeepCopy(ctx context.Context, key string, crt AzureCacheReadType) (interface{}, error)
 	Delete(key string) error
 	Set(key string, data interface{})
 	Update(key string, data interface{})
```
zarvd, line 72: Will you add ctx for these 3 operations?

zarvd (approved):
LGTM

pkg/cache/azure_cache.go
MartinForReal: no. get function will send request to arm api. However, these three will only update local cache store. 
It will be easier if ctx is passed through when adding tracing support in cloud provider code base.

feiskyer (commented):
/lgtm

pkg/cache/azure_cache.go
zarvd: I am thinking "program to interfaces, not implementations" though 😂

MartinForReal: I agree. Lets adopt context package step by step.

PR #7276, opened 2024-10-11
Feature: Add support to respect ARM throttling and avoid sending requests until the resource is throttled.

vpatelsj:
#### What type of PR is this?
/kind feature

This PR adds capability to throttle ARM requests originating from cloud provider when arm services are responding with 429s. This capability helps in stopping requests temporarily so that the cloud provider can escape the throttling penalties imposed by each service.

```release-note
A new bool configuration parameter CloudProviderThrottling is added to set cloud provider throttling. By default this parameter is set to false to maintain existing behavior.
```


MartinForReal:
@vpatelsj Thanks for your pr! As disused offline, this pr is not needed.

PR #8054, opened 2025-01-17
Container Load Balancer - Network Security Group (NSG)

georgeedward2000:
#### What type of PR is this?
/kind feature

For a container load balancer (CLB) with floating IP disabled, the Network Security Group (NSG) must allow traffic to dynamic pod IPs. To manage this:

- **Allow Rule for Pod Subnet**: Configure the NSG to allow traffic from the entire pod subnet of the managed cluster, accommodating dynamic pod IPs.
- **Backend Pool Configuration**: Ensure the NSG rules permit traffic to the pod IPs in the backend pool.
This setup ensures secure and efficient communication with the pod backend pool.
- **Unit Tests**

```release-note
Added the ability to set the type of backend pool to PodIP. 
This action allows the Load balancer attached to the service to route traffic directly to the containers, 
instead of routing to AKS nodes and then AKS node routing it to the PODs hosting the service.
```

```docs

```


pkg/provider/config/azure.go
```diff
@@ -168,6 +168,12 @@ type Config struct {
 	ClusterServiceSharedLoadBalancerHealthProbePort int32 `json:"clusterServiceSharedLoadBalancerHealthProbePort,omitempty" yaml:"clusterServiceSharedLoadBalancerHealthProbePort,omitempty"`
 	// ClusterServiceSharedLoadBalancerHealthProbePath defines the target path of the shared health probe. Default to `/healthz`.
 	ClusterServiceSharedLoadBalancerHealthProbePath string `json:"clusterServiceSharedLoadBalancerHealthProbePath,omitempty" yaml:"clusterServiceSharedLoadBalancerHealthProbePath,omitempty"`
+
```
kartickmsft, line 171: I think comment can be reworded to mention that bool tracks if pod subnet for the cluster has been retrieved. We could have multiple rules per subnet with different protocol,destination port.

pkg/provider/config/azure.go
```diff
@@ -168,6 +168,12 @@ type Config struct {
...
+	// RetrievedClusterPodCidr indicates whether the cluster has set the pod subnet prefix within Network Security Group (NSG).
```
kartickmsft, line 172: I think comment can be reworded to mention that it tracks whether pod subnet for the cluster has been retrieved. We could have multiple rules per subnet with different protocol,destination port.

pkg/provider/azure_loadbalancer.go
```diff
@@ -3062,49 +3062,68 @@ func (az *Cloud) reconcileSecurityGroup(
...
+			return nil, fmt.Errorf("failed to get managed cluster: %w", err)
+		}
+
+		if managedCluster.Properties == nil || managedCluster.Properties.NetworkProfile == nil ||
+			managedCluster.Properties.NetworkProfile.PodCidr == nil {
+			klog.Errorf("Failed to get PodCidr for cluster %q", clusterName)
+			return nil, fmt.Errorf("failed to get PodCidr for cluster %q", clusterName)
+		}
+
+		podCidrString := ptr.Deref(managedCluster.Properties.NetworkProfile.PodCidr, "")
```
kartickmsft, line 3081: I think it's better to read from PodCidrs field to account for dual-stack.
References below.
https://learn.microsoft.com/en-us/rest/api/aks/managed-clusters/get?view=rest-aks-2024-09-01&tabs=HTTP#containerservicenetworkprofile
https://learn.microsoft.com/en-us/azure/aks/azure-cni-overlay?tabs=kubectl#deploy-a-dual-stack-aks-cluster

I think we need to store PODCidrs within az after retrieving it so that for LB service that follow the 1st service, we already have it for use. Better to store one PODCIDR for Ipv4 and another for Ipv6.

pkg/provider/azure_loadbalancer.go
```diff
@@ -3118,7 +3137,7 @@ func (az *Cloud) reconcileSecurityGroup(
 	}
 
 	if wantLb {
-		err := accessControl.PatchSecurityGroup(dstIPv4Addresses, dstIPv6Addresses)
+		err := accessControl.PatchSecurityGroup(dstIPv4Addresses, dstIPv6Addresses, dstAddressPrefix)
```
kartickmsft, line 3140: Based on earlier comment, I think we need to pass dstIpv4AddressPrefix and dstIpv6AddressPrefix and use it in corresponding v4 and v6 rules respectively.

pkg/provider/azure_loadbalancer.go
kartickmsft: NetworkProfile json looks as shown below from an existing cluster with CNI type as Overlay.

            "podCidr": "192.168.0.0/16",
            "serviceCidr": "10.0.0.0/16",
            "dnsServiceIP": "10.0.0.10",
            "outboundType": "loadBalancer",
            "podCidrs": [
                "192.168.0.0/16"
            ],
            "serviceCidrs": [
                "10.0.0.0/16"
            ],
            "ipFamilies": [
                "IPv4"
            ]

pkg/provider/azure_loadbalancer.go
```diff
@@ -3062,49 +3062,68 @@ func (az *Cloud) reconcileSecurityGroup(
...
-			}
+		if err != nil {
+			return nil, err
+		}
+		if wantLb && !lbFound {
+			logger.Error(err, "Failed to get load balancer")
+			return nil, fmt.Errorf("unable to get lb %s", lbName)
 		}
 		var backendIPv4List, backendIPv6List []string
 		if lbFound {
```
kartickmsft, line 3108: For POD IP based backendpool, we are not populating the POD IPs in the NSG. Rather, the POD subnet. So, we could skip GetBackendPrivateIPs for POD IP based Backend pool.

pkg/provider/azure_loadbalancer.go
```diff
@@ -3062,49 +3062,68 @@ func (az *Cloud) reconcileSecurityGroup(
...
-		lbIPAddresses, _                                 = iputil.ParseAddresses(lbIPs)
-		lbIPv4Addresses, lbIPv6Addresses                 = iputil.GroupAddressesByFamily(lbIPAddresses)
-		additionalIPv4Addresses, additionalIPv6Addresses = iputil.GroupAddressesByFamily(additionalIPs)
-		backendIPv4Addresses, backendIPv6Addresses       []netip.Addr
-	)
-	{
+	var dstIPv4Addresses, dstIPv6Addresses []netip.Addr
+	var dstAddressPrefix []netip.Prefix
+
+	if az.IsLBBackendPoolTypePodIP() && !az.RetrievedClusterPodCidr {
```
kartickmsft, line 3068: I think we could modify the logic as below.

if az.IsLBBackendPoolTypePodIP() {
   if !az.RetrievedClusterPodCidr {
        //API to retrieve Pod Cidr as present in the PR.
   }
   else {
       //Fill in dstIpv4AddressPrefix and dstIpv6AddressPrefix from az
   }        
}

//Existing code with PatchSecurityGroup taking the ipv4 and ipv6 Prefix arguments.

pkg/provider/loadbalancer/accesscontrol.go
```diff
@@ -195,7 +195,7 @@ func (ac *AccessControl) AllowedIPv6Ranges() []netip.Prefix {
 }
 
 // PatchSecurityGroup checks and adds rules for the given destination IP addresses.
-func (ac *AccessControl) PatchSecurityGroup(dstIPv4Addresses, dstIPv6Addresses []netip.Addr) error {
+func (ac *AccessControl) PatchSecurityGroup(dstIPv4Addresses, dstIPv6Addresses []netip.Addr, dstAddressPrefix []netip.Prefix) error {
```
kartickmsft, line 198: Please modify this function to take dstIpv4AddressPrefix []netip.Prefix and dstIpv6AddressPrefix []netip.Prefix to account for dual-stack AKS cluster. This need not be an array but simplifies the logic to put prefix in the rule.
Then the logic will be similar to what exists for dstIpV4Addresses and dstIpV6Addresses.

pkg/provider/azure_loadbalancer.go
georgeedward2000: This is part of the initial flow - node ip based backendpool

internal/testutil/fixture/azure_loadbalancer.go
```diff
@@ -47,6 +47,18 @@ func (f *AzureLoadBalancerFixture) IPv4Addresses() []string {
 	}
```
david-kow, line 47: Can you add some details about the change in the PR description?

pkg/provider/azure_loadbalancer.go
```diff
@@ -3063,48 +3063,84 @@ func (az *Cloud) reconcileSecurityGroup(
...
+				}
+				if prefix.Addr().Is4() {
+					dstIpv4AddressPrefix = append(dstIpv4AddressPrefix, prefix)
+					az.PodCidrIPv4 = prefix
+				} else {
+					dstIpv6AddressPrefix = append(dstIpv6AddressPrefix, prefix)
+					az.PodCidrIPv6 = prefix
+				}
+			}
+			az.RetrievedClusterPodCidr = true
```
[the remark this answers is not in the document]
georgeedward2000: From what I understand, the pod subnet (hence the cidr) is set during cluster's deployment and cannot be updated during its lifetime. Please confirm @kartickmsft

internal/testutil/fixture/azure_loadbalancer.go
georgeedward2000: updated the PR description

pkg/provider/azure_loadbalancer.go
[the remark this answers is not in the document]
david-kow: I think that can happen when completely new node pool is added: https://learn.microsoft.com/en-us/azure/aks/create-node-pools#add-a-node-pool-with-a-unique-subnet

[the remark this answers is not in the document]
kartickmsft: POD subnet can be newly configured for a new nodepool for Azure CNI Dynamic IP allocation and Enhanced subnet option (https://learn.microsoft.com/en-us/azure/aks/configure-azure-cni-dynamic-ip-allocation#adding-node-pool).
So, yeah, we need to handle it differently. As per my understanding, cloud-provider doesn't get any notification about nodepool addition/deletion. If this understanding is correct, maybe reading PODCIDRs each time maybe the only option.

feiskyer:
@georgeedward2000 Thanks for the contribution. Could you sign CLA?

feiskyer:
And could you use the template (the template would show when you open the PR) and fill the template contents (including release notes information together with what this PR does)?

/kind feature

pkg/provider/azure_loadbalancer.go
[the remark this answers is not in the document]
feiskyer: yes, it may change. We'd need to figure out a way to get notified on such changes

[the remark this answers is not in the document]
feiskyer: >  If this understanding is correct, maybe reading PODCIDRs each time maybe the only option.
This may be not enough as there may be down times when new IPs are picked while they are blocked by NSG.

pkg/provider/azure_loadbalancer.go
```diff
@@ -3063,48 +3063,84 @@ func (az *Cloud) reconcileSecurityGroup(
...
+		dstIPv4Addresses, dstIPv6Addresses         []netip.Addr
+		dstIpv4AddressPrefix, dstIpv6AddressPrefix []netip.Prefix
 	)
-	{
-		// Get backend node IPs
-		lb, lbFound, err := az.getAzureLoadBalancer(ctx, lbName, azcache.CacheReadTypeDefault)
-		{
+
+	if az.IsLBBackendPoolTypePodIP() {
+		if !az.RetrievedClusterPodCidr {
```
feiskyer, line 3071: Generally, this would block the security boundary of the Kubernetes cluster. The cluster identity should not have the permission to manage itself.

feiskyer: let's use new cloud config options and ask the provisioning service (e.g. AKS or capz) to setup the CIDRs on changing.

pkg/provider/config/azure.go
```diff
@@ -168,6 +169,19 @@ type Config struct {
 	ClusterServiceSharedLoadBalancerHealthProbePort int32 `json:"clusterServiceSharedLoadBalancerHealthProbePort,omitempty" yaml:"clusterServiceSharedLoadBalancerHealthProbePort,omitempty"`
 	// ClusterServiceSharedLoadBalancerHealthProbePath defines the target path of the shared health probe. Default to `/healthz`.
 	ClusterServiceSharedLoadBalancerHealthProbePath string `json:"clusterServiceSharedLoadBalancerHealthProbePath,omitempty" yaml:"clusterServiceSharedLoadBalancerHealthProbePath,omitempty"`
+
+	// RetrievedClusterPodCidr tracks whether the pod subnet prefix for the cluster has been retrieved and set within the Network Security Group (NSG).
+	// The NSG for the pod subnet is configured once for the entire cluster.
+	// If the pod subnet prefix is not set, the value is false.
+	// If the pod subnet prefix is set, the value is true.
+	// Note: Multiple rules per subnet with different protocols and destination ports can exist.
+	RetrievedClusterPodCidr bool `json:"retrievedClusterPodCidr" yaml:"retrievedClusterPodCidr"`
```
feiskyer, line 178: As commented, let's remove this option and ask provisioning service to set the PodCIDRs

pkg/provider/config/azure.go
```diff
@@ -168,6 +169,19 @@ type Config struct {
...
+
+	// PodCidrIPv4 is the IPv4 pod subnet prefix for the cluster.
+	// PodCidrIPv6 is the IPv6 pod subnet prefix for the cluster.
+	// The pod subnet prefix is used to configure the NSG for the pod subnet.
+	PodCidrIPv4 netip.Prefix `json:"podCidrIPv4" yaml:"podCidrIPv4"`
+	PodCidrIPv6 netip.Prefix `json:"podCidrIPv6" yaml:"podCidrIPv6"`
```
feiskyer, line 184: For extensibility, let's use LIST here?

pkg/provider/config/azure.go
```diff
@@ -168,6 +169,19 @@ type Config struct {
...
+	// RetrievedClusterPodCidr tracks whether the pod subnet prefix for the cluster has been retrieved and set within the Network Security Group (NSG).
+	// The NSG for the pod subnet is configured once for the entire cluster.
+	// If the pod subnet prefix is not set, the value is false.
+	// If the pod subnet prefix is set, the value is true.
+	// Note: Multiple rules per subnet with different protocols and destination ports can exist.
+	RetrievedClusterPodCidr bool `json:"retrievedClusterPodCidr" yaml:"retrievedClusterPodCidr"`
+
+	// PodCidrIPv4 is the IPv4 pod subnet prefix for the cluster.
+	// PodCidrIPv6 is the IPv6 pod subnet prefix for the cluster.
+	// The pod subnet prefix is used to configure the NSG for the pod subnet.
```
feiskyer, line 182: Could you also add a message to say all Pod CIDR would be opened to internet by default?

and add a TODO comment to improve the security later?

feiskyer (changes requested):
As discussed, there is a security limitation of current design, please also add it as comments and TODOs.

PR #8230, opened 2025-02-05
feat: Add new authentication type UserAssignedIdentityCredential

bryan-cox:
#### What type of PR is this?
/kind feature

This pull request adds the capability to authenticate with Azure using a new authentication type available only to 1st party Microsoft applications, UserAssignedIdentityCredential. More information on this new authentication type can be found here - https://github.com/Azure/msi-dataplane/blob/main/pkg/dataplane/reloadCredentials.go#L60.

Fixes #

```release-note
This pull request adds the capability to authenticate with Azure using a new authentication type available only to 1st party Microsoft applications, UserAssignedIdentityCredential
```

```docs

```


bryan-cox:
Moved PR to ready for Review. I was able to successfully test this out locally.

MartinForReal:
Thanks for the pr!
I'm wondering what is the difference between NewManagedIdentityCredential and UserAssignedIdentityCredential ? I believe the only difference is that msi dataplane sdk is used.

bryan-cox:
> Thanks for the pr! I'm wondering what is the difference between NewManagedIdentityCredential and UserAssignedIdentityCredential ? I believe the only difference is that msi dataplane sdk is used.

Hey @MartinForReal - It's kind of like Managed Identity combined with Client Certificate. With UserAssignedIdentityCredential, you use a Managed Identity, but you can use your Microsoft 1st party status to retrieve the backing certificate from the msi data plane. With UserAssignedIdentityCredential, you can authenticate with a managed identity and its backing certificate from Microsoft.

MartinForReal:
Thank you for clarifying the difference! I'm not sure where it is used. Could you please provide more information? Thanks!
cc @feiskyer

pkg/azclient/auth_conf.go
```diff
@@ -46,6 +48,13 @@ type AzureAuthConfig struct {
 	// Auxiliary token provider for accessing resources from network tenant
 	// Require MSI to be enabled and have permission to access the KeyVault
 	AuxiliaryTokenProvider *AzureAuthAuxiliaryTokenProvider `json:"auxiliaryTokenProvider,omitempty" yaml:"auxiliaryTokenProvider,omitempty"`
+	// The path where a JSON file exists containing the JSON format of a UserAssignedIdentityCredentials struct
+	// See the msi-dataplane for more details on UserAssignedIdentityCredentials - https://github.com/Azure/msi-dataplane/blob/63fb37d3a1aaac130120624674df795d2e088083/pkg/dataplane/internal/generated_client.go#L156C6-L156C37
+	AADMSIDataPlaneIdentityPath string `json:"aadMSIDataPlaneIdentityPath,omitempty" yaml:"aadMSIDataPlaneIdentityPath,omitempty"`
+	// Used with AADMSIDataPlaneIdentityPath to specify the Cloud type.
+	// Can only be one of the following values: AzurePublicCloud, AzureUSGovernmentCloud, or AzureChinaCloud
+	// If a value is not specified, defaults to AzurePublicCloud
+	MSIDataPlaneIdentityCloudType dataplane.AzureCloud `json:"msIDataPlaneIdentityCloudType,omitempty" yaml:"msIDataPlaneIdentityCloudType,omitempty"`
```
feiskyer, line 57: could we infer it from config `Cloud`? values should be same here

pkg/azclient/auth_conf.go
```diff
@@ -46,6 +48,13 @@ type AzureAuthConfig struct {
 	// Auxiliary token provider for accessing resources from network tenant
 	// Require MSI to be enabled and have permission to access the KeyVault
 	AuxiliaryTokenProvider *AzureAuthAuxiliaryTokenProvider `json:"auxiliaryTokenProvider,omitempty" yaml:"auxiliaryTokenProvider,omitempty"`
+	// The path where a JSON file exists containing the JSON format of a UserAssignedIdentityCredentials struct
+	// See the msi-dataplane for more details on UserAssignedIdentityCredentials - https://github.com/Azure/msi-dataplane/blob/63fb37d3a1aaac130120624674df795d2e088083/pkg/dataplane/internal/generated_client.go#L156C6-L156C37
+	AADMSIDataPlaneIdentityPath string `json:"aadMSIDataPlaneIdentityPath,omitempty" yaml:"aadMSIDataPlaneIdentityPath,omitempty"`
```
feiskyer, line 53: seems this is an internal data struct, are we going to consume it or also need to config the contents here?

And if need to config it, could we generate the contents based on our AzureAuthConfig and so we don't need to duplicate the same contents in two different files?

bryan-cox:
> Thank you for clarifying the difference! I'm not sure where it is used. Could you please provide more information? Thanks!
> cc @feiskyer

Hey @MartinForReal - Not sure what you are asking for here, could you clarify? We are planning to use this authentication type in an upstream project, https://github.com/openshift/hypershift, which is being used on a joint project between Microsoft and Red Hat called ARO HCP.

pkg/azclient/auth_conf.go
bryan-cox: Sure

pkg/azclient/auth_conf.go
bryan-cox: > seems this is an internal data struct, are we going to consume it or also need to config the contents here?

Yeah it is an internal data struct, which is why this authentication type is only available to 1st party Microsoft applications. Cloud Provider would just consume the filepath where the JSON file lives and `dataplane.NewUserAssignedIdentityCredential` handles everything from there.

MartinForReal:
I got concerns about cloud config and raised an issue here https://github.com/Azure/msi-dataplane/issues/40

bryan-cox:
> I got concerns about cloud config and raised an issue here [Azure/msi-dataplane#40](https://github.com/Azure/msi-dataplane/issues/40)

Hey @MartinForReal - Steve made a change so its not as strict on what cloud type should be passed now. I updated the PR with the new change.

pkg/azclient/auth.go
```diff
@@ -150,6 +154,15 @@ func NewAuthProvider(armConfig *ARMClientConfig, config *AzureAuthConfig, client
 		}
 	}
 
+	// UserAssignedIdentityCredentials authentication
+	if computeCredential == nil && len(config.AADMSIDataPlaneIdentityPath) > 0 {
+		cloudConfig := AzureCloudConfigFromName(armConfig.Cloud)
+		computeCredential, err = dataplane.NewUserAssignedIdentityCredential(context.Background(), config.AADMSIDataPlaneIdentityPath, dataplane.WithClientOpts(azcore.ClientOptions{Cloud: ptr.Deref(cloudConfig, cloud.AzurePublic)}))
```
MartinForReal, line 160: I'm wondering if clientoption in line 43 can be reused here.

bryan-cox: I'm reusing armConfig here from the previous recommendation - https://github.com/kubernetes-sigs/cloud-provider-azure/pull/8230#discussion_r1948529994. 

Which one is the right one to use?

MartinForReal: sorry just  dataplane.WithClientOpts(azcore.ClientOptions{Cloud: ptr.Deref(cloudConfig, cloud.AzurePublic)
maybe we can replace this with clientops in line 43

bryan-cox: It's been updated

PR #8255, opened 2025-02-08
fix(credential-provider): check acr pattern to avoid spoofing

mainred:
#### What type of PR is this?
/kind bug

There's a potential spoofing when the image 
"<my-test>.azurecr.io.example.com/ubuntu2:latest" is sent to the credential provider plugin, it will fetch credential from <my-test>.azurecr.io unexpectedly.
Currently it's not possible considering kubelet will not call credential provider plugin since the "<my-test>.azurecr.io.example.com/ubuntu2:latest" does not match the matchImages defined below.
Also the credential provider plugin as a binary is installed on k8s node, is currently only called by kubelet if there's no manual operation.

```
apiVersion: kubelet.config.k8s.io/v1
kind: CredentialProviderConfig
providers:
  - name: acr-credential-provider
    matchImages:
      - "*.azurecr.io"
      - "*.azurecr.cn"
      - "*.azurecr.de"
      - "*.azurecr.us"
    defaultCacheDuration: "10m" // follow in-tree setting
    apiVersion: credentialprovider.kubelet.k8s.io/v1
    args:
      - /etc/kubernetes/azure.json
 ```

Fixes #

```release-note
fix(credential-provider): check acr pattern to avoid spoofing
```

```docs

```

### DIFF
--- a/pkg/credentialprovider/azure_credentials.go
+++ b/pkg/credentialprovider/azure_credentials.go
@@ -226,7 +226,7 @@ func (a *acrProvider) getFromACR(ctx context.Context, loginServer string) (strin
 func (a *acrProvider) parseACRLoginServerFromImage(image string) (string, string) {
 	targetImage, sourceRegistry := a.processImageWithRegistryMirror(image)
 	targetRegistry := acrRE.FindString(targetImage)
-	imageWithoutRegistry := strings.Trim(targetImage, targetRegistry)
+	imageWithoutRegistry := strings.TrimPrefix(targetImage, targetRegistry)
 	// for non customer cloud case, return registry only when:
 	//   - the acr pattern match
 	//   - the left string is empty or // credential provider authenticates the image pull request, but not validates the existence of the image
--- a/pkg/credentialprovider/azure_credentials.go
+++ b/pkg/credentialprovider/azure_credentials.go
@@ -47,7 +47,8 @@ const (
 
 var (
 	containerRegistryUrls = []string{"*.azurecr.io", "*.azurecr.cn", "*.azurecr.de", "*.azurecr.us"}
-	acrRE                 = regexp.MustCompile(`^.+?\.(azurecr\.io|azurecr\.cn|azurecr\.de|azurecr\.us)`)
+	// a valid acr image starts with alphanumerics, followed by corresponding acr domain name.
+	acrRE = regexp.MustCompile(`^[a-zA-Z0-9]+\.(azurecr\.io|azurecr\.cn|azurecr\.de|azurecr\.us)`)
 )
 
 // CredentialProvider is an interface implemented by the kubelet credential provider plugin to fetch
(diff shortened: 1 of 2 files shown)


mainred:
@bahe-msft I did not use the function you suggested https://github.com/kubernetes/kubernetes/blob/2bda5dd8c794d933145938affca30f012986e8fa/pkg/credentialprovider/keyring.go#L184-L194 because when it comes down credential provider from kubelet, it's a determined acr image, so we just need to follow  acr pattern to do the match and validation

pkg/credentialprovider/azure_credentials.go
```diff
@@ -224,10 +225,16 @@ func (a *acrProvider) getFromACR(ctx context.Context, loginServer string) (strin
...
-		targetRegistry := match[0]
+	targetRegistry := acrRE.FindString(targetImage)
+	imageWithoutRegistry := strings.Trim(targetImage, targetRegistry)
+	// for non customer cloud case, return registry only when:
+	//   - the acr pattern match
+	//   - the left string is empty or // credential provider authenticates the image pull request, but not validates the existence of the image
+	//   - the left string starts with a image repository, lead by "/"
+	// foo.azurecr.io/bar/image:version -> foo.azurecr.io
+	// foo.azurecr.io -> not match
+	// foo.azurecr.io.example -> not match
```
feiskyer, line 236: we'd better log an info in L253 for not supported images for troubleshooting

feiskyer: oh, omit it. We already have L122

feiskyer:
Looks good, let's wait for e2e validation results

pkg/credentialprovider/azure_credentials.go
```diff
@@ -224,10 +225,16 @@ func (a *acrProvider) getFromACR(ctx context.Context, loginServer string) (strin
...
 	targetImage, sourceRegistry := a.processImageWithRegistryMirror(image)
-
-	match := acrRE.FindAllString(targetImage, -1)
-	if len(match) == 1 {
-		targetRegistry := match[0]
+	targetRegistry := acrRE.FindString(targetImage)
+	imageWithoutRegistry := strings.Trim(targetImage, targetRegistry)
+	// for non customer cloud case, return registry only when:
+	//   - the acr pattern match
+	//   - the left string is empty or // credential provider authenticates the image pull request, but not validates the existence of the image
```
mainred, line 232: fortunately, use acr with repo will raise authorization failure because kubelet will pull this kind of image from docker hub

Failed to pull image "qingchuanhao.azurecr.io": failed to pull and unpack image "docker.io/library/qingchuanhao.azurecr.io:latest": failed to resolve reference "docker.io/library/qingchuanhao.azurecr.io:latest": pull access denied, repository does not exist or may require authorization: server message: insufficient_scope: authorization failed

mainred:
```
  containers:
  - name: container
    image: qingchuanhao.azurecr.io
```

pkg/credentialprovider/azure_credentials_test.go
```diff
@@ -205,8 +205,8 @@ func TestProcessImageWithMirrorMapping(t *testing.T) {
 	for _, test := range testcases {
 		t.Run(test.description, func(t *testing.T) {
 			targetloginServer, sourceloginServer := acrProvider.parseACRLoginServerFromImage(test.image)
-			assert.Equal(t, targetloginServer, test.expectedLoginServer)
```
mainred, line 208: For assert.Equal, expected variable comes before actual variable.
https://github.com/stretchr/testify/blob/master/assert/assertions.go#L460

pkg/credentialprovider/azure_credentials.go
```diff
@@ -224,10 +225,16 @@ func (a *acrProvider) getFromACR(ctx context.Context, loginServer string) (strin
 // If the provided image is not an acr image, this function will return an empty string.
 func (a *acrProvider) parseACRLoginServerFromImage(image string) (string, string) {
 	targetImage, sourceRegistry := a.processImageWithRegistryMirror(image)
-
-	match := acrRE.FindAllString(targetImage, -1)
-	if len(match) == 1 {
-		targetRegistry := match[0]
+	targetRegistry := acrRE.FindString(targetImage)
+	imageWithoutRegistry := strings.Trim(targetImage, targetRegistry)
```
bcho, line 229: is `strings.Trim` here expected? https://pkg.go.dev/strings#Trim It's to remove every character in the 2nd parameter from the leading/trailing of the first parameter

mainred: The aim is to remove the leading targetRegistry from targetImage, i.e. to remove acr from the image.
Updated the PR to use TrimPrefix to be specific.

pkg/credentialprovider/azure_credentials_test.go
```diff
@@ -261,11 +261,23 @@ func TestParseACRLoginServerFromImage(t *testing.T) {
...
+			expected: "",
+		},
+		{
+			image:    "docker/foo.azurecr.io/bar/image:version",
+			expected: "",
+		},
+		{
+			image:    "foo.azurecr.io",
+			expected: "foo.azurecr.io",
+		},
```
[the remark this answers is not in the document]
mainred: Done

bcho (approved):

feiskyer (commented):
Thanks for the fix

/lgtm
/approve

PR #8375, opened 2025-02-19
Set auxiliary token policy when auxiliary token provider enabled

zarvd:
#### What type of PR is this?

/kind feature

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/provider/azure.go
+++ b/pkg/provider/azure.go
@@ -25,6 +25,7 @@ import (
 	"time"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
 
 	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


MartinForReal:
Is this feature only used in cloud provider? I'm wondering if we can move it to https://github.com/kubernetes-sigs/cloud-provider-azure/blob/master/pkg/azclient/auth.go

zarvd:
> Is this feature only used in cloud provider? I'm wondering if we can move it to https://github.com/kubernetes-sigs/cloud-provider-azure/blob/master/pkg/azclient/auth.go

Yes. Will refactor the azclient pkg after testing but not in this PR.

feiskyer (approved):
LGTM, thanks for the fix

/lgtm

feiskyer:
@zarvd could you add a new e2e test to cover this scenario? It could be a following PR.

PR #8464, opened 2025-02-27
Container Load Balancer - DiffTracker Integration

georgeedward2000:
#### What type of PR is this?
/kind features

Fixes #

```release-note

```

```docs

```


pkg/provider/azure.go
```diff
@@ -495,9 +498,41 @@ func (az *Cloud) InitializeCloudFromConfig(ctx context.Context, config *config.C
...
+		NATGateways:   utilsets.NewString(),
+		NRPLocations:  make(map[string]difftracker.NRPLocation),
+	}
+
+	// Initialize the diff tracker state and get the necessary operations to sync the cluster with NRP
+	diffTrackerState, syncOperations := difftracker.InitializeDiffTrackerState(K8sState, NRPState)
+	az.diffTracker = diffTrackerState
+	// Print syncOps to surpress unused variable error
+	klog.V(2).Infof("Sync operations: %v", syncOperations)
+	// TODO (enechitoaia): call NRP APIs (including ServiceGateway) to update the state of the NRP
```
kartickmsft, line 526: I think better to list a TODO for the k8 API calls as well here. The actual implementation could be done in a different function to keep it modular.

pkg/provider/config/azure.go
```diff
@@ -183,6 +183,10 @@ func (az *Config) IsLBBackendPoolTypeNodeIP() bool {
 	return strings.EqualFold(az.LoadBalancerBackendPoolConfigurationType, consts.LoadBalancerBackendPoolConfigurationTypeNodeIP)
 }
 
+func (az *Config) IsLBBackendPoolTypePodIP() bool {
+	return strings.EqualFold(az.LoadBalancerBackendPoolConfigurationType, consts.LoadBalancerBackendPoolConfigurationTypeNodeIP) && az.UseStandardV2LoadBalancer()
```
kartickmsft, line 187: BackendPoolConfigurationType should be PodIP.

pkg/provider/difftracker/k8s_state_updates.go
```diff
@@ -0,0 +1,143 @@
...
+
+func (dt *DiffTrackerState) UpdateK8service(input UpdateK8sResource) error {
+	return updateK8Resource(input, dt.K8s.Services, ResourceTypeService)
+}
+
+func (dt *DiffTrackerState) UpdateK8sEgress(input UpdateK8sResource) error {
+	return updateK8Resource(input, dt.K8s.Egresses, ResourceTypeEgress)
+}
+
+func updateK8Resource(input UpdateK8sResource, set *utilsets.IgnoreCaseSet, resourceType string) error {
```
kartickmsft, line 22: I think utilsets.IgnoreCaseSet is not thread-safe. Better to make the update functions thread-safe as multiple service reconcile threads could update in parallel. Also, a read maybe happening in batch processor go routine when write via update function is taking place.

pkg/provider/difftracker/k8s_state_updates.go
```diff
@@ -0,0 +1,143 @@
...
+		}
+		pod.InboundIdentities.Insert(input.InboundIdentity)
+	}
+
+	for address, location := range input.OldAddresses {
+		if _, exists := input.NewAddresses[address]; exists {
+			continue
+		}
+
+		if location == "" {
```
kartickmsft, line 73: Can be the first check. Continue.

pkg/provider/difftracker/types.go
```diff
@@ -0,0 +1,146 @@
...
+	Removals  *utilsets.IgnoreCaseSet
+}
+
+type SyncDiffTrackerStateReturnType struct {
+	SyncStatus          SyncStatus
+	LoadBalancerUpdates SyncNRPServicesReturnType
+	NATGatewayUpdates   SyncNRPServicesReturnType
+	LocationData        LocationData
+}
+
```
kartickmsft, line 124: I think we need DTOs for Service(referring to LB/NATGW) also.

pkg/provider/difftracker/types.go
```diff
@@ -0,0 +1,146 @@
...
+type NRPAddress struct {
+	NRPServices *utilsets.IgnoreCaseSet // all inbound and outbound identities
+}
+
+type NRPLocation struct {
+	NRPAddresses map[string]NRPAddress
+}
+
+type NRPState struct {
+	LoadBalancers *utilsets.IgnoreCaseSet
```
kartickmsft, line 42: I'm wondering if NRPState should represent Service state in NRP. Service state in NRP will have references to LBBackendpool/NATGWs. To fit into the NRP API for Service/Location, we may need to store more information like SGW_name etc so that we can construct the URI.

pkg/provider/config/azure.go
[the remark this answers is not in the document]
georgeedward2000: let's discuss

pkg/provider/difftracker/types.go
georgeedward2000: let's discuss

pkg/provider/difftracker/types.go
georgeedward2000: let's discuss

pkg/provider/config/azure.go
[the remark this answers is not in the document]
kartickmsft: I think it's better to throw validation error in aks-rp before we pass on the settings to cloud-provider from aks-rp. There may be self-managed clusters wherein, aks-rp is not involved. So, maybe, throw validation error during cloud-provider init so that cloud-provider init fails and customer corrects the config.

pkg/provider/difftracker/handlers.go
```diff
@@ -0,0 +1,37 @@
+package difftracker
+
+func (dt *DiffTracker) handleService(input UpdateK8sResource) SyncServicesReturnType {
+	dt.mu.Lock()
```
kartickmsft, line 4: I think better to move the locks within the UpdateK8service. Same comment applies for the other handlers.

pkg/provider/difftracker/handlers.go
```diff
@@ -0,0 +1,37 @@
...
+	defer dt.mu.Unlock()
+
+	dt.UpdateK8service(input)
+	SyncDiffTrackerReturnType := dt.GetSyncLoadBalancerServices()
```
kartickmsft, line 8: I think we may not need the Diff. It's just update K8 state. So, I think we don't need the Get functions and as a result, locks can be made more granular.

pkg/provider/difftracker/sync_operations.go
```diff
@@ -0,0 +1,148 @@
...
+			continue
+		}
+		syncServices.Removals.Insert(service)
+		klog.V(2).Infof("Removed service %s from syncing object\n", service)
+	}
+
+	return syncServices
+}
+
+func (dt *DiffTracker) GetSyncLoadBalancerServices() SyncServicesReturnType {
```
kartickmsft, line 34: I think better to have lock for Get Diff functions as UpdateK8 state and GetDiff will be called independently from reconcile and batchprocessor threads respectively.

pkg/provider/difftracker/nrp_state_updates.go
```diff
@@ -0,0 +1,100 @@
+package difftracker
+
+import (
+	"fmt"
+
+	"k8s.io/klog/v2"
+	utilsets "sigs.k8s.io/cloud-provider-azure/pkg/util/sets"
+)
+
+func (dt *DiffTracker) UpdateNRPLoadBalancers() {
```
kartickmsft, line 10: As discussed offline, we need to pass in the Diff State that needs to be updated, as current implementation will sync NRP to the latest k8 state. Not the k8 state that we updated in NRP. Same comment applies for all Update functions.

pkg/provider/difftracker/nrp_state_updates.go
```diff
@@ -0,0 +1,100 @@
...
+	dt.mu.Lock()
+	defer dt.mu.Unlock()
+
+	for _, service := range dt.K8sResources.Services.UnsortedList() {
+		if dt.NRPResources.LoadBalancers.Has(service) {
+			continue
+		}
+		dt.NRPResources.LoadBalancers.Insert(service)
```
kartickmsft, line 18: What happens if we insert a duplicate entry? I guess it will just replace with the latest entry, which won't impact any impact if the entry is already present. If my understanding is correct, then we can remove the find in line 15. Will avoid cpu cycles for the new entries.

pkg/provider/difftracker/sync_operations.go
```diff
@@ -0,0 +1,148 @@
...
+	// Iterate over all nodes in the K8s state
+	for nodeIp, node := range dt.K8sResources.Nodes {
+		nrpLocation, exists := dt.NRPResources.Locations[nodeIp]
+		location := initializeLocation(exists)
+
+		for address, pod := range node.Pods {
+			serviceRef := createServiceRef(pod)
+			addressData := Address{ServiceRef: serviceRef}
+
+			if !exists || !serviceRef.Equals(nrpLocation.Addresses[address].Services) {
```
kartickmsft, line 59: Shouldn't we check if nrpLocation.Addresses[address] is present before comparing services within the address?

david-kow (commented):
some thoughts

pkg/provider/difftracker/util.go
```diff
@@ -0,0 +1,382 @@
...
+func MapLoadBalancerUpdatesToServiceDataDTO(loadBalancerUpdates SyncServicesReturnType) ServiceDataDTO {
+	var serviceDataDTO ServiceDataDTO
+	serviceDataDTO.Action = PartialUpdate
+	serviceDataDTO.Services = []ServiceDTO{}
+	for _, service := range loadBalancerUpdates.Additions.UnsortedList() {
+		serviceDTO := ServiceDTO{
+			Service: service,
+			LoadBalancerBackendPools: []LoadBalancerBackendPoolDTO{
+				{
+					Id: fmt.Sprintf("%s-backendpool", service),
```
kartickmsft, line 344: I think we need to specify the full uri of the LB backendpool resource in NRP.

pkg/provider/difftracker/util.go
```diff
@@ -0,0 +1,382 @@
...
+// Map NATGatewayUpdates to ServiceDataDTO to be used as payload in ServiceGateway API calls
+func MapNATGatewayUpdatesToServiceDataDTO(natGatewayUpdates SyncServicesReturnType) ServiceDataDTO {
+	var serviceDataDTO ServiceDataDTO
+	serviceDataDTO.Action = PartialUpdate
+	serviceDataDTO.Services = []ServiceDTO{}
+	for _, service := range natGatewayUpdates.Additions.UnsortedList() {
+		serviceDTO := ServiceDTO{
+			Service: service,
+			PublicNatGateway: NatGatewayDTO{
+				Id: fmt.Sprintf("%s-natgateway", service),
```
kartickmsft, line 369: I think we need to specify the full URI of NATGateway resource.

pkg/provider/difftracker/nrp_state_updates.go
```diff
@@ -0,0 +1,107 @@
...
+			if isFullUpdate || !addressExists(nrpLocation, addressKey) {
+				nrpLocation.Addresses[addressKey] = NRPAddress{Services: serviceRefs}
+				continue
+			}
+
+			// For partial updates with existing address
+			existingAddress := nrpLocation.Addresses[addressKey]
+			if !serviceRefs.Equals(existingAddress.Services) {
+				existingAddress.Services = serviceRefs
+				nrpLocation.Addresses[addressKey] = existingAddress
```
kartickmsft, line 81: I think we don't need to add the existingAddress back as it's already retrieved in line 78.

pkg/provider/difftracker/types.go
```diff
@@ -0,0 +1,175 @@
...
+	InboundIdentities       *utilsets.IgnoreCaseSet
+	PublicOutboundIdentity  string
+	PrivateOutboundIdentity string
+}
+
+type Node struct {
+	Pods map[string]Pod
+}
+
+type K8s struct {
```
kartickmsft, line 61: Should we rename it to K8s_State and NRP_State for better readability?

pkg/provider/difftracker/types.go
```diff
@@ -0,0 +1,175 @@
...
+
+// LocationDTO represents the DTO for Location
+type LocationDTO struct {
+	Location            string       `json:"Location"`
+	AddressUpdateAction UpdateAction `json:"AddressUpdateAction"`
+	Addresses           []AddressDTO `json:"Addresses"`
+}
+
+// LocationDataDTO represents the DTO for LocationData
+type LocationDataDTO struct {
```
kartickmsft, line 148: LocationsDataDTO is better?

pkg/provider/difftracker/util.go
```diff
@@ -0,0 +1,394 @@
...
+
+func (pod *Pod) HasIdentities() bool {
+	return pod.InboundIdentities.Len() > 0 || pod.PublicOutboundIdentity != "" || pod.PrivateOutboundIdentity != ""
+}
+
+// DeepEqual compares the K8s and NRP states to check if they are in sync
+func (dt *DiffTracker) DeepEqual() bool {
+	// Compare Services with LoadBalancers
+	if dt.K8sResources.Services.Len() != dt.NRPResources.LoadBalancers.Len() {
+		klog.Errorf("Services and LoadBalancers length mismatch")
```
kartickmsft, line 63: Mismatch between k8 and NRP is not really an error. It could be just a transient state where we didn't sync NRP with k8 yet. So, suggesting to not log mismatches as an error in this function.

pkg/provider/difftracker/util.go
```diff
@@ -0,0 +1,394 @@
...
+				klog.Errorf("Combined identities length mismatch for pod %s in node %s\n", podKey, nodeKey)
+				return false
+			}
+
+			for _, identity := range combinedIdentities {
+				if !nrpAddress.Services.Has(identity) {
+					klog.Errorf("Identity %s not found in Services for pod %s in node %s\n", identity, podKey, nodeKey)
+					return false
+				}
+			}
```
kartickmsft, line 141: Shouldn't we check if services in nrp are different from k8?

pkg/provider/difftracker/util.go
```diff
@@ -0,0 +1,394 @@
...
+		serviceDTO := ServiceDTO{
+			Service: service,
+			LoadBalancerBackendPools: []LoadBalancerBackendPoolDTO{
+				{
+					Id: fmt.Sprintf(
+						consts.BackendPoolIDTemplate,
+						subscriptionID,
+						resourceGroup,
+						service,
+						fmt.Sprintf("%s-backendpool", service),
```
kartickmsft, line 350: Please check the name of the backendpool that's used for CLB in createOrUpdateBackendPool.

kartickmsft: Type property has been introduced for Service. We could pass Type as "Inbound" as we are mapping LB to Service.

pkg/provider/difftracker/util.go
```diff
@@ -0,0 +1,394 @@
...
+
+// Map NATGatewayUpdates to ServiceDataDTO to be used as payload in ServiceGateway API calls
+func MapNATGatewayUpdatesToServiceDataDTO(natGatewayUpdates SyncServicesReturnType, subscriptionID string, resourceGroup string) ServiceDataDTO {
+	var serviceDataDTO ServiceDataDTO
+	serviceDataDTO.Action = PartialUpdate
+	serviceDataDTO.Services = []ServiceDTO{}
+	for _, service := range natGatewayUpdates.Additions.UnsortedList() {
+		serviceDTO := ServiceDTO{
+			Service: service,
+			PublicNatGateway: NatGatewayDTO{
```
kartickmsft, line 375: Type property has been introduced for Service. We could pass Type as "Outbound" as we are mapping LB to Service.

pkg/provider/difftracker/util.go
georgeedward2000: It checks on line 72 (Inbound) and line 90 (Outbound) for services in NRP_State and not in K8s_State

pkg/provider/difftracker/util.go
georgeedward2000: This is what I found in a test for createOrUpdateBackendPool, which follows the same structure as the one I used:
`ptr.To(fmt.Sprintf("/subscriptions/subscriptionID/resourceGroups/rg/providers/Microsoft.Network/loadBalancers/%s/backendAddressPools/%s", lbName, bpName)),`

kartickmsft: I thought line 350 should be fmt.Sprintf("%s",service). I mean not "%s-backendpool".

PR #8594, opened 2025-03-12
Merge OTEL metrics with legacy registerer

zarvd:
#### What type of PR is this?

/kind feature

To export all metrics in 1 endpoint.

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/cmd/cloud-controller-manager/app/controllermanager.go
+++ b/cmd/cloud-controller-manager/app/controllermanager.go
@@ -323,7 +323,10 @@ func StartHTTPServer(ctx context.Context, c *cloudcontrollerconfig.CompletedConf
 	if c.SecureServing != nil {
 		unsecuredMux := genericcontrollermanager.NewBaseHandler(&c.ComponentConfig.Generic.Debugging, healthzHandler)
 
-		unsecuredMux.Handle("/metrics/v2", traceProvider.MetricsHTTPHandler()) // Add metricsv2 endpoint
+		const MetricsPath = "/metrics"
+		unsecuredMux.Unregister(MetricsPath) // Unregisterer handler of legacy registerer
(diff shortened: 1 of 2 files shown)


cmd/cloud-controller-manager/app/controllermanager.go
```diff
@@ -323,7 +323,10 @@ func StartHTTPServer(ctx context.Context, c *cloudcontrollerconfig.CompletedConf
 	if c.SecureServing != nil {
 		unsecuredMux := genericcontrollermanager.NewBaseHandler(&c.ComponentConfig.Generic.Debugging, healthzHandler)
 
-		unsecuredMux.Handle("/metrics/v2", traceProvider.MetricsHTTPHandler()) // Add metricsv2 endpoint
+		const MetricsPath = "/metrics"
+		unsecuredMux.Unregister(MetricsPath) // Unregisterer handler of legacy registerer
```
mainred, line 327: Can we just not register the metrics path with old register?

mainred (commented):
Not quite familiar withe old register we have, but left a comment.
And can you list the OTEL metrics from after merging the metrics from legacy register as a test?

cmd/cloud-controller-manager/app/controllermanager.go
zarvd: > Can we just not register the metrics path with old register?

It’s better to stick with the old path for the new metrics since it makes scrape config maintenance easier. Plus, separating the two has already been tested for a while.

> And can you list the OTEL metrics from after merging the metrics from legacy register as a test?

Already tested it in the standalone. Will skip pasting it here as it's a large volume of data.

feiskyer: yes, agreed

zarvd:
@cvvz PTAL

andyzhangx (approved):
/lgtm

PR #8730, opened 2025-03-25
[azclient] Return default auth provider if no valid auth method

zarvd:
#### What type of PR is this?

/kind cleanup

keep the behavior the same as last version

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/azclient/auth.go
+++ b/pkg/azclient/auth.go
@@ -68,7 +68,9 @@ func NewAuthProvider(
 	case len(config.AADMSIDataPlaneIdentityPath) > 0:
 		return newAuthProviderWithUserAssignedIdentity(config, clientOption, opts)
 	default:
-		return nil, ErrNoValidAuthMethodFound
+		return &AuthProvider{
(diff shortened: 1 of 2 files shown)


pkg/azclient/auth.go
```diff
@@ -68,7 +68,9 @@ func NewAuthProvider(
 	case len(config.AADMSIDataPlaneIdentityPath) > 0:
 		return newAuthProviderWithUserAssignedIdentity(config, clientOption, opts)
 	default:
```
feiskyer, line 70: could you add a log for this scenario?

zarvd: Will add the log (when compute / network credential both nil) in the cloud-provider when upgrading the pkg in there. The azclient doesn't enable logging.  cc @MartinForReal

feiskyer: Sure, that makes more sense!

feiskyer (commented):
/lgtm
/approve

PR #9026, opened 2025-05-07
feat: add option to override ARM retry-after if the value is lower than the configured minimum

sunpa93:
#### What type of PR is this?
/kind feature

There is a CRP Extension bug that adversely impacts csi-driver's attach/detach latency. The symptom of the issue is that a high retry-after value of 37 seconds is returned on 4th GET request for async operation tracking. To mitigate the impact, we need to issue GET requests with more delay in between. So, this PR adds option to add custom retry policy that will remove the retry-after header from the http response if the value is lower than the configured minimum.

Example HTTP Header:
Headers:
{
  Retry-After: 37
  x-ms-ratelimit-remaining-resource: Microsoft.Compute/GetOperationResource;41
  x-ms-ratelimit-remaining-resource: Microsoft.Compute/GetOperationSubscriptionMaximum;14994
  x-ms-need-to-refresh-epl-cache: False
  Content-Type: application/json; charset=utf-8
}

Fixes #

```release-note
feat: add option to override ARM retry-after if the value is lower than the configured minimum
```

```docs
feat: add option to override ARM retry-after if the value is lower than the configured minimum
```

### DIFF
--- a/pkg/azclient/arm_conf.go
+++ b/pkg/azclient/arm_conf.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
 
+	"sigs.k8s.io/cloud-provider-azure/pkg/azclient/policy/retryaftermin"
 	"sigs.k8s.io/cloud-provider-azure/pkg/azclient/policy/useragent"
 	"sigs.k8s.io/cloud-provider-azure/pkg/azclient/utils"
 )
@@ -49,6 +50,8 @@ type ARMClientConfig struct {
 	// when setting AzureAuthConfig.Cloud with "AZURESTACKCLOUD" to customize ARM endpoints
 	// while the cluster is not running on AzureStack.
 	DisableAzureStackCloud bool `json:"disableAzureStackCloud,omitempty" yaml:"disableAzureStackCloud,omitempty"`
+	// If true, HTTP responses' retry-after header will be overridden with the configured minimum retry-after value if lower than the configured minimum
+	EnableMinimumRetryAfter bool `json:"enableMinimumRetryAfter,omitempty" yaml:"enableMinimumRetryAfter,omitempty"`
 }
 
 func (config *ARMClientConfig) GetTenantID() string {
@@ -80,6 +83,10 @@ func GetAzCoreClientOption(armConfig *ARMClientConfig) (*policy.ClientOptions, *
 		}
 		if armConfig.CloudProviderBackoff && armConfig.CloudProviderBackoffRetries > 0 {
 			clientConfig.Retry.MaxRetries = armConfig.CloudProviderBackoffRetries
+		}
+		if armConfig.EnableMinimumRetryAfter {
+			// Add the minimum retry-after policy to enforce a a minimum retry-after value configured in clientConfig.Retry.RetryDelay (default 5s)
+			clientConfig.PerRetryPolicies = append(clientConfig.PerRetryPolicies, retryaftermin.NewRetryAfterMinPolicy(clientConfig.Retry.RetryDelay))
 		}
 	}
 	return &clientConfig, env, nil
--- a/pkg/azclient/arm_conf.go
+++ b/pkg/azclient/arm_conf.go
@@ -85,7 +85,7 @@ func GetAzCoreClientOption(armConfig *ARMClientConfig) (*policy.ClientOptions, *
 			clientConfig.Retry.MaxRetries = armConfig.CloudProviderBackoffRetries
 		}
 		if armConfig.EnableMinimumRetryAfter {
-			// Add the minimum retry-after policy to enforce a a minimum retry-after value configured in clientConfig.Retry.RetryDelay (default 5s)
+			// Add the minimum retry-after policy to enforce a minimum retry-after value configured in clientConfig.Retry.RetryDelay (default 5s)
 			clientConfig.PerRetryPolicies = append(clientConfig.PerRetryPolicies, retryaftermin.NewRetryAfterMinPolicy(clientConfig.Retry.RetryDelay))
 		}
 	}
--- a/pkg/azclient/policy/retryaftermin/retryaftermin.go
+++ b/pkg/azclient/policy/retryaftermin/retryaftermin.go
@@ -1,0 +1,86 @@
+/*
+Copyright 2025 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package retryaftermin
+
+import (
+	"net/http"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
+	"k8s.io/klog/v2"
+)
+
+// RetryAfterMinPolicy is a policy that enforces a minimum retry-after value
+type RetryAfterMinPolicy struct {
+	minRetryAfter time.Duration
+}
+
+// NewRetryAfterMinPolicy creates a new RetryAfterMinPolicy with the specified minimum retry-after value
+func NewRetryAfterMinPolicy(minRetryAfter time.Duration) policy.Policy {
+	return &RetryAfterMinPolicy{
+		minRetryAfter: minRetryAfter,
+	}
+}
+
+// GetMinRetryAfter returns the minimum retry-after value
+func (p *RetryAfterMinPolicy) GetMinRetryAfter() time.Duration {
+	return p.minRetryAfter
+}
+
+// Do implements the policy.Policy interface
+func (p *RetryAfterMinPolicy) Do(req *policy.Request) (*http.Response, error) {
+	resp, err := req.Next()
+	// If the request failed or the status code is >= 300, return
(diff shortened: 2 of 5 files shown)


go.mod
```diff
@@ -135,3 +135,5 @@ require (
 	sigs.k8s.io/randfill v1.0.0 // indirect
 	sigs.k8s.io/structured-merge-diff/v4 v4.6.0 // indirect
 )
+
+replace sigs.k8s.io/cloud-provider-azure/pkg/azclient => github.com/sunpa93/cloud-provider-azure/pkg/azclient v0.0.0-20250507210748-afe521f58513
```
feiskyer, line 139: could you remove the replace here?

andyzhangx:
@sunpa93 thanks for the PR, can you revert changes in vendor dir? it's not necessary to make duplicated changes in two folder, you only need to make change in non-vendor dir, and after PR merged, CSI driver would then upgrade azclient directly

pkg/azclient/arm_conf.go
```diff
@@ -81,6 +83,10 @@ func GetAzCoreClientOption(armConfig *ARMClientConfig) (*policy.ClientOptions, *
 		if armConfig.CloudProviderBackoff && armConfig.CloudProviderBackoffRetries > 0 {
 			clientConfig.Retry.MaxRetries = armConfig.CloudProviderBackoffRetries
 		}
+		if armConfig.EnableRetryAfterOverride {
+			// Add the minimum retry-after policy to enforce a minimum of 5 seconds
```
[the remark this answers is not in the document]
sunpa93: Thanks for pointing that out. Updated the comment to remove confusion. I think we should use the `RetryDelay` in `clientConfig` to compare against the http retry-after, given that this `RetryDelay` will be what will be used in the polling function will be using.

pkg/azclient/arm_func.go
```diff
@@ -0,0 +1,80 @@
...
+				overrideRetryAfter(resp.Header, headerName, retryDuration)
+			} else {
+				// Try to parse as seconds (integer)
+				seconds, err := strconv.Atoi(retryAfter)
+				if err == nil {
+					retryDuration := time.Duration(seconds) * time.Second
+					// If the retry-after value is less than the minimum, remove it
+					overrideRetryAfter(resp.Header, headerName, retryDuration)
+				} else {
+					klog.V(5).Infof("RetryAfterMinPolicy: not modifying %s header with unrecognized format: %s", headerName, retryAfter)
```
sunpa93, line 73: @andyzhangx Is verbosity level okay or should this be lowered?

pkg/azclient/arm_func.go
```diff
@@ -0,0 +1,80 @@
...
+
+// Do implements the policy.Policy interface
+func (p *RetryAfterMinPolicy) Do(req *policy.Request) (*http.Response, error) {
+	resp, err := req.Next()
+	if err != nil || resp == nil {
+		return resp, err
+	}
+
+	// Check if the response retry-after header is less than the minimum
+	overrideRetryAfter := func(header http.Header, headerName string, retryAfter time.Duration) {
```
andyzhangx, line 49: I think if it's error(e.g. throttling error), we should respect the retry-after, we only set minimum retry when it's success

pkg/azclient/arm_func.go
andyzhangx: that's ok

pkg/azclient/arm_func.go
```diff
@@ -0,0 +1,80 @@
...
+
+// NewRetryAfterMinPolicy creates a new RetryAfterMinPolicy with the specified minimum retry-after value
+func NewRetryAfterMinPolicy(minSeconds time.Duration) policy.Policy {
+	return &RetryAfterMinPolicy{
+		minRetryAfter: minSeconds,
+	}
+}
+
+// Do implements the policy.Policy interface
+func (p *RetryAfterMinPolicy) Do(req *policy.Request) (*http.Response, error) {
```
andyzhangx, line 42: pls also write unit test for this function

andyzhangx:
and why there is change in `pkg/provider/azure_controller_vmssflex.go`? this is too big change I cannot open in my browser, I think you could split into two commits 1) changes in pkg/azclient/* 2) vendor change, thus I only need to review commit # 1

pkg/provider/azure_controller_vmssflex.go
```diff
@@ -119,7 +119,7 @@ func (fs *FlexScaleSet) AttachDisk(ctx context.Context, nodeName types.NodeName,
 	klog.V(2).Infof("azureDisk - update(%s): vm(%s) - attach disk list(%+v) returned with %v", nodeResourceGroup, vmName, diskMap, rerr)
 
 	if err == nil && result != nil {
-		if rerr := fs.updateCache(ctx, vmName, result); err != nil {
+		if rerr := fs.updateCache(ctx, vmName, result); rerr != nil {
```
andyzhangx, line 122: good catch, pls make change in another PR, and btw, vmssflex is not supported on aks

pkg/azclient/arm_func.go
sunpa93: Added condition to check for the status code.

pkg/azclient/arm_func.go
sunpa93: Done. Also, moved RetryAfterMinPolicy to azclient/policy directory as that's where the policy implementations and tests should be.

pkg/provider/azure_controller_vmssflex.go
sunpa93: Will do

sunpa93:
@andyzhangx As requested, have separated the changes into 3 commits: 1) feature change, 2) unit test addition, and 3) vendor update

andyzhangx:
can you remove the vendor folder under `pkg\azclient`? that's not needed

andyzhangx:
@sunpa93 thanks, there are some golint errors, could you fix it?
```
home/runner/work/cloud-provider-azure/cloud-provider-azure/pkg/azclient/bin/golangci-lint run --timeout 10m ./...
Error: policy/retryaftermin/retryaftermin.go:30:6: exported: type name will be used as retryaftermin.RetryAfterMinPolicy by other packages, and that stutters; consider calling this Policy (revive)
type RetryAfterMinPolicy struct {
     ^
Error: policy/retryaftermin/retryaftermin_test.go:44:11: unused-parameter: parameter 'req' seems to be unused, consider removing or renaming it as _ (revive)
					func(req *policy.Request) (*http.Response, error) {
					     ^
Error: policy/retryaftermin/retryaftermin_test.go:73:11: unused-parameter: parameter 'req' seems to be unused, consider removing or renaming it as _ (revive)
					func(req *policy.Request) (*http.Response, error) {
					     ^
Error: policy/retryaftermin/retryaftermin_test.go:103:11: unused-parameter: parameter 'req' seems to be unused, consider removing or renaming it as _ (revive)
					func(req *policy.Request) (*http.Response, error) {
					     ^
Error: policy/retryaftermin/retryaftermin_test.go:126:11: unused-parameter: parameter 'req' seems to be unused, consider removing or renaming it as _ (revive)
					func(req *policy.Request) (*http.Response, error) {
					     ^
Error: policy/retryaftermin/retryaftermin_test.go:[155](https://github.com/kubernetes-sigs/cloud-provider-azure/actions/runs/14940985477/job/41977932277?pr=9026#step:7:156):11: unused-parameter: parameter 'req' seems to be unused, consider removing or renaming it as _ (revive)
					func(req *policy.Request) (*http.Response, error) {
					     ^
make: *** [Makefile:133: vet-all] Error 1
```

andyzhangx:
btw, do you have a header string which contains `Retry-After`, if yes, just paste it in the PR description, thanks.

andyzhangx (approved):
/lgtm

andyzhangx:
a few hours after this PR merged, a new version of azclient would be created, you could run `go get -u sigs.k8s.io/cloud-provider-azure/pkg/azclient` in csi driver to get the azclient changes, no need to copy the change in this PR to csi driver vendor repo by yourself. thanks for the fix.

pkg/azclient/policy/retryaftermin/retryaftermin.go
```diff
@@ -0,0 +1,86 @@
...
+func (p *Policy) Do(req *policy.Request) (*http.Response, error) {
+	resp, err := req.Next()
+	// If the request failed or the status code is >= 300, return
+	if err != nil || resp == nil || resp.StatusCode >= 300 {
+		return resp, err
+	}
+
+	// Check if the response retry-after header is less than the minimum
+	overrideRetryAfter := func(header http.Header, headerName string, retryAfter time.Duration) {
+		if retryAfter < p.minRetryAfter {
```
andyzhangx, line 56: @sunpa93 
should it be `retryAfter > p.minRetryAfter`? in the PR description case, 37s is bigger than p.minRetryAfter, and I got some below logs, the small retryAfter (<5s) should be respected?

```
I0603 03:27:45.256145       1 retryaftermin.go:57] RetryAfterMinPolicy: retry-after value 4s is less than minimum 5s, removing retry-after header..
I0603 03:27:45.335463       1 retryaftermin.go:57] RetryAfterMinPolicy: retry-after value 2s is less than minimum 5s, removing retry-after header..
I0603 03:27:50.413500       1 retryaftermin.go:57] RetryAfterMinPolicy: retry-after value 2s is less than minimum 5s, removing retry-after header..
```

sunpa93: The goal here is to override the small retry-after values so that we issue the GETs less frequently and are less likely to hit the retry-after threshold that causes the VMSS service to return a large retry-after. I don't think we can ignore the large retry-after and keep spamming GET requests once throttled as the request will likely not get accepted.

andyzhangx: I see, it's replacing 2s with 5s in the PR, thanks for the explanations, I think finally we need that bug fix in crp extension and revert this tricy workaround, thanks for the work!

PR #9269, opened 2025-07-17
Update Helm chart to current patch releases

mboersma:
#### What type of PR is this?

/kind cleanup

Updates the cloud-provider-azure Helm chart to use the current patch releases.

Fixes #

```release-note
NONE
```


helm/cloud-provider-azure/templates/cloud-provider-azure.yaml
```diff
@@ -17,13 +17,13 @@
 {{- else if eq .Capabilities.KubeVersion.Minor "29" -}}
   {{- $_ := set . "cloudProviderAzureVersion" "v1.29.15" }}
 {{- else if eq .Capabilities.KubeVersion.Minor "30" -}}
-  {{- $_ := set . "cloudProviderAzureVersion" "v1.30.12" }}
+  {{- $_ := set . "cloudProviderAzureVersion" "v1.30.13" }}
```
nojnhuh: Sorry, just getting to this now. I see new patch releases for cloud-provider-azure are available now since this was first opened, v1.30.14 etc. Is it important that the chart doesn't skip any versions or could we go straight to those latest versions?

mboersma: I don't think there's a need to release a Chart for the now-old patch versions. I'll update this PR, thanks for noticing that!

mboersma: Updated.

nojnhuh (approved):
/lgtm

PR #9469, opened 2025-09-18
chore: enhance argument check and context for acr-credential-provider command

qweeah:
#### What type of PR is this?

/kind cleanup

This PR enhances argument check and context for acr-credential-provider command

n/a

```release-note
NONE
```

```docs

```

### DIFF
--- a/cmd/acr-credential-provider/main.go
+++ b/cmd/acr-credential-provider/main.go
@@ -44,7 +44,7 @@ func main() {
 		Use:   "acr-credential-provider configFile",
 		Short: "Acr credential provider for Kubelet",
 		Long:  `The acr credential provider is responsible for providing ACR credentials for kubelet`,
-		Args: func(cmd *cobra.Command, args []string) error {
+		Args: func(_ *cobra.Command, args []string) error {
 			if len(args) == 0 {
 				return errors.New("Config file is not specified")
 			}


cmd/acr-credential-provider/main.go
```diff
@@ -70,7 +77,11 @@ func main() {
 		"Mirror a source registry host to a target registry host, and image pull credential will be requested to the target registry host when the image is from source registry host")
 
 	logs.AddFlags(command.Flags())
-	if err := command.Execute(); err != nil {
+	if err := func() error {
```
mainred, line 80: I am interested in the coding style, why not simply
```
ctx, cancel := signal.NotifyContext(context.Background(), os.Interrupt)
defer cancel()
if err :=command.ExecuteContext(ctx); err != nil {
os.Exit(1)
}
```

cmd/acr-credential-provider/main.go
```diff
@@ -38,27 +41,31 @@ func main() {
...
+			}
+			return nil
+		},
+		Version: version.Get().GitVersion,
+		RunE: func(cmd *cobra.Command, args []string) error {
 			acrProvider, err := credentialprovider.NewAcrProviderFromConfig(args[0], RegistryMirrorStr)
 			if err != nil {
 				klog.Errorf("Failed to initialize ACR provider: %v", err)
-				os.Exit(1)
+				return err
```
mainred, line 61: I think returning error will lead the plugin to exiting with non-zero, right?

cmd/acr-credential-provider/main.go
qweeah: In your proposal, `os.Exit(1)` will skip the deferred call of `cancel()`. Using an anonymous function will avoid that.

cmd/acr-credential-provider/main.go
qweeah: The error will be passed on to the return value of `command.ExecuteContext` and finally causing os.Exit(1), which is equivalent to exiting with non-zero.

mainred (approved):

cmd/acr-credential-provider/main.go
qweeah: In my understanding, os.Exist(1) should be called somewhere and it is here.

mainred: Looks good, forget my last reply. I have removed my last comment for my misunderstanding the code.

PR #9687, opened 2025-11-27
feat: KSA support with AKS identity bindings

qweeah:
#### What type of PR is this?

/kind feature

This PR adds support for Kubernetes Service Account (KSA) based authentication with identity bindings for the ACR credential provider. This enables secret-less ACR image pull via the Kubernetes API server's identity bindings endpoint for token exchange.

**Key Changes:**

1. **Configuration Enhancement**:
   - Implemented comprehensive configuration parsing and validation in new `cmd/acr-credential-provider/pkg/config` package
   - Added CLI flags: `--ib-sni-name`, `--ib-default-client-id`, `--ib-default-tenant-id`, `--ib-apiserver-ip`

2. **Validation Logic**:
   - Protocol prefix validation: Rejects `https://` or `http://` prefixes in SNI name
   - IP address format validation using `net.ParseIP`
   - Mutual dependency enforcement between SNI name and API server IP
   - Validates client/tenant ID requires SNI name to be set

3. **Identity Bindings Token Credential**:
   - Implemented custom HTTP transport with SNI-based routing to configurable API server IP
   - Implemented OAuth2 client credentials flow with service account token exchange
   - Refactored to store token, client ID, and tenant ID directly in credential struct for improved performance

4. **Comprehensive Testing**:
   - Added unit tests for configuration parsing (`cmd/acr-credential-provider/pkg/config/identity_bindings_config_test.go`) with 100% coverage
   - Added unit tests for identity bindings credentials (`pkg/credentialprovider/identity_bindings_credentials_test.go`)

5. **Code Organization**:
   - Refactored configuration parsing from `main.go` to dedicated config package
   - Improved separation of concerns between CLI, configuration, and credential logic
   - Early resolution of client and tenant IDs from annotations or defaults during credential creation

Fixes #

- **API Server IP Configuration**: The implementation uses a configurable API server IP (instead of hardcoded constant) to allow flexibility for different cluster configurations, though most AKS clusters use `10.0.0.1` as the cluster-internal API server address.

- **SNI-Based Routing**: The custom dialer always resolves the SNI hostname to the configured API server IP, ensuring the TLS connection uses SNI for certificate validation while connecting to the correct cluster endpoint.

- **CA Certificate Rotation**: The transport implementation detects CA certificate changes by comparing file content and rebuilds the transport with updated CA pool when rotation is detected.

- **Validation Strategy**: The configuration validation enforces mutual dependencies (SNI name ↔ API server IP) to prevent misconfiguration. Protocol prefixes are rejected during parsing rather than being stripped.

- **Test Approach**: The test suite uses `httptest.NewUnstartedServer` with self-signed certificates to test the full mTLS flow including custom dialer, TLS configuration, and token exchange logic.

- **Tenant ID for Future SDK Compatibility**: The `--ib-default-tenant-id` parameter is included from day one to ensure smooth transition if the ACR credential provider switches to using the official Azure SDK in the future.

```release-note
ACR credential provider now supports Kubernetes Service Account (KSA) based authentication with identity bindings. Configure via flags: --ib-sni-name, --ib-apiserver-ip, --ib-default-client-id, and --ib-default-tenant-id.
```

### DIFF
--- a/cmd/acr-credential-provider/main.go
+++ b/cmd/acr-credential-provider/main.go
@@ -31,13 +31,20 @@ import (
 	"k8s.io/component-base/logs"
 	"k8s.io/klog/v2"
 
+	"sigs.k8s.io/cloud-provider-azure/cmd/acr-credential-provider/pkg/config"
 	"sigs.k8s.io/cloud-provider-azure/pkg/version"
 )
 
 func main() {
 	rand.Seed(time.Now().UnixNano())
 
-	var RegistryMirrorStr string
+	var (
+		RegistryMirrorStr string
+		IBSNIName         string
+		IBDefaultClient   string
+		IBDefaultTenant   string
+		IBAPIIP           string
+	)
 
 	command := &cobra.Command{
 		Use:   "acr-credential-provider configFile",
@@ -54,7 +61,12 @@ func main() {
 		},
 		Version: version.Get().GitVersion,
 		RunE: func(_ *cobra.Command, args []string) error {
-			if err := NewCredentialProvider(args[0], RegistryMirrorStr).Run(context.TODO()); err != nil {
+			ibConfig, err := config.ParseIdentityBindingsConfig(IBSNIName, IBDefaultClient, IBDefaultTenant, IBAPIIP)
+			if err != nil {
+				klog.Errorf("Error parsing identity bindings config: %v", err)
+				return err
+			}
+			if err := NewCredentialProvider(args[0], RegistryMirrorStr, ibConfig).Run(context.TODO()); err != nil {
 				klog.Errorf("Error running acr credential provider: %v", err)
 				return err
 			}
@@ -68,6 +80,14 @@ func main() {
 	// Flags
 	command.Flags().StringVarP(&RegistryMirrorStr, "registry-mirror", "r", "",
 		"Mirror a source registry host to a target registry host, and image pull credential will be requested to the target registry host when the image is from source registry host")
+	command.Flags().StringVar(&IBSNIName, config.FlagIBSNIName, "",
+		"SNI name for identity bindings")
+	command.Flags().StringVar(&IBDefaultClient, config.FlagIBDefaultClient, "",
+		"Default Azure AD client ID for identity bindings")
+	command.Flags().StringVar(&IBDefaultTenant, config.FlagIBDefaultTenant, "",
+		"Default Azure AD tenant ID for identity bindings")
+	command.Flags().StringVar(&IBAPIIP, config.FlagIBAPIIP, "",
+		"API server IP address for identity bindings endpoint")
 
 	logs.AddFlags(command.Flags())
 	if err := func() error {
--- a/pkg/credentialprovider/identity_bindings_credentials.go
+++ b/pkg/credentialprovider/identity_bindings_credentials.go
@@ -1,0 +1,267 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package credentialprovider
+
+import (
+	"context"
+	"crypto/tls"
+	"crypto/x509"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net"
+	"net/http"
+	"net/url"
+	"os"
+	"strings"
+	"time"
+
+	providerconfig "sigs.k8s.io/cloud-provider-azure/pkg/provider/config"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
+	"k8s.io/klog/v2"
+	v1 "k8s.io/kubelet/pkg/apis/credentialprovider/v1"
+)
+
+const (
+	// Kubernetes certificate path
+	KubernetesCACertPath = "/etc/kubernetes/certs/ca.crt"
+)
+
+// identityBindingsTokenCredential implements azcore.TokenCredential interface
+// using identity bindings token exchange
+type identityBindingsTokenCredential struct {
+	token    string
+	clientID string
+	// tenantID is reserved for future SDK compatibility and may be used in token endpoint construction.
+	tenantID  string
+	config    *providerconfig.AzureClientConfig
+	ibConfig  IdentityBindingsConfig
+	endpoint  string
+	transport *http.Transport
+}
+
+// tokenResponse represents the response from identity bindings token exchange
+type tokenResponse struct {
+	AccessToken string `json:"access_token"`
+	ExpiresIn   int64  `json:"expires_in"`
+}
+
+// createTransport creates an HTTP transport with custom CA
+// The transport uses a custom dialer that resolves the SNI name to the configured API server IP
+func createTransport(sniName string, apiServerIP string, caPool *x509.CertPool) *http.Transport {
+	transport := http.DefaultTransport.(*http.Transport).Clone()
+	// reset Proxy to avoid using environment proxy settings
+	transport.Proxy = nil
+
+	// Custom dialer that resolves the SNI hostname to the fixed API server IP
+	transport.DialContext = func(ctx context.Context, network, addr string) (net.Conn, error) {
+		// Extract port from addr (format is "host:port")
+		_, port, err := net.SplitHostPort(addr)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse address %s: %w", addr, err)
+		}
+
+		// Always connect to the configured API server IP
+		fixedAddr := net.JoinHostPort(apiServerIP, port)
+		klog.V(5).Infof("Identity bindings: resolving %s to %s", addr, fixedAddr)
+
+		dialer := &net.Dialer{
+			Timeout:   30 * time.Second,
+			KeepAlive: 30 * time.Second,
+		}
+		return dialer.DialContext(ctx, network, fixedAddr)
+	}
+
+	if transport.TLSClientConfig == nil {
+		transport.TLSClientConfig = &tls.Config{
+			MinVersion: tls.VersionTLS12, // #nosec G402
+		}
+	}
+	transport.TLSClientConfig.ServerName = sniName
+	// Explicitly set minimum TLS version to TLS 1.2 for security
+	transport.TLSClientConfig.MinVersion = tls.VersionTLS12
+
+	// Set custom CA pool if provided
+	if caPool != nil {
+		transport.TLSClientConfig.RootCAs = caPool
+	}
+
+	return transport
+}
+
+// getTransport provides the transport to use for the request
+func (c *identityBindingsTokenCredential) getTransport() (*http.Transport, error) {
+	// Return existing transport if already created
+	if c.transport != nil {
+		return c.transport, nil
+	}
+
+	// Read CA file
+	b, err := os.ReadFile(KubernetesCACertPath)
+	if err != nil {
+		return nil, fmt.Errorf("read CA file %q: %w", KubernetesCACertPath, err)
+	}
+	if len(b) == 0 {
+		return nil, fmt.Errorf("CA file %q is empty", KubernetesCACertPath)
+	}
+
+	// Create CA pool
+	caPool := x509.NewCertPool()
+	if !caPool.AppendCertsFromPEM(b) {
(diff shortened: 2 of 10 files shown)


cmd/acr-credential-provider/main.go
```diff
@@ -68,6 +75,8 @@ func main() {
 	// Flags
 	command.Flags().StringVarP(&RegistryMirrorStr, "registry-mirror", "r", "",
 		"Mirror a source registry host to a target registry host, and image pull credential will be requested to the target registry host when the image is from source registry host")
+	command.Flags().StringToStringVar(&KSAAuthConfigMap, "ksa-auth-config", map[string]string{},
+		"KSA auth configuration as key=value pairs (sni-name=xxx,default-client-id=xxx,default-tenant-id=xxx,api-server-ip=xxx)")
```
norshtein: I think this is not very “standard" and structural? Generally we should have 1 parameter for one configuration? Otherwise you can use one parameter to pass everything.

qweeah: This is a well-established approach for popular CLI tools when applying grouped configurations or options, such as:
```bash
docker buildx create \
  --driver-opt network=host \
  --driver-opt image=moby/buildkit:latest
```
or
```
docker run --mount type=bind,source=/src,target=/dst,readonly
```

qweeah: To simplify input, we might be able to do something like `--ksa-auth sni=xxx,ip=xxx,default-client=xxx`.

clnv: 1. For Docker, I think they do that because the CLI is just a proxy to the Docker daemon, and this approach preserves compatibility.
2. Are they optional?
3. Separate flags could provide more helpful descriptions for each one, as well as clearer type restrictions.

pkg/credentialprovider/identity_bindings_credentials.go
```diff
@@ -0,0 +1,246 @@
...
+
+const (
+	// Kubernetes certificate paths
+	KubernetesCACertPath = "/etc/kubernetes/certs/ca.crt"
+)
+
+// identityBindingsTokenCredential implements azcore.TokenCredential interface
+// using identity bindings token exchange
+type identityBindingsTokenCredential struct {
+	req           *v1.CredentialProviderRequest
```
clnv: Could you extract just the required fields instead of including the entire request here?

fyi, we have a PR to add IB mode to the SDK. It would be easier to migrate if you keep the Azure credential a bit more independent from the credential provider. https://github.com/Azure/azure-sdk-for-go/pull/25664

example use: https://github.com/Azure/secrets-store-csi-driver-provider-azure/compare/master...clnv:secrets-store-csi-driver-provider-azure:ib#diff-0ccbf9a2a7d709b87dbf2f30cc730528c6299d537bf721300d49502cfe521c0aR199-R220

cmd/acr-credential-provider/main.go
qweeah: 1. Not exactly since CLI can make below change to separate the flags but still proxy same data to daemon, right?
```bash
docker buildx create \
  --driver-opt-network host \
  --driver-opt-image moby/buildkit:latest
```

2. Yes, they are optional.
3. I agree on the better description. But we cannot enforce type restriction even if the flags are separated with each other.

Based on the fact that current flag number is small, I can apply your suggestion and split ksa and ib related input to different flags.

pkg/credentialprovider/identity_bindings_credentials.go
norshtein: just to confirm, from the example code, tenant ID is required when requesting access token from IB? If that's the case, we should also have a default tenant ID as a parameter @qweeah

norshtein: I noticed Billy's change in this file does not pass `tenantID` to IB proxy, so my assumption is IB proxy does not need tenant ID to work, but the `azidentity` library requires `tenantID`?

clnv: yes. though the underlying impl might not use it directly, it’s still required.

norshtein: I see. Then let's also add a parameter `--ib-default-tenant-id` and always pass this argument since day 1. This will make our transition much smoother if some day the acr credential provider switches the use official SDK.

clnv (approved):

cmd/acr-credential-provider/main.go
```diff
@@ -68,6 +78,14 @@ func main() {
 	// Flags
 	command.Flags().StringVarP(&RegistryMirrorStr, "registry-mirror", "r", "",
 		"Mirror a source registry host to a target registry host, and image pull credential will be requested to the target registry host when the image is from source registry host")
+	command.Flags().StringVar(&IBSNIName, config.FlagIBSNIName, "",
+		"SNI name for identity bindings")
+	command.Flags().StringVar(&IBDefaultClient, config.FlagIBDefaultClient, "",
+		"Default Azure AD client ID for identity bindings")
```
mainred, line 86: By saying default, we can configure a non-default one or overwrite this by some other setting?

pkg/credentialprovider/identity_bindings_credentials.go
```diff
@@ -0,0 +1,271 @@
...
+)
+
+const (
+	// Kubernetes certificate paths
+	KubernetesCACertPath = "/etc/kubernetes/certs/ca.crt"
+)
+
+// identityBindingsTokenCredential implements azcore.TokenCredential interface
+// using identity bindings token exchange
+type identityBindingsTokenCredential struct {
```
mainred, line 48: Is there any existing official identityBindingsTokenCredential we import instead?

mainred: Which allows us only to provide the GetToken function.

pkg/credentialprovider/identity_bindings_credentials.go
```diff
@@ -0,0 +1,271 @@
...
+
+	// Parse response
+	var tokenResp tokenResponse
+	if err := json.Unmarshal(body, &tokenResp); err != nil {
+		return azcore.AccessToken{}, fmt.Errorf("failed to parse token response: %w", err)
+	}
+
+	expiresOn := time.Now().Add(time.Duration(tokenResp.ExpiresIn) * time.Second)
+
+	klog.V(4).Infof("Successfully obtained token from identity bindings, expires at: %s", expiresOn)
```
mainred, line 207: credential provider leverages defaultCacheDuration to cache the token, should we match ExpiresIn with defaultCacheDuration?

cmd/acr-credential-provider/main.go
```diff
@@ -31,13 +31,18 @@ import (
...
 
+	"sigs.k8s.io/cloud-provider-azure/cmd/acr-credential-provider/pkg/config"
 	"sigs.k8s.io/cloud-provider-azure/pkg/version"
 )
 
 func main() {
 	rand.Seed(time.Now().UnixNano())
 
 	var RegistryMirrorStr string
+	var IBSNIName string
```
nilo19: please wrap them into one var block

nilo19:
@qweeah please check UT, linting, etc locally.

cmd/acr-credential-provider/main.go
qweeah: Yes, this is designed to help with the migration from kubelet (node-level) identity to KSA token auth.

If a service account is created without annotations, then images for workloads using that service account will still be pulled using the kubelet identity. In this case, existing MI users just need to ensure that every workload uses a service account. If this is already set up, there’s no need to update the deployment spec to use this new feature.

pkg/credentialprovider/identity_bindings_credentials.go
qweeah: It requires Identity Bindings SDK, which is still WIP. In the long term, it should be used here.

pkg/credentialprovider/identity_bindings_credentials.go
qweeah: Can you help explain more on 'match'? Do you mean reset the CredentialProviderResponse.CacheDuration to an ideal value, e.g. plus `ExpiresIn / 2`?

pkg/credentialprovider/identity_bindings_credentials.go
```diff
@@ -0,0 +1,271 @@
...
+	caPool := x509.NewCertPool()
+	if !caPool.AppendCertsFromPEM(b) {
+		return nil, fmt.Errorf("parse CA file %q: no valid certificates found", KubernetesCACertPath)
+	}
+
+	// Create and cache transport
+	c.transport = createTransport(c.ibConfig.SNIName, c.ibConfig.APIServerIP, caPool)
+
+	return c.transport, nil
+}
```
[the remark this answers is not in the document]
qweeah: No such use case. Will migrate to dedicated SDK.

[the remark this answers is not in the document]
qweeah: Obsolete description, removed

pkg/credentialprovider/identity_bindings_credentials_test.go
```diff
@@ -0,0 +1,483 @@
...
+	// Test CA file reading (would need to mock os.ReadFile for proper testing)
+	// For now, just verify the transport is created and cached
+	t.Run("transport caching", func(t *testing.T) {
+		if cred.transport != nil {
+			t.Error("transport should be nil initially")
+		}
+		// Note: getTransport() will fail in test environment without proper CA file
+		// This test verifies the structure and caching logic
+	})
+}
```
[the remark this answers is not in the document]
qweeah: No need to test, removed.

pkg/credentialprovider/identity_bindings_credentials.go
```diff
@@ -0,0 +1,271 @@
...
+	defer resp.Body.Close()
+
+	// Read response
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return azcore.AccessToken{}, fmt.Errorf("failed to read response body: %w", err)
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		return azcore.AccessToken{}, fmt.Errorf("token request failed with status %d: %s", resp.StatusCode, string(body))
```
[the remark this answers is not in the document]
qweeah: No need. Just log status

pkg/credentialprovider/identity_bindings_credentials.go
mainred: var expiresAt time.Time
    // nil cache duration means use the default cache duration
    if response.CacheDuration == nil {
        if p.defaultCacheDuration == 0 {
            return dockerConfig, serviceAccountCoordinates, nil
        }
        expiresAt = p.clock.Now().Add(p.defaultCacheDuration)
    } else {
        expiresAt = p.clock.Now().Add(response.CacheDuration.Duration)
    }

mainred: kubelet will use the expiration returned from the token request instead of the default one

qweeah: Confirmed such change is not required for KSA ib support. Currently kubelet will use `defaultCacheDuration` if `CredentialProviderResponse.CacheDuration` is not provided. `CredentialProviderResponse.CacheDuration` by default is set to 5 minute. This logic is applied to all credentials, including service account token and managed identity token

pkg/credentialprovider/identity_bindings_credentials.go
```diff
@@ -0,0 +1,275 @@
...
+	ExpiresIn   int64  `json:"expires_in"`
+}
+
+// createTransport creates an HTTP transport with custom CA
+// The transport uses a custom dialer that resolves the SNI name to the configured API server IP
+func createTransport(sniName string, apiServerIP string, caPool *x509.CertPool) *http.Transport {
+	var transport *http.Transport
+	if tr, ok := http.DefaultTransport.(*http.Transport); ok {
+		transport = tr.Clone()
+	} else {
```
nilo19: how is it possible?

pkg/credentialprovider/identity_bindings_credentials.go
```diff
@@ -0,0 +1,275 @@
...
+		transport = &http.Transport{
+			ForceAttemptHTTP2:   true,
+			MaxIdleConns:        100,
+			IdleConnTimeout:     90 * time.Second,
+			TLSHandshakeTimeout: 10 * time.Second,
+		}
+	}
+
+	// Custom dialer that resolves the SNI hostname to the fixed API server IP
+	transport.DialContext = func(ctx context.Context, network, addr string) (net.Conn, error) {
```
nilo19, line 73: the transport is cloned from defaulttransport, so it hornors `HTTPS_PROXY/HTTP_PROXY`. On a node with this env set, the traffic would go to apiserverIP:proxy_port? which would fail?

qweeah: Good catch. I think if we have `--ib-apiserver-ip` set then the related credential should be anyway sent to the provided IP. Will reset the proxy then.

nilo19:
Do we have unit tests to cover http proxy case?

qweeah:
> Do we have unit tests to cover http proxy case?

Added nil proxy check in the unit test @nilo19

PR #9704, opened 2025-12-01
fix: Revert Azure Stack API Version Bump

patrickdillon:
Reverts #9308

This reverts commit 7427a9c3a6281b32b7437e53c300931a2b96d15d.

2024-07-01 is not a stable version supported by Azure Stack Hub. 2018-11-01 is the latest supported version.

The pull request mentions support for privateIPAddressVersion. I do not think we are likely to see support for such a field in Azure Stack, as Azure Stack does not support private IP addresses. Instead, we should consider how to default or bypass these values for Azure Stack Hub.

#### What type of PR is this?

/kind bug

After #9308 our deployments fail with:

```
level=error msg=    "message": "No registered resource provider found for location 'mtcazs' and API version '2024-07-01' for type 'loadBalancers'. The supported api-versions are '2014-12-01-preview, 2015-05-01-preview, 2015-06-15, 2016-03-30, 2016-06-01, 2016-07-01, 2016-08-01, 2016-09-01, 2016-10-01, 2016-11-01, 2016-12-01, 2017-03-01, 2017-04-01, 2017-06-01, 2017-08-01, 2017-09-01, 2017-10-01, 2017-11-01, 2018-01-01, 2018-02-01, 2018-03-01, 2018-04-01, 2018-05-01, 2018-06-01, 2018-07-01, 2018-08-01, 2018-10-01, 2018-11-01, 2018-12-01, 2019-02-01, 2019-04-01, 2019-06-01, 2019-07-01, 2019-08-01, 2019-09-01, 2019-11-01, 2019-12-01, 2020-01-01, 2020-03-01, 2020-04-01, 2020-05-01, 2020-06-01, 2020-07-01, 2020-08-01, 2020-11-01, 2021-01-01'. The supported locations are 'mtcazs'."
```

Fixes #

```release-note
fix: Revert Azure Stack API Version Bump
```

```docs
fix: Revert Azure Stack API Version Bump
```

### DIFF
--- a/pkg/azclient/loadbalancerclient/interface.go
+++ b/pkg/azclient/loadbalancerclient/interface.go
@@ -25,7 +25,7 @@ import (
 	"sigs.k8s.io/cloud-provider-azure/pkg/azclient/utils"
 )
 
-// +azure:client:verbs=get;createorupdate;delete;list,resource=LoadBalancer,packageName=github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/network/armnetwork/v6,packageAlias=armnetwork,clientName=LoadBalancersClient,expand=true,rateLimitKey=loadBalancerRateLimit,etag=true,azureStackCloudAPIVersion="2024-07-01"
+// +azure:client:verbs=get;createorupdate;delete;list,resource=LoadBalancer,packageName=github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/network/armnetwork/v6,packageAlias=armnetwork,clientName=LoadBalancersClient,expand=true,rateLimitKey=loadBalancerRateLimit,etag=true,azureStackCloudAPIVersion="2018-11-01"
 type Interface interface {
 	utils.GetWithExpandFunc[armnetwork.LoadBalancer]
 	utils.CreateOrUpdateFunc[armnetwork.LoadBalancer]
--- a/pkg/azclient/loadbalancerclient/zz_generated_client.go
+++ b/pkg/azclient/loadbalancerclient/zz_generated_client.go
@@ -31,7 +31,7 @@ import (
 	"sigs.k8s.io/cloud-provider-azure/pkg/azclient/utils"
 )
 
-const AzureStackCloudAPIVersion = "2024-07-01"
+const AzureStackCloudAPIVersion = "2018-11-01"
 
 type Client struct {
 	*armnetwork.LoadBalancersClient


andyzhangx (commented):
/ok-to-test

damdo:
/release-note bug: Revert Azure Stack API Version Bump

damdo:
@andyzhangx will you be able to approve this revert or do we need to tag other Azure folks? Thanks!

andyzhangx:
@tyler-lloyd are you aware of this issue on Azure stack?

andyzhangx:
per https://github.com/kubernetes-sigs/cloud-provider-azure/pull/9308, I think we need to use `2021-01-01` for azure stack, or at least use [2019-04-01](https://github.com/Azure/azure-rest-api-specs/blob/dfb1837bab3c6940fb0a0e745ee1db9bdbd440b3/specification/network/resource-manager/Microsoft.Network/stable/2019-04-01/loadBalancer.json#L1165C10-L1165C33)

patrickdillon:
> per #9308, I think we need to use `2021-01-01` for azure stack, or at least use [2019-04-01](https://github.com/Azure/azure-rest-api-specs/blob/dfb1837bab3c6940fb0a0e745ee1db9bdbd440b3/specification/network/resource-manager/Microsoft.Network/stable/2019-04-01/loadBalancer.json#L1165C10-L1165C33)

I would like to learn more about the problem being addressed in #9308, but I doubt we are going to be able to resolve it by bumping the API version. Azure Stack only has IP addresses; it does not distinguish private and public, so I doubt there will be support for a `privateIPAddressVersion` field. I suspect we will need to fix whatever the issue is similar to how we handled #8755, which is to gracefully handle when `privateIPAddressVersion` is not populated. 

My understanding is that `2018-11-01` is the latest stable Azure-Stack-compatible network API version. The way I arrived at that conclusion is: The [docs here](https://learn.microsoft.com/en-us/azure-stack/user/azure-stack-version-profiles?view=azs-2501#azure-api-profiles-and-azure-stack-hub-compatibility) state you should use a profile. You can determine supported API versions by inspecting a profile for a particular release. So in this case, the latest azure-sdk-for-go release is v68.0.0, the latest profile is 2020-09-01, checking the network package, shows [the compatible API is 2018-11-01](https://github.com/Azure/azure-sdk-for-go/blob/v68.0.0/profiles/2020-09-01/network/mgmt/network/models.go#L15).

tyler-lloyd:
I think there was some confusion around how AzureStackCloud works in AKS. We continue to use the AZURESTACKCLOUD [hack](https://github.com/Azure/go-autorest/blob/be633f7b83b41fc9bd41f4e99aa8664a9fa070fb/autorest/azure/environments.go#L296) to override ARM endpoints but we're not actually ever running in/with Azure Stack Hub. (note we've mostly gotten off track1 clients but the debt of having a dependency on azurestack env config is still there).

The reason I updated the API version was to fix an issue in our air-gapped clouds because it was using the AzureStackCloud env config and letting the API version be overridden ([example](https://github.com/kubernetes-sigs/cloud-provider-azure/blob/b452751132f923ab58af6b7d4175cfa5a5aacc69/pkg/azclient/factory_gen.go#L770)).

After looking at this I see we're just not setting the `DisableAzureStackCloud` in the ARM config which will prevent us using the older API version hard coded for azure stack. That PR is here: https://github.com/Azure/kube-egress-gateway/pull/1272 . Since we're not going to use the azure stack API version then obviously it's fine to revert my original change. Sorry for bug!

tl;dr - I'm fine with the revert - it was made without full context around Azure Stack and we have more correct ways to fix the original issue.

damdo:
Thanks @tyler-lloyd and, just to confirm, you were the original author of the PR being reverted right?

tyler-lloyd:
Yes

tyler-lloyd (approved):

andyzhangx (approved):
/lgtm

PR #9798, opened 2026-01-03
Skip VMSS VM update when no matching disks found during detach operation

landreasyan:
## What type of PR is this?
/kind improvement

## What this PR does / why we need it:
This PR optimizes the VMSS DetachDisk operation by skipping unnecessary VM updates when no matching disks are found to detach. This can occur in two scenarios:
1. The VM has no data disks attached
2. The requested disk(s) to detach are not present on the VM

Previously, the code would proceed with a VM update even when no disks needed to be detached, which could lead to:
- Unnecessary API calls to Azure
- VM entering a pending state during critical failures (e.g., zone outages)
- Blocking clients from seeing successful disk operations
- Delayed recovery during outage scenarios

## Changes made:
- **azure_controller_vmss.go**: Early return in `DetachDisk()` when no matching disks are found, avoiding the VM update call
- **azure_controller_vmss_test.go**: Added comprehensive test cases covering:
  - VM with no disks attached
  - Single non-matching disk to detach
  - Multiple non-matching disks to detach
  - Verification that VM update is skipped in these scenarios

Fixes #

```release-note
none
```

```docs
none
```

### DIFF
--- a/pkg/provider/azure_controller_vmss.go
+++ b/pkg/provider/azure_controller_vmss.go
@@ -172,19 +172,26 @@ func (ss *ScaleSet) DetachDisk(ctx context.Context, nodeName types.NodeName, dis
 	}
 
 	if !bFoundDisk {
-		// only log here, next action is to update VM status with original meta data
-		klog.Warningf("detach azure disk on node(%s): disk list(%s) not found", nodeName, diskMap)
-	} else {
-		if ss.IsStackCloud() {
-			// Azure stack does not support ToBeDetached flag, use original way to detach disk
-			var newDisks []*armcompute.DataDisk
-			for _, disk := range disks {
-				if !ptr.Deref(disk.ToBeDetached, false) {
-					newDisks = append(newDisks, disk)
-				}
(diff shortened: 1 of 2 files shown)


pkg/provider/azure_controller_vmss.go
```diff
@@ -172,19 +172,26 @@ func (ss *ScaleSet) DetachDisk(ctx context.Context, nodeName types.NodeName, dis
 	}
 
 	if !bFoundDisk {
-		// only log here, next action is to update VM status with original meta data
-		klog.Warningf("detach azure disk on node(%s): disk list(%s) not found", nodeName, diskMap)
```
andyzhangx, line 176: while the original intension for this update VM operation is to get rid of VM failure state, it seems there is conflict with this zone outage issue.

landreasyan: Yes, exactly and since recovering the VM state is not necessarily the goal of the Detach method, I believe making it more zonally resilient makes sense.

PR #9841, opened 2026-01-16
fix: network isolated clusters should always use managed identity cre…

norshtein:
…dential

#### What type of PR is this?
/kind bug 

This PR fixes a bug that network isolated cluster try to use identity binding credential. It should always use managed identity credential.

Fixes #

```release-note
fix: network isolated clusters should always use managed identity credential
```

```docs

```

### DIFF
--- a/pkg/credentialprovider/azure_credentials_test.go
+++ b/pkg/credentialprovider/azure_credentials_test.go
@@ -387,491 +387,226 @@ func TestProcessMirrorMapping(t *testing.T) {
 	}
 }
 
-func TestNewAcrProvider_WithEmptyServiceAccountToken(t *testing.T) {
-	// Test case when ServiceAccountToken is empty - should use managed identity credential
+// createTempConfigFile creates a temporary config file with the given content and
+// registers cleanup to remove it when the test completes.
+func createTempConfigFile(t *testing.T, content string) string {
+	t.Helper()
 	configFile, err := os.CreateTemp(".", "config.json")
 	assert.NoError(t, err)
-	defer os.Remove(configFile.Name())
+	t.Cleanup(func() { os.Remove(configFile.Name()) })
 
-	configStr := `{
+	_, err = configFile.WriteString(content)
+	assert.NoError(t, err)
+	assert.NoError(t, configFile.Close())
+	return configFile.Name()
+}
+
+const (
+	testTenantIDConfig = `{
+		"tenantID": "test-tenant-id"
+	}`
+	testManagedIdentityConfig = `{
 		"useManagedIdentityExtension": true,
 		"userAssignedIdentityID": "test-client-id"
 	}`
-	_, err = configFile.WriteString(configStr)
-	assert.NoError(t, err)
-	assert.NoError(t, configFile.Close())
+)
 
-	req := &v1.CredentialProviderRequest{
-		Image:               "test.azurecr.io/test:latest",
-		ServiceAccountToken: "", // Empty token
-	}
-
-	provider, err := NewAcrProvider(req, "", configFile.Name(), IdentityBindingsConfig{})
-	assert.NoError(t, err)
-	assert.NotNil(t, provider)
-
-	acrProv := provider.(*acrProvider)
-	assert.NotNil(t, acrProv.credential)
-	assert.Equal(t, true, acrProv.config.UseManagedIdentityExtension)
-}
-
-func TestNewAcrProvider_WithServiceAccountToken(t *testing.T) {
-	// Test case when ServiceAccountToken is provided - should use service account credential
-	configFile, err := os.CreateTemp(".", "config.json")
-	assert.NoError(t, err)
-	defer os.Remove(configFile.Name())
-
-	configStr := `{
-		"tenantID": "test-tenant-id"
-	}`
-	_, err = configFile.WriteString(configStr)
-	assert.NoError(t, err)
-	assert.NoError(t, configFile.Close())
-
-	req := &v1.CredentialProviderRequest{
-		Image:               "test.azurecr.io/test:latest",
-		ServiceAccountToken: "test-service-account-token",
-		ServiceAccountAnnotations: map[string]string{
-			"kubernetes.azure.com/acr-client-id": "test-client-id",
-			"kubernetes.azure.com/acr-tenant-id": "test-tenant-id",
+func TestNewAcrProvider_CredentialSelection(t *testing.T) {
+	tests := []struct {
+		name                      string
+		configStr                 string
+		image                     string
+		serviceAccountToken       string
+		serviceAccountAnnotations map[string]string
+		registryMirror            string
+		ibConfig                  IdentityBindingsConfig
+		expectError               bool
+		errorContains             string
+		expectManagedIdentity     bool
+	}{
+		{
+			name:                  "empty service account token should use managed identity",
+			configStr:             testManagedIdentityConfig,
+			image:                 "test.azurecr.io/test:latest",
+			expectManagedIdentity: true,
+		},
+		{
+			name:                "service account token should use client assertion credential",
+			configStr:           testTenantIDConfig,
+			image:               "test.azurecr.io/test:latest",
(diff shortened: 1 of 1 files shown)


pkg/credentialprovider/azure_credentials.go
```diff
@@ -84,14 +84,26 @@ func NewAcrProvider(req *v1.CredentialProviderRequest, registryMirrorStr string,
...
+		config:         config,
+		environment:    env,
+		cloudConfig:    clientOption.Cloud,
+		registryMirror: parseRegistryMirror(registryMirrorStr),
+	}
+
+	targetloginServer, _ := provider.parseACRLoginServerFromImage(req.Image)
+	isACR := (targetloginServer != "")
+	isNICluster := (len(registryMirrorStr) > 0)
+	shouldUseManagedIdentity := (isACR && isNICluster)
```
qweeah: What if the registryMirrorStr is set from mcr.microsoft.com to xxx.azurecr.io and the user is requesting xxx.azurecr.io/some:image? In this case should use ibip if `SNIName` is set, right?

qweeah: Managed identity should **ONLY** be forced when the image's registry host matches a source registry in the mirror configuration (i.e., when the host name of the image reference appears as a key of `registryMirror`). Unit test also needed for it.

qweeah: Looks like it can be done via checking the `sourceRegistry` returned by `parseACRLoginServerFromImage`. If it is not an empty string, then the host name of the image reference does appear as a key of `registryMirror`

norshtein: Synced offline. Updated code based on the comment.

pkg/credentialprovider/azure_credentials.go
```diff
@@ -84,14 +85,26 @@ func NewAcrProvider(req *v1.CredentialProviderRequest, registryMirrorStr string,
...
+		config:         config,
+		environment:    env,
+		cloudConfig:    clientOption.Cloud,
+		registryMirror: parseRegistryMirror(registryMirrorStr),
+	}
+
+	targetloginServer, sourceRegistry := provider.parseACRLoginServerFromImage(req.Image)
+	mcrMirroredToACR := (targetloginServer != "" && strings.EqualFold(sourceRegistry, MCRHost))
+	isNICluster := (len(registryMirrorStr) > 0)
+	shouldUseManagedIdentity := (mcrMirroredToACR && isNICluster)
```
qweeah: We can remove MAR domain name here simply via flagging the target and source of a registry mirror:
```suggestion
	targetloginServer, sourceRegistry := provider.parseACRLoginServerFromImage(req.Image)
	isMirrorTarget := targetloginServer != ""
	_, isMirrorSource := provider.registryMirror[sourceRegistry]
	mirroredViaNI := (isMirrorTarget && sourceRegistry != "" && isMirrorSource)
```

pkg/credentialprovider/azure_credentials_test.go
```diff
@@ -579,3 +579,259 @@ func TestNewAcrProvider_InvalidConfig(t *testing.T) {
 	assert.Nil(t, provider)
 	assert.Contains(t, err.Error(), "failed to load config")
 }
+
```
qweeah: Need one more test to cover
1) Registry Mirror is set, e.g. `mcr.microsoft.com:test.azurecr.io`
2) Image uses mirror target directly, e.g. `test.azurecr.io/azure-cli:latest`
3) ibConfig.SNIName is set

Test should use identity bindings

pkg/credentialprovider/azure_credentials.go
qweeah: Actually `mirroredViaNI` should be `mirrored`, no need to mention NI

pkg/credentialprovider/azure_credentials_test.go
norshtein: makes great sense, updated test.

pkg/credentialprovider/azure_credentials.go
norshtein: discussed offline, simplified this part logic according to suggestion.

nilo19:
added release note

pkg/credentialprovider/azure_credentials_test.go
nilo19: could you please suppress the alarm?

pkg/credentialprovider/azure_credentials.go
```diff
@@ -84,14 +84,26 @@ func NewAcrProvider(req *v1.CredentialProviderRequest, registryMirrorStr string,
...
+		config:         config,
+		environment:    env,
+		cloudConfig:    clientOption.Cloud,
+		registryMirror: parseRegistryMirror(registryMirrorStr),
+	}
+
+	_, sourceRegistry := provider.parseACRLoginServerFromImage(req.Image)
+	// Assumption here is mirrored image is only used on NI cluster and for mirrored image,
+	// managed identity authentication is always used.
+	mirrored := sourceRegistry != ""
```
nilo19, line 97: for self-hosted clusters will they configure to use mirrored image in non-NI clusters in the future?

pkg/credentialprovider/azure_credentials_test.go
```diff
@@ -579,3 +579,299 @@ func TestNewAcrProvider_InvalidConfig(t *testing.T) {
...
+func TestNewAcrProvider_MCRImageMirroredToACR_UsesManagedIdentity(t *testing.T) {
```
nilo19: it would be better to use table driven tests because I can see there are duplicates across tests.

norshtein: updated to use table driven tests.

pkg/credentialprovider/azure_credentials_test.go
norshtein: fixed the lint alert.

pkg/credentialprovider/azure_credentials.go
norshtein: I have little context about self-hosted clusters. I think we can update the code if this assumption changes in future..?

nilo19: The assumption has no problem at all, and is restricted to AKS. I'm just asking for non-aks clusters where the cluster admin has full control over the cluster, will they be hitting any issue if they want to use non-MI authentication here?

norshtein: Got your point. I think "image mirror" is only useful for AKS NI scenario - cx need to pull image from MCR(for AKS managed addons), but due to the network is isolated, they cannot connect to MCR and has to use ACR as a mirror. For non-AKS clusters, I don't think cx has motivation to specify a image registry A as container image but they actually want to pull from image registry B - why don't they just specify registry B as the container image?

nilo19: do we want to block this use by emitting a clear error?

norshtein: do we have a way to distinguish the usage of acr-credential-provider in AKS and non-AKS? my understanding is inside the binary, it does not know whether it is invoked in AKS or non-AKS(unless we add an additional parameter specifying the running environment)

norshtein: Another perspective: there're 3 authentication methods in acr-credential-provider right now:
1. Identity binding - which is Billy added and is AKS-only
2. Service account token - Qingchuan added that and requires OIDC issuer to be enabled, also scoped to AKS.
3. Managed identity - the original authentication method, most widely used.

We can see for non-AKS clusters, both 1 and 2 are not available, they can only use managed identity authentication method. So I think this PR should be safe enough:
1. For AKS scenarios, we know NI is the only scenario which will use the image mirror, we checked that and let it fallback to managed identity.
2. For non-AKS scenarios, since managed identity is the only one authentication method cx can use, and this PR is basically making a scenario which originally does not use managed identity to change to use managed identity, it will only make non-AKS scenarios safer.

PR #9937, opened 2026-02-09
fix(multi-slb): support IP sharing across multiple services

### LINKED ISSUE #9867: [Multi-SLB] support ip sharing across multiple services
#### What happened:

Service A has been created to use IP 1. Now we want to create a service B point to the same IP. During the lb selection phase of service B, it will select another LB with less existing services than service A's LB. Since the LB selected is a different one, LB migration logic will be triggered, and the service A's frontend IP config will be removed by accident.
https://github.com/kubernetes-sigs/cloud-provider-azure/blob/master/pkg/provider/azure_loadbalancer.go#L846

#### Environment:

- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

Liunardy:
#### What type of PR is this?
/kind bug

In multi-SLB mode, when Service B shares an IP with existing Service A, `getAzureLoadBalancerName` picks the LB with the fewest rules instead of the LB where the IP resides. This triggers migration logic that accidentally removes Service A's frontend IP configuration. This fixes that by picking the LB where the IP resides.

Additional fixes:
- Block LB migration when it would remove a frontend IP shared with other services.
- Block conflicting LB config annotation when the service also specifies an IP.
- Block IP specification pointing to an ineligible LB.

Fixes #9867

```release-note
fix(multi-slb): support IP sharing across multiple services

When a service specifies an IP address that already exists on a load balancer, the service is now placed on that load balancer instead of picking one with the fewest rules, provided the service is eligible for that load balancer. The load balancer configuration annotation cannot be combined with an IP specification. Migration to a different load balancer is blocked if the frontend IP is still referenced by other resources.

Switching internal/external issues (10050 and 10117) will be fixed in another change.
```

```docs

```

### DIFF
--- a/pkg/provider/azure_loadbalancer.go
+++ b/pkg/provider/azure_loadbalancer.go
@@ -2541,8 +2541,11 @@ func (az *Cloud) recordExistingNodesOnLoadBalancers(clusterName string, lbs []*a
 	return nil
 }
 
-// lbHasServiceOwnedResources returns true if any load-balancing rule or probe
-// on the given LB is owned by the service (matched by name prefix).
+// lbHasServiceOwnedResources returns true if any load-balancing rule or
+// probe on the given LB is owned by the service (matched by name prefix).
+// The shared health probe is skipped: if the service uses it, there must
+// also be a service-owned rule referencing it, which the rule check
+// already catches.
 func (az *Cloud) lbHasServiceOwnedResources(lb *armnetwork.LoadBalancer, service *v1.Service) bool {
 	if lb.Properties == nil {
 		return false
@@ -2553,7 +2556,9 @@ func (az *Cloud) lbHasServiceOwnedResources(lb *armnetwork.LoadBalancer, service
 		}
 	}
 	for _, probe := range lb.Properties.Probes {
-		if probe.Name != nil && az.serviceOwnsRule(service, *probe.Name) {
+		if probe.Name != nil &&
+			!strings.EqualFold(*probe.Name, consts.SharedProbeName) &&
+			az.serviceOwnsRule(service, *probe.Name) {
 			return true
 		}
 	}
--- a/pkg/provider/azure_loadbalancer.go
+++ b/pkg/provider/azure_loadbalancer.go
@@ -846,11 +846,8 @@ func (az *Cloud) getServiceLoadBalancer(
 
 				deletedLBName, deletedPLS, err := az.removeStaleServiceLBResources(ctx, existingLB, existingLBs, clusterName, service)
 				if err != nil {
-					logger.Error(err, "failed to remove stale service resources from old load balancer",
-						"service", service.Name,
-						"oldLB", ptr.Deref(existingLB.Name, ""),
-					)
-					return nil, nil, nil, nil, false, false, err
+					return nil, nil, nil, nil, false, false, fmt.Errorf("remove stale service resources from load balancer %q for service %q: %w",
+						ptr.Deref(existingLB.Name, ""), getServiceName(service), err)
 				}
 				if deletedPLS {
 					return nil, nil, nil, nil, false, true, nil
@@ -877,13 +874,14 @@ func (az *Cloud) getServiceLoadBalancer(
 					if az.backendPoolUpdater != nil {
 						az.backendPoolUpdater.removeOperation(svcName)
 					}
-					newLBs, err := az.cleanupLocalServiceBackendPool(ctx, service, nodes, existingLBs, clusterName)
-					if err != nil {
-						logger.Error(err, "failed to clean up backend pool for local service",
-							"service", service.Name, "oldLB", ptr.Deref(existingLB.Name, ""))
-						return nil, nil, nil, nil, false, false, err
+					if deletedLBName == "" {
+						newLBs, err := az.cleanupLocalServiceBackendPool(ctx, service, nodes, existingLBs, clusterName)
+						if err != nil {
+							return nil, nil, nil, nil, false, false, fmt.Errorf("clean up backend pool for local service %q on load balancer %q: %w",
+								getServiceName(service), ptr.Deref(existingLB.Name, ""), err)
+						}
+						existingLBs = newLBs
 					}
-					existingLBs = newLBs
 				}
 			}
 			continue
@@ -913,7 +911,7 @@ func (az *Cloud) getServiceLoadBalancer(
 					}
 					if unsafe {
 						return nil, nil, nil, nil, false, false, fmt.Errorf(
-							"service %q cannot migrate from load balancer %q to %q because frontend IP %q is shared with other services; "+
+							"service %q cannot migrate from load balancer %q to %q because frontend IP %q is also referenced by other resources; "+
 								"remove the %s annotation to stay on load balancer %q",
 							service.Name, ptr.Deref(existingLB.Name, ""), defaultLBName, ptr.Deref(fip.Name, ""),
 							consts.ServiceAnnotationLoadBalancerConfigurations, ptr.Deref(existingLB.Name, ""))
@@ -2607,11 +2605,8 @@ func (az *Cloud) removeStaleServiceLBResources(
 			}
 			unsafe, err := az.isFrontendIPConfigUnsafeToDelete(lb, service, fip.ID)
 			if err != nil {
-				logger.Error(err, "failed to check if FIP is safe to delete",
-					"fipID", ptr.Deref(fip.ID, ""),
-					"lbName", ptr.Deref(lb.Name, ""),
-				)
-				continue
+				return "", false, fmt.Errorf("check if frontend IP configuration %q on load balancer %q is safe to delete: %w",
+					ptr.Deref(fip.ID, ""), ptr.Deref(lb.Name, ""), err)
 			}
 			if !unsafe {
 				orphanedFIPs = append(orphanedFIPs, fip)
@@ -2624,21 +2619,13 @@ func (az *Cloud) removeStaleServiceLBResources(
 	if len(orphanedFIPs) > 0 {
 		deletedLBName, deletedPLS, err := az.removeFrontendIPConfigurationFromLoadBalancer(ctx, lb, existingLBs, orphanedFIPs, clusterName, service)
 		if err != nil {
-			logger.Error(err, "failed to remove orphaned frontend IP configurations",
-				"lbName", ptr.Deref(lb.Name, ""),
-				"serviceName", serviceName,
-			)
-			return "", false, err
+			return "", false, fmt.Errorf("remove orphaned frontend IP configurations: %w", err)
 		}
 		return deletedLBName, deletedPLS, nil
 	}
 
 	// 5. No orphaned FIPs, just update the LB with the rule/probe changes.
 	if err := az.CreateOrUpdateLB(ctx, service, *lb); err != nil {
-		logger.Error(err, "failed to update load balancer to remove stale service resources",
-			"lbName", ptr.Deref(lb.Name, ""),
-			"serviceName", serviceName,
-		)
 		return "", false, err
 	}
 	logger.V(2).Info("Updated load balancer to remove stale service resources",
@@ -4514,7 +4501,7 @@ func getMostEligibleLBForService(
 ) string {
 	logger := log.Background().WithName("getMostEligibleLBForService")
 	// 1. If the LB is eligible and being used, choose it.
-	if StringInSlice(currentLBName, eligibleLBs) {
+	if StringInSliceIgnoreCase(currentLBName, eligibleLBs) {
 		logger.V(4).Info("choose LB as it is eligible and being used", "currentLBName", currentLBName)
 		return currentLBName
 	}
@@ -4542,7 +4529,7 @@ func getMostEligibleLBForService(
 	ruleCount := 301
 	for i := range existingLBs {
 		existingLB := existingLBs[i]
-		if StringInSlice(trimSuffixIgnoreCase(ptr.Deref(existingLB.Name, ""), consts.InternalLoadBalancerNameSuffix), eligibleLBs) &&
+		if StringInSliceIgnoreCase(trimSuffixIgnoreCase(ptr.Deref(existingLB.Name, ""), consts.InternalLoadBalancerNameSuffix), eligibleLBs) &&
 			isInternalLoadBalancer(existingLB) == isInternal {
 			if existingLB.Properties != nil &&
 				existingLB.Properties.LoadBalancingRules != nil {
--- a/pkg/provider/azure_loadbalancer.go
+++ b/pkg/provider/azure_loadbalancer.go
@@ -844,13 +844,16 @@ func (az *Cloud) getServiceLoadBalancer(
 				az.shouldChangeLoadBalancer(service, ptr.Deref(existingLB.Name, ""), clusterName, defaultLBName) &&
 				az.lbHasServiceOwnedResources(existingLB, service) {
 
-				deletedLBName, err := az.removeStaleServiceLBResources(ctx, existingLB, existingLBs, clusterName, service)
+				deletedLBName, deletedPLS, err := az.removeStaleServiceLBResources(ctx, existingLB, existingLBs, clusterName, service)
 				if err != nil {
 					logger.Error(err, "failed to remove stale service resources from old load balancer",
 						"service", service.Name,
 						"oldLB", ptr.Deref(existingLB.Name, ""),
 					)
 					return nil, nil, nil, nil, false, false, err
+				}
+				if deletedPLS {
+					return nil, nil, nil, nil, false, true, nil
 				}
 				logger.V(2).Info("Removed stale service resources from old load balancer",
 					"service", service.Name,
@@ -2570,7 +2573,7 @@ func (az *Cloud) removeStaleServiceLBResources(
 	existingLBs []*armnetwork.LoadBalancer,
 	clusterName string,
 	service *v1.Service,
-) (string, error) {
+) (string, bool /* deletedPLS */, error) {
 	logger := log.FromContextOrBackground(ctx).WithName("removeStaleServiceLBResources")
 	serviceName := getServiceName(service)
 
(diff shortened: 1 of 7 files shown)


pkg/provider/azure_loadbalancer.go
```diff
@@ -4107,12 +4164,56 @@ func (az *Cloud) getAzureLoadBalancerName(
...
+		// If service specifies an IP, find which LB has that IP. The service should use the LB where the IP resides.
+		// This also detects secondary services sharing an IP with a primary service already on an LB.
+		if pinsIP {
+			if requiresInternalLoadBalancer(service) {
+				currentLBName = az.getLoadBalancerNameByPrivateIP(ctx, service, existingLBs)
+			} else {
+				currentLBName = az.getLoadBalancerNameByPublicIP(ctx, service, clusterName)
+			}
+			// Block service if the IP resides on an LB that's not eligible.
+			if currentLBName != "" && !StringInSlice(currentLBName, eligibleLBs) {
```
nilo19: please ignore case here as we have ignored case in getLoadBalancerNameByPrivateIP.

pkg/provider/azure_loadbalancer.go
```diff
@@ -807,10 +807,20 @@ func (az *Cloud) getServiceLoadBalancer(
 		existingLBs = lbs
 	}
 
+	// A service must use the LB that owns its IP. In IP-pinned multi-SLB scenario,
+	// The service may currently exist on a different LB needing migration,
+	// so all LBs should be scanned instead of returning on the first match.
+	lbIPs := getServiceLoadBalancerIPs(service)
+	pipNames := getServicePIPNames(service)
+	pinsIP := len(lbIPs) > 0 || lo.ContainsBy(pipNames, func(s string) bool { return s != "" })
```
nilo19: can we move the empty check into getServicePIPNames?

pkg/provider/azure_loadbalancer.go
```diff
@@ -843,6 +853,29 @@ func (az *Cloud) getServiceLoadBalancer(
...
+			// Block migration if the frontend IP is unsafe to delete
+			// (shared with other services, or referenced by outbound/NAT rules).
+			if az.UseMultipleStandardLoadBalancers() {
+				for _, fip := range fipConfigs {
+					unsafe, err := az.isFrontendIPConfigUnsafeToDelete(existingLB, service, fip.ID)
+					if err != nil {
+						return nil, nil, nil, nil, false, false, err
+					}
+					if unsafe {
+						logger.Info("Blocking load balancer migration because frontend IP is shared with other services",
```
nilo19: according to the logging guide, don't emit an error log before returning an error. you may put the useful information into the error message.

pkg/provider/azure_loadbalancer.go
Liunardy: Added `StringInSliceIgnoreCase` function and used it here.

pkg/provider/azure_loadbalancer.go
Liunardy: Moved the check there and updated the unit test. Now it only returns list with non-empty string.

pkg/provider/azure_loadbalancer.go
Liunardy: Removed the log and added additional info (the LB to migrate to) into the error message.

pkg/provider/azure_loadbalancer.go
```diff
@@ -825,7 +835,49 @@ func (az *Cloud) getServiceLoadBalancer(
 			return nil, nil, nil, nil, false, false, err
 		}
 		if status == nil {
-			// service is not on this load balancer
+			// Service is not on this LB by frontend IP ownership.
+			// However, if multi-SLB is enabled and the service should move to a
```
nilo19, line 839: only for multi-slb? what if we remove the ip pin in single-slb mode?

nilo19: ok, I don't think we have this bug for single-slb.

nilo19:
If the service is the only secondary service, and moved to another LB, the stale frontend ip config won't be removed, thus the old pip cannot be removed as well.

pkg/provider/azure_loadbalancer.go
```diff
@@ -825,7 +835,54 @@ func (az *Cloud) getServiceLoadBalancer(
...
+			// different LB, the old LB may still have stale rules/probes left
+			// from when the service shared a frontend IP. Remove them to
+			// prevent orphaned rules and probes on the old LB.
+			if wantLb && az.UseMultipleStandardLoadBalancers() &&
+				az.shouldChangeLoadBalancer(service, ptr.Deref(existingLB.Name, ""), clusterName, defaultLBName) &&
+				az.lbHasServiceOwnedResources(existingLB, service) {
+
+				deletedLBName, err := az.removeStaleServiceLBResources(ctx, existingLB, existingLBs, clusterName, service)
```
nilo19: removeFrontendIPConfigurationFromLoadBalancer returns a deletedPLS bool. When a Private Link Service exists on the FIP, it exits early with deletedPLS=true without actually updating/deleting the LB. The existing migration path at L930 handles this correctly by returning early with deletedPLS=true to trigger a retry. But the new removeStaleServiceLBResources at L2622 ignores this signal entirely — it discards deletedPLS with _, err := and treats success as full completion.

pkg/provider/azure_loadbalancer.go
```diff
@@ -4122,6 +4391,122 @@ func (az *Cloud) getAzureLoadBalancerName(
...
+	loadBalancerIPs := getServiceLoadBalancerIPs(service)
+	pipResourceGroup := az.getPublicIPAddressResourceGroup(service)
+	for _, ip := range loadBalancerIPs {
+		if ip == "" {
+			continue
+		}
+		pip, err := az.findMatchedPIP(ctx, ip, "", pipResourceGroup)
+		if err != nil {
+			logger.V(2).Info("Failed to find PIP by IP", "ip", ip, "error", err)
+			continue
```
nilo19: why do we want to continue instead of returning an error?
this means if there is an error here (e.g., by a transient arm failure), it will sliently falling back to the generic fewest-rules chooser, which means the bug fix fixes nothing.

pkg/provider/azure_loadbalancer.go
nilo19: not related to this fix but need to ignore case here as well, similiar with L4375.

pkg/provider/azure_loadbalancer.go
```diff
@@ -4107,12 +4330,58 @@ func (az *Cloud) getAzureLoadBalancerName(
...
 			return "", err
 		}
 
 		currentLBName := az.getServiceCurrentLoadBalancerName(service)
+
+		// If service specifies an IP, find which LB has that IP. The service should use the LB where the IP resides.
+		// This also detects secondary services sharing an IP with a primary service already on an LB.
+		if pinsIP {
+			if requiresInternalLoadBalancer(service) {
+				currentLBName = az.getLoadBalancerNameByPrivateIP(ctx, service, existingLBs)
```
nilo19, line 4360: we are 100% sure existingLBs cannot be empty, right? or it will fallback to existing lb selection logic which contains bug.

pkg/provider/azure_loadbalancer.go
```diff
@@ -825,7 +835,54 @@ func (az *Cloud) getServiceLoadBalancer(
...
+				}
+
+				az.reconcileMultipleStandardLoadBalancerConfigurationStatus(
+					false,
+					getServiceName(service),
+					ptr.Deref(existingLB.Name, ""),
+				)
+
+				// Same backend pool cleanup as the normal LB migration path below.
+				if isLocalService(service) {
```
nilo19, line 872: You removed deletedLBName != "" check, which seems to be harmless, as in this case cleanupLocalServiceBackendPool will be a no-op, but better have this check to keep aligned with the existing logics below.

pkg/provider/azure_loadbalancer.go
```diff
@@ -825,7 +835,54 @@ func (az *Cloud) getServiceLoadBalancer(
...
+					svcName := getServiceName(service)
+					if az.backendPoolUpdater != nil {
+						az.backendPoolUpdater.removeOperation(svcName)
+					}
+					newLBs, err := az.cleanupLocalServiceBackendPool(ctx, service, nodes, existingLBs, clusterName)
```
nilo19: have we tested this case either manually or by a test case?

pkg/provider/azure_loadbalancer.go
```diff
@@ -825,7 +835,54 @@ func (az *Cloud) getServiceLoadBalancer(
...
+			// different LB, the old LB may still have stale rules/probes left
+			// from when the service shared a frontend IP. Remove them to
+			// prevent orphaned rules and probes on the old LB.
+			if wantLb && az.UseMultipleStandardLoadBalancers() &&
+				az.shouldChangeLoadBalancer(service, ptr.Deref(existingLB.Name, ""), clusterName, defaultLBName) &&
+				az.lbHasServiceOwnedResources(existingLB, service) {
+
+				deletedLBName, err := az.removeStaleServiceLBResources(ctx, existingLB, existingLBs, clusterName, service)
+				if err != nil {
+					logger.Error(err, "failed to remove stale service resources from old load balancer",
```
nilo19: there are many places in this pr where you emit an error log before returning the error, this is not recommended as per logging conventions doc.

pkg/provider/azure_loadbalancer.go
```diff
@@ -843,6 +900,24 @@ func (az *Cloud) getServiceLoadBalancer(
...
+			// (shared with other services, or referenced by outbound/NAT rules).
+			if az.UseMultipleStandardLoadBalancers() {
+				for _, fip := range fipConfigs {
+					unsafe, err := az.isFrontendIPConfigUnsafeToDelete(existingLB, service, fip.ID)
+					if err != nil {
+						return nil, nil, nil, nil, false, false, err
+					}
+					if unsafe {
+						return nil, nil, nil, nil, false, false, fmt.Errorf(
+							"service %q cannot migrate from load balancer %q to %q because frontend IP %q is shared with other services; "+
```
nilo19: minor concern:  isFrontendIPConfigUnsafeToDelete can also report outbound/NAT rule references, making the message misleading.

pkg/provider/azure_loadbalancer.go
Liunardy: getAzureLoadBalancerName receives existingLBs from getServiceLoadBalancer and selectLoadBalancer.

existingLBs is whatever the caller of getServiceLoadBalancer passes:
- reconcileLoadBalancer passes the result of ListManagedLBs. This can return empty only if no LBs exist yet. In empty case, this means no LB has the private IP yet because no LB exists. 
https://github.com/kubernetes-sigs/cloud-provider-azure/blob/0f2687ccdf71e86f6cf07321f5bc3d10f2fd96e1/pkg/provider/azure_loadbalancer.go#L1839

- EnsureLoadBalancerDeleted passes empty existingLBs, but wantLb=false (I'm changing the LB lookup by IP to only be done if wantLB=true, so this skips)
https://github.com/kubernetes-sigs/cloud-provider-azure/blob/0f2687ccdf71e86f6cf07321f5bc3d10f2fd96e1/pkg/provider/azure_loadbalancer.go#L476

- GetLoadBalancer and shouldUpdateLoadBalancer both pass wantLb=false, skipped
https://github.com/kubernetes-sigs/cloud-provider-azure/blob/0f2687ccdf71e86f6cf07321f5bc3d10f2fd96e1/pkg/provider/azure_loadbalancer.go#L106
https://github.com/kubernetes-sigs/cloud-provider-azure/blob/0f2687ccdf71e86f6cf07321f5bc3d10f2fd96e1/pkg/provider/azure_loadbalancer.go#L112
https://github.com/kubernetes-sigs/cloud-provider-azure/blob/0f2687ccdf71e86f6cf07321f5bc3d10f2fd96e1/pkg/provider/azure_loadbalancer.go#L3295

selectLoadBalancer is only called for basic LB clusters. The multi-SLB LB lookup is never reached from this call.
https://github.com/kubernetes-sigs/cloud-provider-azure/blob/0f2687ccdf71e86f6cf07321f5bc3d10f2fd96e1/pkg/provider/azure_loadbalancer.go#L896-L898

pkg/provider/azure_loadbalancer.go
Liunardy: Returned deletedPLS and added a test case for removeStaleServiceLBResources.

pkg/provider/azure_loadbalancer.go
Liunardy: Returned the error and only look up the LB if wantLb.

pkg/provider/azure_loadbalancer.go
Liunardy: Ignored case in this function.

pkg/provider/azure_loadbalancer.go
Liunardy: Added the check.

pkg/provider/azure_loadbalancer.go
Liunardy: Added test case "remove backend pool when a local secondary service changes its load balancer".

pkg/provider/azure_loadbalancer.go
Liunardy: Removed the error log.

pkg/provider/azure_loadbalancer.go
Liunardy: Updated the reason to mention "also referenced by other resources". When user encountered this error, they'll see other events from isFrontendIPConfigUnsafeToDelete as well for example:

Warning DeletingFrontendIPConfiguration 6s azure-cloud-provider isFrontendIPConfigUnsafeToDelete: frontend IP configuration with ID [the id] on LB [the lb] cannot be deleted because it is being referenced by load balancing rules of other services

Warning  SyncLoadBalancerFailed 6s service-controller Error syncing load balancer: failed to ensure load balancer: service "[the svc]" cannot migrate from load balancer "[the lb from]" to "[the lb to]" because frontend IP "[the fip]" is also referenced by other resources; remove the service.beta.kubernetes.io/azure-load-balancer-configurations annotation to stay on load balancer "[the lb from]"

pkg/provider/azure_loadbalancer.go
```diff
@@ -2074,7 +2167,12 @@ func (az *Cloud) reconcileLoadBalancer(ctx context.Context, clusterName string,
 		}
 	}
 
-	if fipChanged {
+	// Update multi-SLB ActiveServices tracking. Two signals indicate a real change:
+	//   - fipChanged: FIP added/removed (service is sole owner of its IP)
+	//   - lbRulesChanged: rules added/removed but FIP kept (shared-IP scenario)
+	// The "flipped" reconcile pass (wantLb=false for the opposite LB type)
+	// triggers neither, so it won't incorrectly remove the service.
+	if fipChanged || (az.UseMultipleStandardLoadBalancers() && lbRulesChanged) {
```
nilo19, line 2175: in an non ip-sharing case where lbRulesChanged=truen fipChanged=false, there will be an extra operation to insert the svc into the lbconfig. but it would be harmless, right?

pkg/provider/azure_loadbalancer_test.go
```diff
@@ -1940,579 +1940,1591 @@ func TestGetServiceLoadBalancerMultiSLB(t *testing.T) {
...
+			expectedError: fmt.Errorf(
+				"service %q cannot migrate from load balancer %q to %q because frontend IP %q is also referenced by other resources; "+
+					"remove the %s annotation to stay on load balancer %q",
+				"primary", "lb1", "lb2", "aprimary",
+				consts.ServiceAnnotationLoadBalancerConfigurations, "lb1"),
 		},
 		{
-			desc: "getServiceLoadBalancer should not change LB when names are the same but with different case with standard LB",
-			SKU:  consts.LoadBalancerSKUStandard,
+			description: "internal: block migration when primary svc shares IP with secondary",
```
nilo19, line 2181: after this pr merges, can we enrich our doc as well to state the multi-slb ip-sharing behavior?

tests/e2e/network/multiple_standard_lb.go
```diff
@@ -311,3 +1867,93 @@ func waitLBCountEqualTo(tc *utils.AzureTestClient, interval, timeout time.Durati
 	})
 	return lbNames, err
 }
+
+// waitForServiceEventAfter waits for an event of the given type on the service with the expected Reason
+// that occurred after the specified time and contains the expected message substring.
+func waitForServiceEventAfter(
```
nilo19: can we move the utils funcs to service_utils.go?

nilo19:
AI generated e2e overlap analysis:

image

pkg/provider/azure_loadbalancer.go
Liunardy: Should be harmless when lbRulesChanged=true and fipChanged=false. For example, port change. SafeInsert existing service on a set is idempotent.

tests/e2e/network/multiple_standard_lb.go
Liunardy: Moved the service event and IP change funcs to tests/e2e/utils/service_utils.go

tests/e2e/network/multiple_standard_lb.go
```diff
@@ -299,6 +1048,169 @@ func getLBsFromPublicIPs(tc *utils.AzureTestClient, pips []*string) sets.Set[str
 	return lbNames
 }
 
+// verifyFIPRemoved checks that the specified frontend IP configuration no longer exists on the LB.
+// Returns nil if the FIP is gone (or the entire LB is gone). Returns an error if the FIP still exists.
+func verifyFIPRemoved(tc *utils.AzureTestClient, fipID string) error {
```
nilo19: could you please move the helpers to loadbalancer utils func?

nilo19:
lgtm, thanks for the fix.

tests/e2e/network/multiple_standard_lb.go
Liunardy: Moved new lb/fip helpers, that can be easily moved, to utils.

PR #10142, opened 2026-04-04
chore: remove deprecated rand.Seed

YurDuiachenko:
#### What type of PR is this?

/kind cleanup

Removes the deprecated top-level `rand.Seed` call from cloud-controller-manager and node-manager startup.

The project builds with a modern Go version, where the global `math/rand` generator is already auto-seeded, so the explicit `rand.Seed(...)` call is no longer needed.

NONE

This is a cleanup-only change and does not change runtime behavior.
`rand.Float64()` is still used for resync jitter and continues to work as expected without an explicit seed.

```release-note
NONE

### DIFF
--- a/cmd/cloud-controller-manager/controller-manager.go
+++ b/cmd/cloud-controller-manager/controller-manager.go
@@ -20,9 +20,7 @@ limitations under the License.
 package main
 
 import (
-	"math/rand"
 	"os"
-	"time"
 
 	_ "k8s.io/component-base/metrics/prometheus/clientgo" // load all the prometheus client-go plugins
 
@@ -32,7 +30,6 @@ import (
 )
 
 func main() {
-	rand.Seed(time.Now().UnixNano()) // FIXME: should use crypto/rand for better randomness
 	defer log.Flush()
 
 	command := app.NewCloudControllerManagerCommand()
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: cmd/cloud-controller-manager/controller-manager.go [hunk 1]
```diff
@@ -32,7 +30,6 @@ import (
 )
 
 func main() {
-	rand.Seed(time.Now().UnixNano()) // FIXME: should use crypto/rand for better randomness
```
elmiko, line 35: is there anything to the comment here about using crypto/rand instead of rand?

elmiko (commented):
this seems like a reasonable cleanup but i'm not familiar enough with the rand deprecation to know if we should be using something else.

cmd/cloud-controller-manager/controller-manager.go
YurDuiachenko: Thank you a lot for taking a look, @elmiko,

In this case I don't think `crypto/rand` is needed, because this randomness is not used for anything security-sensitive. It is only used for non-cryptographic behavior such as jitter/resync timing.

The cleanup here is only removing the explicit top-level `rand.Seed(...)` call. With modern Go, the global `math/rand` generator is already automatically seeded at startup, so behavior should remain the same without requiring a replacement. https://pkg.go.dev/math/rand#Seed

YurDuiachenko:
Hm, this looks flaky

andyzhangx (approved):
/lgtm

PR #10164, opened 2026-04-08
Handle NICs without VM associations gracefully

cgiradkar:
#### What type of PR is this?

/kind bug

Load balancer creation fails completely when any node in the cluster has a NIC that is not yet fully associated with a VM. This occurs during normal cluster operations when:
1. New node pools are scaling up - VMs take time to fully provision and associate with their NICs
2. Quota or capacity constraints - VM creation is delayed while Azure allocates resources
3. Any VM provisioning delay - Network, region capacity, or transient API issues

##### Impact:

- Load balancer services cannot be created until all VMs are fully provisioned
- Blocks application deployment that requires load balancer services
- No automatic recovery - LB creation remains blocked even after VMs finish provisioning (requires manual intervention or CCM restart)
- Affects both VMSS and VMSS Flex deployments

##### Expected Behavior: 
The load balancer should be created with currently-available nodes, and automatically add delayed nodes once their VMs are fully provisioned (on subsequent sync cycles).

Fixes #

The change modifies error handling behavior in several key areas:

1. **`GetVMNameByIPConfigurationName`** - Returns `("", nil)` instead of an error when a NIC has no VM association
2. **`getVMManagementTypeByIPConfigurationID`** - Returns new `ManagedByNoVM` type when vmName is empty
3. **`GetNodeNameByIPConfigurationID`** (VMSS) - Checks for `ManagedByNoVM` and returns `("", "", nil)` early
4. **`getNodeInformationByIPConfigurationID`** (Flex) - Checks for empty vmName and returns early with logging
5. **Backend pool reconciliation** - Guards against empty nodeName in two locations:
   - `ReconcileBackendPools` - skips processing
   - `GetBackendPrivateIPs` - skips IP collection
6. **`ensureBackendPoolDeleted`** - Skips nodes with empty nodeName

Added comprehensive test coverage:
- New test case in `TestGetNodeNameByIPConfigurationID` for NIC without VM
- New test case in `TestEnsureBackendPoolDeleted` for orphaned NIC in backend pool
- New test case in `TestGetNodeNameByIPConfigurationIDVmssFlex` for orphaned NIC
- Updated `TestGetVMManagementTypeByIPConfigurationID` to expect `ManagedByNoVM` instead of error

```release-note
Handle NICs without VM associations gracefully in VMSS and VMSS Flex code paths
```

```docs
NONE
```

### DIFF
--- a/pkg/provider/azure_vmss.go
+++ b/pkg/provider/azure_vmss.go
@@ -1697,6 +1697,11 @@ func (ss *ScaleSet) GetNodeNameByIPConfigurationID(ctx context.Context, ipConfig
 		return "", "", err
 	}
 
+	if vmManagementType == ManagedByNoVM {
+		logger.V(2).Info("No VM attached, skipping node", "ipConfigurationID", ipConfigurationID)
+		return "", "", nil
+	}
+
 	if vmManagementType == ManagedByAvSet {
 		// vm is managed by availability set.
 		return ss.availabilitySet.GetNodeNameByIPConfigurationID(ctx, ipConfigurationID)
@@ -1961,6 +1966,10 @@ func (ss *ScaleSet) ensureBackendPoolDeleted(ctx context.Context, service *v1.Se
 
 			logger.Error(err, "Failed to GetNodeNameByIPConfigurationID", "ipConfigurationID", ipConfigurationID)
 			allErrs = append(allErrs, err)
+			continue
+		}
+		if nodeName == "" {
+			logger.V(2).Info("Empty nodeName, skipping node", "service", getServiceName(service), "ipConfigurationID", ipConfigurationID)
 			continue
 		}
 
--- a/pkg/provider/azure_vmss_cache.go
+++ b/pkg/provider/azure_vmss_cache.go
@@ -63,6 +63,7 @@ const (
 	ManagedByVmssFlex     VMManagementType = "ManagedByVmssFlex"
 	ManagedByAvSet        VMManagementType = "ManagedByAvSet"
 	ManagedByUnknownVMSet VMManagementType = "ManagedByUnknownVMSet"
+	ManagedByNoVM         VMManagementType = "ManagedByNoVM"
 )
 
 func (ss *ScaleSet) newVMSSCache() (azcache.Resource, error) {
@@ -530,6 +531,9 @@ func (ss *ScaleSet) getVMManagementTypeByIPConfigurationID(ctx context.Context, 
 	if err != nil {
 		return ManagedByUnknownVMSet, fmt.Errorf("failed to get vm name by ip config ID %s: %w", ipConfigurationID, err)
 	}
+	if vmName == "" {
+		return ManagedByNoVM, nil
+	}
 	if cachedAvSetVMs.Has(vmName) {
 		return ManagedByAvSet, nil
 	}
@@ -542,10 +546,14 @@ func (az *Cloud) GetVMNameByIPConfigurationName(ctx context.Context, nicResource
 	if rerr != nil {
 		return "", fmt.Errorf("failed to get interface of name %s: %w", nicName, rerr)
 	}
+	// Return empty when VM association is missing
 	if nic.Properties == nil || nic.Properties.VirtualMachine == nil || nic.Properties.VirtualMachine.ID == nil {
-		return "", fmt.Errorf("failed to get vm ID of nic %s", ptr.Deref(nic.Name, ""))
+		return "", nil
 	}
 	vmID := ptr.Deref(nic.Properties.VirtualMachine.ID, "")
+	if vmID == "" {
+		return "", nil
+	}
 	matches := vmIDRE.FindStringSubmatch(vmID)
 	if len(matches) != 2 {
 		return "", fmt.Errorf("invalid virtual machine ID %s", vmID)
--- a/pkg/provider/azure_loadbalancer_backendpool.go
+++ b/pkg/provider/azure_loadbalancer_backendpool.go
@@ -231,6 +231,11 @@ func (bc *backendPoolTypeNodeIPConfig) ReconcileBackendPools(
 						}
 					}
 
+					if nodeName == "" {
+						logger.V(2).Info("empty nodeName, skipping ipConfID", "serviceName", serviceName, "ipConfID", ipConfID)
+						continue
+					}
+
 					// If a node is not supposed to be included in the LB, it
 					// would not be in the `nodes` slice. We need to check the nodes that
 					// have been added to the LB's backendpool, find the unwanted ones and
@@ -361,6 +366,10 @@ func (bc *backendPoolTypeNodeIPConfig) GetBackendPrivateIPs(ctx context.Context,
 					nodeName, _, err := bc.VMSet.GetNodeNameByIPConfigurationID(ctx, ipConfigID)
 					if err != nil {
 						logger.Error(err, "failed: GetNodeNameByIPConfigurationID", "service", serviceName)
+						continue
+					}
+					if nodeName == "" {
+						logger.V(2).Info("empty nodeName, skipping ipConfigID", "serviceName", serviceName, "ipConfigID", ipConfigID)
 						continue
 					}
 					privateIPsSet, ok := bc.nodePrivateIPs[strings.ToLower(nodeName)]
--- a/pkg/provider/azure_vmss_cache_test.go
+++ b/pkg/provider/azure_vmss_cache_test.go
@@ -18,7 +18,6 @@ package provider
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"net/http"
 	"testing"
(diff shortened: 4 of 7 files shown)


cgiradkar:
@feiskyer  @nilo19 @andyzhangx would you be able to have a look at this PR please?

nilo19:
will check.

cgiradkar:
@nilo19 can I get a ok-to-test in the meanwhile. This bug is having some impact in my code, so I'd thankful if you review this small PR sooner. Thanks

pkg/provider/azure_vmss.go
```diff
@@ -1963,6 +1967,9 @@ func (ss *ScaleSet) ensureBackendPoolDeleted(ctx context.Context, service *v1.Se
 			allErrs = append(allErrs, err)
 			continue
 		}
+		if nodeName == "" {
+			continue
```
marek-veber, line 1973: IMHO: Some logger.Info(...) will be useful here.

pkg/provider/azure_vmss.go
```diff
@@ -1697,6 +1697,10 @@ func (ss *ScaleSet) GetNodeNameByIPConfigurationID(ctx context.Context, ipConfig
 		return "", "", err
 	}
 
+	if vmManagementType == ManagedByUnknownVMSet {
+		return "", "", nil
```
marek-veber, line 1702: IMHO: Some logger.Info(...) will be useful here.

marek-veber (changes requested):
The fix is directionally correct and solves a real user-facing issue. I'd want to see:                                                                                                                            
  - Logging added to the VMSS paths for parity with VMSS Flex
  - At least one test per new branch (especially the ensureBackendPoolDeleted skip)                                                                                                                                 
  - A quick audit of other GetNodeNameByIPConfigurationID callers

Concerns:
 1. Insufficient test coverage. Only the existing TestGetVMManagementTypeByIPConfigurationID test was updated. The new behavior in these functions lacks test cases:                                               
    - GetNodeNameByIPConfigurationID — no test for the ManagedByUnknownVMSet early return                                                                                                                           
    - ensureBackendPoolDeleted — no test for the nodeName == "" skip                                                                                                                                                
    - getNodeInformationByIPConfigurationID — no test for the empty vmName path                                                                                                                                     
 2. Callers of GetNodeNameByIPConfigurationID beyond ensureBackendPoolDeleted. The PR adds an empty-string return for ManagedByUnknownVMSet, and ensureBackendPoolDeleted handles it with a continue. But if there  are other callers of GetNodeNameByIPConfigurationID elsewhere in the codebase, they may not expect ("", "", nil) and could misbehave (e.g., passing an empty node name to a downstream function). This should be audited.                                                                                                                                                                                                          
  3. ManagedByUnknownVMSet semantics are overloaded. Before this PR, ManagedByUnknownVMSet with a non-nil error meant "something went wrong." Now it's also used (with nil error) to mean "NIC exists but no VM yet." This dual meaning could confuse future readers. A dedicated constant like ManagedByNoVM or a comment clarifying the nil-error variant would improve clarity.

pkg/provider/azure_vmss.go
cgiradkar: added logs aptly

pkg/provider/azure_vmss.go
cgiradkar: added logs aptly

cgiradkar:
> The fix is directionally correct and solves a real user-facing issue. I'd want to see:
> 
> * Logging added to the VMSS paths for parity with VMSS Flex
> * At least one test per new branch (especially the ensureBackendPoolDeleted skip)
> * A quick audit of other GetNodeNameByIPConfigurationID callers
> 
> Concerns:
> 
> 1. Insufficient test coverage. Only the existing TestGetVMManagementTypeByIPConfigurationID test was updated. The new behavior in these functions lacks test cases:
>    
>    * GetNodeNameByIPConfigurationID — no test for the ManagedByUnknownVMSet early return
>    * ensureBackendPoolDeleted — no test for the nodeName == "" skip
>    * getNodeInformationByIPConfigurationID — no test for the empty vmName path
> 2. Callers of GetNodeNameByIPConfigurationID beyond ensureBackendPoolDeleted. The PR adds an empty-string return for ManagedByUnknownVMSet, and ensureBackendPoolDeleted handles it with a continue. But if there  are other callers of GetNodeNameByIPConfigurationID elsewhere in the codebase, they may not expect ("", "", nil) and could misbehave (e.g., passing an empty node name to a downstream function). This should be audited.
> 3. ManagedByUnknownVMSet semantics are overloaded. Before this PR, ManagedByUnknownVMSet with a non-nil error meant "something went wrong." Now it's also used (with nil error) to mean "NIC exists but no VM yet." This dual meaning could confuse future readers. A dedicated constant like ManagedByNoVM or a comment clarifying the nil-error variant would improve clarity.

- [✓ ] Add logs aptly
- [✓ ] Add more unit tests for coverage over new code
- [ ✓] Audit callers of GetNodeNameByIPConfigurationID for empty-string safety
- [ ✓] Improve ManagedByUnknownVMSet semantics

cgiradkar:
@nilo19 @cheftako can you PTAL

pkg/provider/azure_vmss_cache.go
```diff
@@ -542,10 +546,14 @@ func (az *Cloud) GetVMNameByIPConfigurationName(ctx context.Context, nicResource
 	if rerr != nil {
 		return "", fmt.Errorf("failed to get interface of name %s: %w", nicName, rerr)
 	}
+	// Return empty when VM association is missing
 	if nic.Properties == nil || nic.Properties.VirtualMachine == nil || nic.Properties.VirtualMachine.ID == nil {
-		return "", fmt.Errorf("failed to get vm ID of nic %s", ptr.Deref(nic.Name, ""))
+		return "", nil
```
nilo19, line 551: Do you manually create the nic? When will the nic has no vm association? If this is a problem we must have been received a lot of tickets, but we haven't. May I know if there is a special use case from your side?

cgiradkar: A customer had reported that when creating a Kubernetes service of type LoadBalancer, the operation fails if any VMs in the node pool are still provisioning or stuck in creation. The expectation is that the load balancer should skip VMs that are not yet ready and add them later once they complete provisioning, rather than failing the entire operation.

```
SyncLoadBalancerFailed

Error syncing load balancer: failed to ensure load balancer: failed to get vm name by ip config ID /subscriptions/.../xxxx-xxxx-xxxx-nic/ipConfigurations/pipConfig: failed to get vm ID of xxxx-xxxx-xxxx-nic
```

cgiradkar: ### Steps to Reproduce

1. Create an ARO-HCP cluster
2. Create a node pool that requests VMs exceeding existing quota (to simulate delayed/stuck provisioning)
3. Wait until node provisioning gets stuck or delayed
4. Attempt to provision a service of type LoadBalancer

nilo19: > A customer had reported that when creating a Kubernetes service of type LoadBalancer, the operation fails if any VMs in the node pool are still provisioning or stuck in creation. The expectation is that the load balancer should skip VMs that are not yet ready and add them later once they complete provisioning, rather than failing the entire operation.
> 
> ```
> SyncLoadBalancerFailed
> 
> Error syncing load balancer: failed to ensure load balancer: failed to get vm name by ip config ID /subscriptions/.../xxxx-xxxx-xxxx-nic/ipConfigurations/pipConfig: failed to get vm ID of xxxx-xxxx-xxxx-nic
> ```

If this is the case, in the next reconcile when the vm has been ready, it should auto fixed.

nilo19: BTW, what is ARO-HCP cluster? Is this issue only in ARO-HCP clusters?

cgiradkar: > BTW, what is ARO-HCP cluster? Is this issue only in ARO-HCP clusters?

ARO-HCP: https://github.com/Azure/ARO-HCP : uses standard OpenShift cloud-controller-manager (which is based on cloud-provider-azure)
This issue will occur in any cluster using `cloud-provider-azure`

cgiradkar: > If this is the case, in the next reconcile when the vm has been ready, it should auto fixed.

In this specific case, the whole array of VMS are never gonna get provisioned (due to said constraints), so waiting for next cycle wont do it as the state didnt change so no processing would follow from this file.
This PR makes the process of VM state transiotion (among an array of VMs to be provisioned) more granular even if the re-sync duration is bearable.

nilo19: But if there is a genuine  issue from the NRP/NIC side, this change will silently hide the error.

cgiradkar: would it be an acceptable solution of we log it and proceed? This use case is an edge case but still encountered in production.

nilo19: I think it would be more reasonable to solve the root cause, so the ccm reconcile will be unblocked. This change introduces behavior change to solve an edge case, and I'm not sure if it's worth it, as it may introduce other issues.

cgiradkar:
PR [cluster-api-provider-azure #6569](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/6569) merged, closing this PR.

PR #10188, opened 2026-04-15
feature: add env variables for Kubernetes service

MehdiRtal:
#### What type of PR is this?
/kind feature

**The Problem:** When a Pod is running in hostNetwork: true mode, it inherits the network namespace of the host. By default, Kubernetes injects KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT pointing to the internal ClusterIP of the API server. In certain Azure networking topologies (or when using specific local proxies), these default values can be unreachable or conflict with host-level services.

**The Fix:** This PR introduces the ability to specify custom values for these variables. This ensures that host-networked components (like the Azure Cloud Controller Manager or Node-level agents) can reliably reach the API server via a specific endpoint (e.g., a Load Balancer IP or localhost) rather than the default ClusterIP.

NONE

Allows overriding KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT for components running in hostNetwork mode to improve connectivity in complex Azure networking environments.


nilo19:
@mboersma what do you think of this change?

mboersma:
@nilo19 sorry, I missed this until now!

This seems like a safe, backward-compatible change. But I'd like to understand the use case better. 

The CNM/CCM only fall back to in-cluster config (which reads these env vars) when no  `--kubeconfig` is set, and the Windows CNM always uses a kubeconfig so it wouldn't be affected. Could you describe a concrete scenario where the default `KUBERNETES_SERVICE_HOST / PORT`  were unreachable and this override fixed it?

Also, can we drop the changes to `helm/repo/` (the `.tgz` and `index.yaml`)? The master branch is already on chart 1.36.0, so this conflicts.

PR #10211, opened 2026-04-22
fix(multi-slb): support switching internal/external when IP sharing across multiple services

### LINKED ISSUE #10050: [Multi-SLB] external->internal transition can lose current-LB hint and target sibling LB
#### What happened:

  In multi-SLB mode, after manually flipping a `Service` from external to internal, the service did not stay on its current LB configuration on the next reconcile.

  The service was originally associated with one LB config (for example `lb1`). After the external->internal flip, the controller correctly reconciled the service onto the internal sibling LB (`lb1-internal`). But a
  later reconcile no longer considered the service to be “currently on `lb1`” and instead tried to migrate it to another eligible LB that did not even exist yet (for example `lb2-internal`).

  This looks like `activeServices` state for the current LB config is being removed during the opposite-LB cleanup reconcile.

  After flipping a service from external to internal in multi-SLB mode, the service should stay on the same LB configuration (`lb1` -> `lb1-internal`) unless placement rules actually require it to move.

  A follow-up reconcile should keep the service on its current LB configuration, not treat it as unplaced and attempt migration to another eligible LB.

  1. Enable multi-SLB with at least two eligible LB configs, for example `lb1` and `lb2`.
  2. Create an external `type: LoadBalancer` service that lands on `lb1`.
  3. Manually flip the service to internal by setting `service.beta.kubernetes.io/azure-load-balancer-internal: "true"`.
  4. Let the controller reconcile once so the service is moved onto `lb1-internal`.
  5. Trigger another reconciliation for the same service.
  6. Observe that the controller may now try to move the service to another eligible LB, especially an eligible LB whose internal sibling resource does not exist yet (for example `lb2-internal`), instead of keeping the
  service on `lb1-internal`.

  I traced the failure path to `ActiveServices` bookkeeping in `pkg/provider/azure_loadbalancer.go`.

  The problem appears to be:

  - `reconcileService()` first reconciles the desired service, then always runs a cleanup reconcile for the flipped service to remove the opposite LB flavor:
    - https://github.com/kubernetes-sigs/cloud-provider-azure/blob/4b5dcc8807a02efd86e94f5b8bf7fc199ae3ddeb/pkg/provider/azure_loadbalancer.go#L174-L179
  - During that cleanup reconcile, if the old frontend IP config is removed, `fipChanged` becomes true and the controller updates multi-SLB status with `wantLb=false`:
    - https://github.com/kubernetes-sigs/cloud-provider-azure/blob/4b5dcc8807a02efd86e94f5b8bf7fc199ae3ddeb/pkg/provider/azure_loadbalancer.go#L2077-L2078
    - https://github.com/kubernetes-sigs/cloud-provider-azure/blob/4b5dcc8807a02efd86e94f5b8bf7fc199ae3ddeb/pkg/provider/azure_loadbalancer.go#L2582-L2608
  - `reconcileMultipleStandardLoadBalancerConfigurationStatus()` trims the `-internal` suffix before updating `ActiveServices`, so external and internal sibling LBs share the same config-level status bucket:
    - https://github.com/kubernetes-sigs/cloud-provider-azure/blob/4b5dcc8807a02efd86e94f5b8bf7fc199ae3ddeb/pkg/provider/azure_loadbalancer.go#L2446-L2459
  - That means cleanup of the old external LB can delete the same `ActiveServices["lb1"]` entry that is needed to remember the service is now active on `lb1-internal`.
  - On the next reconcile, current LB selection relies only on that config-level `ActiveServices` state:
    - https://github.com/kubernetes-sigs/cloud-provider-azure/blob/4b5dcc8807a02efd86e94f5b8bf7fc199ae3ddeb/pkg/provider/azure_loadbalancer.go#L4115-L4116
    - https://github.com/kubernetes-sigs/cloud-provider-azure/blob/4b5dcc8807a02efd86e94f5b8bf7fc199ae3ddeb/pkg/provider/azure_loadbalancer.go#L4180-L4186
  - Once the current LB name is empty, `getMostEligibleLBForService()` explicitly prefers the first eligible LB that does not exist yet:
    - https://github.com/kubernetes-sigs/cloud-provider-azure/blob/4b5dcc8807a02efd86e94f5b8bf7fc199ae3ddeb/pkg/provider/azure_loadbalancer.go#L4125-L4152

  So the issue is not just the later selection logic. The underlying bug is that opposite-LB cleanup removes the config-level `ActiveServices` entry even when the service is still active on the sibling LB of the same
  config.

  This also looks symmetric for internal->external flips.

  I also ran nearby provider tests locally:
  `go test -run 'TestGetMostEligibleLBName|TestGetServiceLoadBalancerMultiSLB|TestEnsureLoadBalancerDeleted' ./pkg/provider`
  Those passed, which suggests this path is missing regression coverage rather than already being covered by a failing test.

  #### Environment:

  - Kubernetes version (use `kubectl version`): `<fill in>`
  - Cloud provider or hardware configuration: Azure, Standard LoadBalancer SKU, multi-SLB enabled
  - OS (e.g: `cat /etc/os-release`): `<fill in>`
  - Kernel (e.g. `uname -a`): `<fill in>`
  - Install tools: `<fill in>`
  - Network plugin and version (if this is a network-related bug): `<fill in>`

### LINKED ISSUE #10117: [Multi-SLB] support switching internal/external when IP sharing across multiple services
#### What happened:
#9937 introduced a bug where load balancer rules and probes are not properly cleaned up or service migration is incorrectly blocked during service external/internal transitions.

1. Secondary service transitions
    - Two Services share the same frontend IP.
    - Change the secondary Service to internal.
    - LB rules and probes are not cleaned up.

1. Secondary service + LB config change
    - Two Services share the same frontend IP.
    - Change the secondary Service to internal and annotate it to a different LB.
    - LB rules and probes are not cleaned up.

1. Primary service external → internal → external
    - Two Services share an IP on lb-1.
    - Change the primary Service to internal (lb-1-internal).
    - Change it back to external and remove LB config annotation.
    - Reconciliation is blocked with error (cannot migrate from load balancer "lb-1" to "lb-2").
    - At this point, the controller loses the original LB hint, picks a sibling LB (lb-2) with fewer rules, incorrectly blocks reuse of the existing shared IP.

1. Primary service external → internal + target lb-2 → external
    - Two Services share an IP on lb-1.
    - Change the primary Service to internal and annotate it to lb-2-internal.
    - Change it back to external and remove LB config annotation.
    - Reconciliation is blocked with error

Load balancer rules and probes should be cleaned up correctly when a Service transitions from external ↔ internal, regardless of whether it is the primary or secondary Service sharing the IP.
External ↔ internal transitions should not block reuse of an existing shared frontend IP.
Behavior should be consistent with the non-multi-SLB case, where the existing IP is reused.

- Enable multi-SLB with at least two eligible LB configs, for example lb-1 and lb-2.
- Create 2 external type: LoadBalancer services sharing the same frontend IP that lands on lb-1.
- Test the following internal transitions by setting service.beta.kubernetes.io/azure-load-balancer-internal: "true":
   - Change the secondary service to internal and remove the IP pin.
   - Change the secondary service to internal, remove the IP pin, and annotate it to lb-2.
   - Change the primary service from external → internal → external.
   - Change the primary service from external → internal and annotate it to lb-2 → external and remove lb annotation.
- Observe:
   - Orphaned LB rules / probes.
   - Blocked reconciliation due to perceived LB migration despite reuse of the same IP.

This is a follow-up bug introduced by #9937 to fix #9867
Likely overlaps in root cause with #10050 

#### Environment:

- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

Liunardy:
#### What type of PR is this?
/kind bug

Follow-up to #9937. When services sharing a frontend IP switch between external and internal in multi-SLB mode, several issues arise:

1. Stale rules/probes on opposite-type LB: When a secondary service switches from external to internal (or vice versa), load balancing rules and probes are not cleaned up from the old LB.
2. Primary service round-trip blocked: When a primary service goes external → internal → external, the controller loses the original LB hint and incorrectly tries to migrate to a sibling LB, blocking reconciliation.
3. ActiveServices eviction: The wantLb=false cleanup reconcile for the old LB type removes the config-level ActiveServices entry, even though the service is still active on the sibling (internal/external) LB of the same config.

This PR adds an opposite-type cleanup block in getServiceLoadBalancer that detects stale rules/probes on the wrong-type LB during the wantLb=true reconcile and removes them proactively. It also adds frontend IP name prefix matching (getLoadBalancerNameByFrontendIPName) so that services without pinned IPs correctly identify their current LB when their FIP already exists.

Fixes #10117
Fixes #10050

```release-note
fix(multi-slb): support switching internal/external when IP sharing across multiple services

Correctly clean up stale rules and probes when services sharing a frontend IP switch between external and internal in multi-SLB mode.
```

```docs

```

### DIFF
--- a/pkg/provider/azure_loadbalancer.go
+++ b/pkg/provider/azure_loadbalancer.go
@@ -826,6 +826,53 @@ func (az *Cloud) getServiceLoadBalancer(
 			defaultLB = existingLB
 		}
 		if isInternalLoadBalancer(existingLB) != isInternal {
+			// External<->internal LB cleanup (multi-SLB only): after switching a service
+			// between external and internal, the opposite LB may still have stale
+			// rules/probes from this service. Clean them up here during the wantLb=true
+			// reconcile so that the subsequent wantLb=false reconcile finds nothing to
+			// remove and won't incorrectly evict the service from ActiveServices.
+			if wantLb && az.UseMultipleStandardLoadBalancers() &&
+				az.lbHasServiceOwnedResources(existingLB, service) {
+
+				deletedLBName, deletedPLS, err := az.removeStaleServiceLBResources(ctx, existingLB, existingLBs, clusterName, service)
+				if err != nil {
+					return nil, nil, nil, nil, false, false, fmt.Errorf("remove stale service resources from opposite-type load balancer %q for service %q: %w",
+						ptr.Deref(existingLB.Name, ""), getServiceName(service), err)
+				}
+				if deletedPLS {
+					return nil, nil, nil, nil, false, true, nil
+				}
+				logger.V(2).Info("Removed stale service resources from opposite-type load balancer",
+					"service", service.Name,
+					"oldLB", ptr.Deref(existingLB.Name, ""),
+					"targetLB", defaultLBName,
+				)
+
+				if deletedLBName != "" {
+					removeLBFromList(&existingLBs, deletedLBName)
+				}
+
+				az.reconcileMultipleStandardLoadBalancerConfigurationStatus(
+					false,
+					getServiceName(service),
+					ptr.Deref(existingLB.Name, ""),
+				)
+
+				if isLocalService(service) {
+					svcName := getServiceName(service)
+					if az.backendPoolUpdater != nil {
+						az.backendPoolUpdater.removeOperation(svcName)
+					}
+					if deletedLBName == "" {
+						newLBs, err := az.cleanupLocalServiceBackendPool(ctx, service, nodes, existingLBs, clusterName)
+						if err != nil {
+							return nil, nil, nil, nil, false, false, fmt.Errorf("clean up backend pool for local service %q on opposite-type load balancer %q: %w",
+								getServiceName(service), ptr.Deref(existingLB.Name, ""), err)
+						}
+						existingLBs = newLBs
+					}
+				}
+			}
 			continue
 		}
 
@@ -4322,6 +4369,7 @@ func (az *Cloud) getAzureLoadBalancerName(
 		lbIPs := getServiceLoadBalancerIPs(service)
 		pipNames := getServicePIPNames(service)
 		pinsIP := len(lbIPs) > 0 || len(pipNames) > 0
+		annotatedLBs := consts.GetLoadBalancerConfigurationsNames(service)
 
 		// Only block when creating resources (wantLb=true).
 		// Cleanup operations (wantLb=false) don't need this constraint.
@@ -4330,7 +4378,6 @@ func (az *Cloud) getAzureLoadBalancerName(
 			// When specifying an IP, the LB is determined by where the IP resides.
 			// Specifying a different LB is contradictory. This early check also
 			// avoids unnecessary PIP lookups on every reconcile.
-			annotatedLBs := consts.GetLoadBalancerConfigurationsNames(service)
 			logger.V(5).Info("checking LB and IP config", "service", service.Name, "annotatedLBs", annotatedLBs, "lbIPs", lbIPs, "pipNames", pipNames, "pinsIP", pinsIP)
 			if len(annotatedLBs) > 0 && pinsIP {
 				return "", fmt.Errorf(
@@ -4351,6 +4398,14 @@ func (az *Cloud) getAzureLoadBalancerName(
 		}
 
 		currentLBName := az.getServiceCurrentLoadBalancerName(service)
+
+		// If a service does not pin a specific IP, has no LB annotation, and its primary frontend IP already exists
+		// on a same-kind (internal/external) LB, use that LB.
+		if wantLb && !pinsIP && len(annotatedLBs) == 0 {
+			if fipLBName := az.getLoadBalancerNameByFrontendIPName(ctx, service, existingLBs, isInternal); fipLBName != "" {
+				currentLBName = fipLBName
+			}
+		}
 
 		// If service specifies an IP, find which LB has that IP. The service should use the LB where the IP resides.
 		// This also detects secondary services sharing an IP with a primary service already on an LB.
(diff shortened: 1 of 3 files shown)


pkg/provider/azure_loadbalancer.go
```diff
@@ -826,6 +826,53 @@ func (az *Cloud) getServiceLoadBalancer(
...
 		if isInternalLoadBalancer(existingLB) != isInternal {
+			// External<->internal LB cleanup (multi-SLB only): after switching a service
+			// between external and internal, the opposite LB may still have stale
+			// rules/probes from this service. Clean them up here during the wantLb=true
+			// reconcile so that the subsequent wantLb=false reconcile finds nothing to
+			// remove and won't incorrectly evict the service from ActiveServices.
+			if wantLb && az.UseMultipleStandardLoadBalancers() &&
+				az.lbHasServiceOwnedResources(existingLB, service) {
+
+				deletedLBName, deletedPLS, err := az.removeStaleServiceLBResources(ctx, existingLB, existingLBs, clusterName, service)
```
nilo19: Seems we have many duplicates in this function. Can we reuse one helper function here and the block inside if status == nil?

nilo19:
May consider to add following UTs:
1. ActiveServices says lb2 but the same-kind FIP lives on lb1 (both eligible).
2. getLoadBalancerNameByFrontendIPName returns an LB no longer in the eligible set.

pkg/provider/azure_loadbalancer.go
```diff
@@ -4384,6 +4423,36 @@ func (az *Cloud) getAzureLoadBalancerName(
...
+		if isInternalLoadBalancer(lb) != isInternal {
+			continue
+		}
+		if lb.Properties == nil || lb.Properties.FrontendIPConfigurations == nil {
+			continue
+		}
+		for _, fip := range lb.Properties.FrontendIPConfigurations {
+			if strings.HasPrefix(ptr.Deref(fip.Name, ""), baseName) {
+				lbName := trimSuffixIgnoreCase(ptr.Deref(lb.Name, ""), consts.InternalLoadBalancerNameSuffix)
+				logger.V(2).Info("Found LB by frontend IP name prefix", "service", service.Name, "fipNamePrefix", baseName, "lbName", lbName)
```
nilo19: use V(4) or 5.

Liunardy: Changed to V(4). Also changed the verbosity level for 2 other LB name retrieval functions.

pkg/provider/azure_loadbalancer.go
```diff
@@ -826,6 +826,28 @@ func (az *Cloud) getServiceLoadBalancer(
...
 		if isInternalLoadBalancer(existingLB) != isInternal {
+			// External<->internal LB cleanup (multi-SLB only): after switching a service
+			// between external and internal, the opposite LB may still have stale
+			// rules/probes from this service. Clean them up here during the wantLb=true
+			// reconcile so that the subsequent wantLb=false reconcile finds nothing to
+			// remove and won't incorrectly evict the service from ActiveServices.
+			if wantLb && az.UseMultipleStandardLoadBalancers() &&
+				az.lbHasServiceOwnedResources(existingLB, service) {
+
+				existingLBs, deletedPLS, err = az.removeServiceFromLB(ctx, existingLB, existingLBs, clusterName, service, nodes, nil)
```
nilo19, line 839: may add a guard comment to make sure removeServiceFromLB only deletes the current LB. if it deletes other LBs in the future, the lb for loop may skip some lb items.

nilo19:
lgtm overall, despite some nit issues.

pkg/provider/azure_loadbalancer.go
Liunardy: Added comments to the removeServiceFromLB funcion and before the loop, see https://github.com/kubernetes-sigs/cloud-provider-azure/pull/10211/changes/9217234aad8d4b60f4a733432a5bc11243deb885

PR #10328, opened 2026-05-12
fix(multi-slb): serialize backendPoolUpdater with service reconcile loop

### LINKED ISSUE #9839: Potential race: backendPoolUpdater concurrent with Service LB reconciliation for local services
#### What happened:

In multiple standard load balancer mode, local services (`ExternalTrafficPolicy=Local`) can trigger backend pool updates from two independent paths:

1. The main service reconciliation loop (`EnsureLoadBalancer` / `UpdateLoadBalancer`).
2. The background `backendPoolUpdater` driven by EndpointSlice updates.

These paths do not share the same serialization mechanism, so they can concurrently call ARM `BackendAddressPool` `CreateOrUpdate` for the same backend pool. This can lead to intermittent ARM conflicts (e.g. ETag precondition failures / operation canceled), noisy warning events (`LoadBalancerBackendPoolUpdateFailed`), and slower convergence during EndpointSlice churn or service LB migration.

Code pointers:
- Updater write path: `pkg/provider/azure_local_services.go` (`loadBalancerBackendPoolUpdater.process` -> `BackendAddressPoolClient.Get/CreateOrUpdate`).
- Main reconcile write path: `pkg/provider/azure_loadbalancer_backendpool.go` (`EnsureHostsInPool` -> `CreateOrUpdateLBBackendPool`).
- Main reconcile serialization: `pkg/provider/azure_loadbalancer.go` (`serviceReconcileLock`) and optional `azureResourceLocker` (lease-based).

Backend pool updates should be serialized with the main service reconciliation loop (e.g., share `serviceReconcileLock` and/or `azureResourceLocker`), or otherwise handle transient ARM conflicts without emitting failure events.

1. Enable multiple standard load balancers in cloud-provider-azure.
2. Create a `LoadBalancer` Service with `ExternalTrafficPolicy=Local`.
3. Cause rapid EndpointSlice changes (e.g., scale a Deployment up/down) while the service is being reconciled (initial LB creation, or while moving the service between LBs).
4. Observe intermittent backend pool update failures in CCM logs and/or Service events.

- The updater currently performs ARM calls while holding its internal queue mutex; if serialization is added, it likely should dequeue under the queue mutex and then do ARM operations outside of it.
- The updater does not currently use `azureResourceLocker`, so in HA deployments multiple CCM replicas could also run the updater concurrently.

#### Environment:

- Cloud provider or hardware configuration: Azure (cloud-provider-azure), multiple standard load balancers enabled
- Install tools: cloud-controller-manager + cloud-provider-azure
- Others: Services with `ExternalTrafficPolicy=Local` (EndpointSlice-driven updates)

Liunardy:
#### What type of PR is this?
/kind bug

The `backendPoolUpdater.process()` performs backend pool updates without holding `serviceReconcileLock` or `azureResourceLocker`, allowing concurrent writes to the same backend pool from the updater and the main service reconciliation path.

This PR serializes the updater with the main reconciliation loop by acquiring `serviceReconcileLock` and `azureResourceLocker` in `process()` before ARM calls. The `updater.lock` is released before acquiring `serviceReconcileLock` and `azureResourceLocker` to avoid deadlock with the main path, which holds `serviceReconcileLock` and calls `addOperation()` (which acquires the `updater.lock`).

If the distributed lease lock cannot be acquired, operations are re-queued for retry on the next tick.

Fixes #9839

```release-note
fix: serialize backendPoolUpdater with service reconciliation to prevent concurrent backend pool writes in multi-standard-load-balancer configurations with externalTrafficPolicy: Local
```

```docs

```

### DIFF
--- a/pkg/provider/azure_local_services.go
+++ b/pkg/provider/azure_local_services.go
@@ -160,13 +160,15 @@ func (updater *loadBalancerBackendPoolUpdater) removeOperation(serviceName strin
 	}
 }
 
-// drainOperations drains all pending operations from the queue, groups them
-// by loadBalancerName:backendPoolName, and returns the groups. The queue is
-// cleared before returning.
-//
-// The caller must not hold updater.lock when calling this method.
-func (updater *loadBalancerBackendPoolUpdater) drainOperations() map[string][]batchOperation {
-	logger := log.Background().WithName("loadBalancerBackendPoolUpdater.drainOperations")
+// countOperations returns the number of pending operations in the queue.
+func (updater *loadBalancerBackendPoolUpdater) countOperations() int {
+	updater.lock.Lock()
+	defer updater.lock.Unlock()
+	return len(updater.operations)
+}
+
+// drainOperations drains all pending operations from the queue and clears it.
+func (updater *loadBalancerBackendPoolUpdater) drainOperations() []batchOperation {
 	updater.lock.Lock()
 	defer updater.lock.Unlock()
 
@@ -174,9 +176,19 @@ func (updater *loadBalancerBackendPoolUpdater) drainOperations() map[string][]ba
 		return nil
 	}
 
-	// Group operations by loadBalancerName:backendPoolName
+	ops := updater.operations
+	updater.operations = make([]batchOperation, 0)
+	return ops
+}
+
+// groupOperations filters and groups operations by loadBalancerName:backendPoolName.
+// Must be called under serviceReconcileLock so that
+// localServiceNameToServiceInfoMap reads are consistent.
+func (updater *loadBalancerBackendPoolUpdater) groupOperations(ctx context.Context, ops []batchOperation) map[string][]batchOperation {
+	logger := log.FromContextOrBackground(ctx).WithName("loadBalancerBackendPoolUpdater.groupOperations")
+
 	groups := make(map[string][]batchOperation)
-	for _, op := range updater.operations {
+	for _, op := range ops {
 		lbOp := op.(*loadBalancerBackendPoolUpdateOperation)
 		si, found := updater.az.getLocalServiceInfo(strings.ToLower(lbOp.serviceName))
 		if !found {
@@ -195,21 +207,7 @@ func (updater *loadBalancerBackendPoolUpdater) drainOperations() map[string][]ba
 		groups[key] = append(groups[key], op)
 	}
 
-	// Clear all jobs.
-	updater.operations = make([]batchOperation, 0)
-
 	return groups
-}
-
-// requeueOperations puts operations back into the queue so they can be
-// retried on the next tick.
-func (updater *loadBalancerBackendPoolUpdater) requeueOperations(groups map[string][]batchOperation) {
-	updater.lock.Lock()
-	defer updater.lock.Unlock()
-
-	for _, ops := range groups {
-		updater.operations = append(updater.operations, ops...)
-	}
 }
 
 // process processes all operations in the loadBalancerBackendPoolUpdater.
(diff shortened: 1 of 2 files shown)


pkg/provider/azure_local_services.go
```diff
@@ -199,6 +198,66 @@ func (updater *loadBalancerBackendPoolUpdater) process(ctx context.Context) {
...
+	}
+
+	// Serialize with the main service reconciliation loop.
+	updater.az.serviceReconcileLock.Lock()
+	defer updater.az.serviceReconcileLock.Unlock()
+
+	// Serialize across multiple CCM replicas in HA deployments.
+	if updater.az.azureResourceLocker != nil {
+		if err := updater.az.azureResourceLocker.Lock(ctx); err != nil {
+			logger.Error(fmt.Errorf(
```
nilo19: No need to re-log the error, just use a V(2) message to log the requeue behavior.

pkg/provider/azure_local_services.go
```diff
@@ -199,6 +198,66 @@ func (updater *loadBalancerBackendPoolUpdater) process(ctx context.Context) {
...
+// if it is retriable, otherwise all operations in the batch targeting to
+// this backend pool will fail.
+func (updater *loadBalancerBackendPoolUpdater) process(ctx context.Context) {
+	logger := log.FromContextOrBackground(ctx).WithName("loadBalancerBackendPoolUpdater.process")
+
+	// Drain under updater.lock first, then acquire serviceReconcileLock.
+	// The reverse order would deadlock because the main reconcile path
+	// holds serviceReconcileLock and calls addOperation, which acquires
+	// updater.lock.
+	groups := updater.drainOperations()
```
nilo19: This design seems wrong.
1. removeOperation calls from main loop may be useless after drainOpeartion.
2. during the backendpool updater waiting for lock, the service can be changed, e.g., moving from 1 lb to another.

pkg/provider/azure_local_services.go
```diff
@@ -199,6 +198,66 @@ func (updater *loadBalancerBackendPoolUpdater) process(ctx context.Context) {
...
+
+	// Serialize across multiple CCM replicas in HA deployments.
+	if updater.az.azureResourceLocker != nil {
+		if err := updater.az.azureResourceLocker.Lock(ctx); err != nil {
+			logger.Error(fmt.Errorf(
+				consts.AzureResourceLockFailedToLockErrorTemplate,
+				"loadBalancerBackendPoolUpdater.process",
+				err,
+			), "Re-queuing operations for retry")
+			updater.requeueOperations(groups)
```
nilo19: Maybe can be fixed in the later refinement PR, but need to call these out:
1. Unbounded retry. If another component is holding the lock, we will requeue at each tick and never stop doing that. Do we need a retry policy?
2. For the next tick where the requeued operation is processed again, is it still a valid operation? Do we need to guard before processing?

pkg/provider/azure_local_services_test.go
```diff
@@ -662,3 +664,276 @@ func TestCheckAndApplyLocalServiceBackendPoolUpdates(t *testing.T) {
...
+		},
+		{
+			name:          "keeps valid operation when another is filtered",
+			operations:    []batchOperation{addPool1, addPool2},
+			localServices: map[string]*serviceInfo{"ns1/svc1": {lbName: "lb1"}},
+			expectedGroups: map[string][]batchOperation{
+				"lb1:pool1": {addPool1},
+			},
+		},
+	}
```
nilo19, line 877: Please also cover negative cases such as failed to acquire azure resource lock.

pkg/provider/azure_local_services.go
anndono: Should `ObserveOperationWithResult` be called at the end of each iteration rather than deferred? Since defer executes at `process` function exit, the recorded latency for earlier iterations will include the processing time of all subsequent operations.

Liunardy: Good catch! I was only checking isOperationSucceeded value and didn't check when the defer executes. Removed the defer and call `ObserveOperationWithResult` directly instead.

pkg/provider/azure_local_services_test.go
Liunardy: Added azureResourceLocker unlock fail case `TestLoadBalancerBackendPoolUpdaterCompletesOnUnlockFailure`. azureResourceLocker lock fail case is now `TestLoadBalancerBackendPoolUpdaterPreservesOperationsOnLeaseLockFailure`.

pkg/provider/azure_local_services.go
Liunardy: Operations are now preserved in the queue when lock fails (they were never drained). On the next tick, `groupOperations` validates each operation against the current `localServiceNameToServiceInfoMap` under `serviceReconcileLock`, so stale operations should be filtered out.

pkg/provider/azure_local_services.go
Liunardy: Good catch! Restructured `process()` to acquire `serviceReconcileLock` before draining. This fixes both issues:
1. `removeOperation` now works because the main reconcile loop holds `serviceReconcileLock`, which blocks `process()` before it can drain. Operations are still in the queue when `removeOperation` is called.
1. `groupOperations` reads `localServiceNameToServiceInfoMap` under `serviceReconcileLock`, so it sees the latest LB assignment and filters operations targeting the old LB.

Added tests for both cases: `TestLoadBalancerBackendPoolUpdaterFiltersOperationsWhenLBChangedDuringProcess` and `TestLoadBalancerBackendPoolUpdaterRemoveOperationCancelsOperationsBeforeDrain`.

pkg/provider/azure_local_services.go
Liunardy: Removed redundant error log. Operations are now preserved (not requeued) when lock fails, so no log needed. Also removed the unlock error log.

pkg/provider/azure_local_services.go
```diff
@@ -160,24 +160,35 @@ func (updater *loadBalancerBackendPoolUpdater) removeOperation(serviceName strin
...
+	ops := updater.operations
+	updater.operations = make([]batchOperation, 0)
+	return ops
+}
+
+// groupOperations filters and groups operations by loadBalancerName:backendPoolName.
+// Must be called under serviceReconcileLock so that
+// localServiceNameToServiceInfoMap reads are consistent.
+func (updater *loadBalancerBackendPoolUpdater) groupOperations(ops []batchOperation) map[string][]batchOperation {
+	logger := log.Background().WithName("loadBalancerBackendPoolUpdater.groupOperations")
```
nilo19: should be log.FromContextOrBackground(ctx)

Liunardy: Updated.

PR #10335, opened 2026-05-13
ci(update-trivy-db): declare contents: read + actions: write

arpitjain099:
Pins the default `GITHUB_TOKEN` on `update-trivy-db.yaml` to the minimum scope it actually uses:
- `actions: write` for `actions/cache/save@v5` (the Trivy DB save path).
- `contents: read` is included for consistency, though the workflow has no checkout step.

The `oras pull` of `ghcr.io/aquasecurity/trivy-db:2` is anonymous (public image). Hardens against the GitHub-side write surface.

YAML validated locally with `yaml.safe_load`.

### DIFF
--- a/.github/workflows/update-trivy-db.yaml
+++ b/.github/workflows/update-trivy-db.yaml
@@ -7,6 +7,12 @@ on:
     - cron: '0 0 * * *'  # Run daily at midnight UTC
   workflow_dispatch:  # Allow manual triggering
 
+# Pulls the Trivy DB from GHCR (anonymous) and saves it to the GHA cache;
+# no source checkout is needed, only actions: write for the cache save.
+permissions:
+  contents: read
+  actions: write
+
 jobs:
   update-trivy-db:
     runs-on: ubuntu-latest


nilo19:
why do we need this change as we don't hit any failures. Why do we need the actions: write?

arpitjain099:
Good questions, two parts:

1. **Why declare it at all** - it's defense in depth, not a current failure. Without a `permissions:` block the workflow inherits the repo-wide default (often "Read and write permissions"), so `GITHUB_TOKEN` ends up with write across most scopes. Pinning `contents: read` shrinks the blast radius if a third-party action ever gets compromised (cf. CVE-2025-30066 on `tj-actions/changed-files` earlier this year).

2. **Why `actions: write`** - `actions/cache/save@v5` uploads the Trivy DB to the GitHub Actions cache. Once we drop from the default "write-all" down to explicit `contents: read`, that save step would fail without `actions: write`. So the scope is the minimum the workflow actually exercises, not an addition.

Happy to split into two PRs if either part is contentious.

PR #10350, opened 2026-05-15
fix: update generate-release-note script

anndono:
#### What type of PR is this?

/kind bug
This PR updates generate-release-note.sh to use the current release-notes CLI interface. 

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/hack/generate-release-note.sh
+++ b/hack/generate-release-note.sh
@@ -18,9 +18,10 @@ OUTPUT=${2:-release-notes.md}
 UPDATE_SITE=${3:-false}
 
 install_cli() {
+  export PATH="$(go env GOPATH)/bin:${PATH}"
   if ! [[ -x "$(command -v release-notes)" ]]; then
     echo "CLI release-notes not found, installing..."
-    GO111MODULE=on go install k8s.io/release/cmd/release-notes@latest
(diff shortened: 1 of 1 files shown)


Liunardy (approved):
/lgtm

hack/generate-release-note.sh
```diff
@@ -18,9 +18,10 @@ OUTPUT=${2:-release-notes.md}
...
 
 install_cli() {
+  export PATH="$(go env GOPATH)/bin:${PATH}"
   if ! [[ -x "$(command -v release-notes)" ]]; then
     echo "CLI release-notes not found, installing..."
-    GO111MODULE=on go install k8s.io/release/cmd/release-notes@latest
+    GO111MODULE=on go install k8s.io/release/cmd/release-notes@v0.21.1
   else
     echo "CLI release-notes found, skip installing. If you want to upgrade, run 'GO111MODULE=on go install k8s.io/release/cmd/release-notes@latest'"
   fi
```
Liunardy, line 27: This only installs if the binary isn't already found on PATH. If a release-notes binary already exists (any version), it skips the install entirely and uses whatever version is present. Since version is pinned, we should always use the pinned version.

PR #10253, opened 2026-05-03
docs: add local service backend pool updater design

nilo19:
/kind design

#### What this PR does / why we need it:

Adds a design document for refining the local service backend pool updater retry and error handling behavior.

The design covers bounded updater-level retries, 429 throttling handling, 409/412 conflict retries, SDK retry boundaries, event semantics, queue merge behavior, shutdown behavior, metrics, and the expected unit test coverage.

#### Which issue(s) this PR fixes:

NONE

#### Special notes for your reviewer:

This is a design-only PR. It intentionally treats `retryrepectthrottled.GetRetriableStatusCode()` statuses as terminal at the updater layer because the Azure SDK retry policy already handles those statuses before the updater sees the error.

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

#### Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.:

```docs
Adds a design doc under content/en/development/design-docs/.
```

PR #10417, opened 2026-05-21
doc: add local service backend pool updater retry design

Liunardy:
#### What type of PR is this?
/kind design

Adds a design document for refining the local service backend pool updater retry and error handling behavior.

The design covers bounded updater-level retries, 429 throttling handling, 409/412 conflict retries, SDK retry boundaries, event semantics, queue merge behavior, shutdown behavior, metrics, and the expected unit test coverage.

Based on #10253 with refinements and aligned with #10328.

NONE

This is a design-only PR. It intentionally treats retryrepectthrottled.GetRetriableStatusCode() statuses as terminal at the updater layer because the Azure SDK retry policy already handles those statuses before the updater sees the error.

```release-note
NONE
```

```docs
Adds a design doc under content/en/development/design-docs/.
```


content/en/development/design-docs/refining-local-service-backend-pool-updater.md
```diff
@@ -0,0 +1,261 @@
...
+
+Use requeue-on-retriable-error. Each `loadBalancerBackendPoolUpdateOperation` carries retry metadata:
+
+- `retryCount int`, tracking how many retryable failures have already been consumed
+- `nextEligibleAt time.Time`, tracking when a throttled operation is eligible to be processed again
+
+On each updater tick:
+
+1. Acquire `serviceReconcileLock` to serialize with the main service reconciliation loop.
+2. If the queue is empty (`countOperations() == 0`), release the lock and return.
```
nilo19, line 73: swap 1 and 2?

content/en/development/design-docs/refining-local-service-backend-pool-updater.md
```diff
@@ -0,0 +1,261 @@
...
+3. Acquire `azureResourceLocker` (distributed lease lock) to serialize with other components that update load balancer resources. If the lease lock fails, return without draining. Operations remain in the queue for the next tick.
+4. Drain the queued operations via `drainOperations()` (acquires `updater.lock`, moves all operations out, clears the queue, releases `updater.lock`).
+5. Filter and group operations via `groupOperations()` under `serviceReconcileLock`:
+   - the Service is still in `localServiceNameToServiceInfoMap`
+   - the Service still points to the same load balancer
+   - group relevant operations by `loadBalancerName/backendPoolName`
+6. If any operation in a group has `nextEligibleAt` in the future, requeue the whole group. Do not call ARM, emit an event, record a metric, or increment retry state for skipped groups.
```
nilo19, line 80: worth experimenting in real cases if we can get nextEligibleAt header from the response.

content/en/development/design-docs/refining-local-service-backend-pool-updater.md
```diff
@@ -0,0 +1,261 @@
...
+- `stale`: drop quietly
+- `retriable`: emit retrying event and requeue while budget remains
+- `terminal`: emit failed event and drop
+
+Rules:
+
+- ARM resource-not-found errors as detected by `errutils.CheckResourceExistsFromAzcoreError` are `stale`.
+- ARM wire `429 TooManyRequests` responses are `retriable`.
+  - Classify them by checking `errors.As(err, &respErr)` for `*azcore.ResponseError` with `StatusCode == http.StatusTooManyRequests`.
+  - Extract `Retry-After` from `respErr.RawResponse.Header` when `RawResponse` is non-nil.
```
nilo19: I'm wondering if this can be extracted or not. It seems we converts 429 into retryrepectthrottled.ErrTooManyRequest while reading the header internally, and the backend-pool wrapper returns generated-client errors directly.
https://github.com/kubernetes-sigs/cloud-provider-azure/blob/master/pkg/azclient/policy/retryrepectthrottled/throttle.go#L74

content/en/development/design-docs/refining-local-service-backend-pool-updater.md
```diff
@@ -0,0 +1,261 @@
...
+9. Requeue operations whose retry budget remains (acquire `updater.lock` to append back to the live queue).
+10. Drop exhausted operations and emit `LoadBalancerBackendPoolUpdateFailed` once per distinct Service in the group.
+
+Retries do not carry over the in-memory backend pool object. The next tick re-issues `Get` for the group before re-applying queued operations, which is required for `CreateOrUpdate` conflicts and stale etags.
+
+The updater must re-acquire the lock only when appending retry operations back to the live queue. New EndpointSlice events can continue adding fresh operations while a previous snapshot is doing Azure calls.
+
+### Merge Rule
+
+Fresh operations and requeued operations for the same `loadBalancerName/backendPoolName` merge on the next updater tick through the existing grouping logic. Each operation keeps its own `retryCount` and `nextEligibleAt`; fresh operations start at `retryCount=0` and no `nextEligibleAt`.
```
nilo19, line 94: The requeued operation can be stale, and may overwrite desired operations, making the final state unwanted.
Example:
- start state:
Backend pool currently has: [A]
Service endpoints should be: [A]
- Then endpoint moves from node/IP A to B.
Desired endpoints: [B]
Queued operations:
1. remove A
2. add B
- Updater processes those operations, but CreateOrUpdate fails with a retriable error. Before the retry runs, endpoints change again from B to C.
New desired endpoints: [C]
New queued operations:
remove B
add C
- Now the queue may contain fresh operations first, then the old retry appended after them:
remove B
add C
remove A
add B   <- stale old operation
- On the next tick, updater does a fresh Get. Since the earlier write failed, Azure still has:
Backend pool: [A]
- Then it applies the queued operations in order:
remove B  -> [A]      // no-op, B is not there
add C     -> [A, C]
remove A  -> [C]
add B     -> [C, B]   // stale B comes back
- Final backend pool becomes:
[C, B]
- But the correct latest desired state is:
[C]

content/en/development/design-docs/refining-local-service-backend-pool-updater.md
```diff
@@ -0,0 +1,261 @@
...
+
+`addOperation` and `removeOperation` only acquire `updater.lock`. This prevents slow Azure calls from blocking EndpointSlice event handlers that need to enqueue newer operations.
+
+The `removeOperation(serviceName)` method cannot remove operations already in a processing snapshot. To avoid stale retry behavior, the processing path must re-check service relevance before requeueing and before sending retry/failure events.
+
+Parked operations waiting for `nextEligibleAt` live in `updater.operations`, so `removeOperation(serviceName)` can remove them while they are parked. If removal races with the short snapshot-to-requeue window, the next tick's relevance check drops the stale operation quietly.
+
+The relevance re-check depends on the Service cleanup path also removing the Service from `localServiceNameToServiceInfoMap`. The implementation must preserve that invariant: a deleted Service should either be removed from the live queue by `removeOperation(serviceName)` or be rejected by the map-based relevance check before a retry is requeued or reported.
+
+### Queue Growth During Throttling
```
nilo19: T1: endpoint A -> B
queue: remove A, add B
T2: retry is parked by nextEligibleAt
T3: endpoint B -> C
queue now: remove A, add B, remove B, add C
T4: endpoint C -> D
queue now: remove A, add B, remove B, add C, remove C, add D
Even though all operations target the same backend pool, the queue still grows with every EndpointSlice update. Grouping only means they may be processed together later. It does not prevent queue from being very long quickly.
A stronger model: coalescing operations per backend pool, which means after A -> B -> C -> D, the updater has one pending item: `lb1/pool1 should be [D]`.
Is it doable?

content/en/development/design-docs/refining-local-service-backend-pool-updater.md
```diff
@@ -0,0 +1,261 @@
...
+```
+
+Semantics:
+
+- `nil` uses the default value `3`.
+- Explicit `0` disables updater-level retry and preserves today's drop-after-failure behavior for retriable errors.
+- Negative values are normalized to `0`.
+- The count means extra retries after the first failed attempt. With the default `3`, a retriable error can be observed up to four times: one initial attempt plus three requeued retry attempts.
+
+Use a pointer so the config loader can distinguish "unset" from "explicitly set to 0". Normalization happens next to the existing `LoadBalancerBackendPoolUpdateIntervalInSeconds` defaulting in `pkg/provider/azure.go`: if the pointer is nil, use `DefaultLoadBalancerBackendPoolUpdateMaxRetries`; if the pointer is non-nil and negative, clamp it to `0`.
```
nilo19, line 59: nit: in pkg/provider/config/azure.go

content/en/development/design-docs/refining-local-service-backend-pool-updater.md
```diff
@@ -0,0 +1,261 @@
...
+
+For 429 throttling, `Retry-After` overrides the next normal updater tick by setting `nextEligibleAt`. Ticks before `nextEligibleAt` only preserve the operation in the queue after re-checking Service/LB relevance; they do not call ARM, emit retrying events, or consume retry budget. `LoadBalancerBackendPoolUpdateMaxRetries` bounds failed processing attempts, not elapsed wall-clock time, so a long `Retry-After` can delay final success or failure beyond the normal interval-based timing.
+
+During sustained Azure throttling, operators can reduce `LoadBalancerBackendPoolUpdateMaxRetries` or increase `LoadBalancerBackendPoolUpdateIntervalInSeconds` to avoid retry amplification when `Retry-After` is unavailable.
+
+Exponential backoff is not used because the retriable errors at this layer either use server-directed delays (429 with `Retry-After`) or benefit from a fresh `Get` rather than longer waits (409/412). When all retries are exhausted without success, the operator receives a `LoadBalancerBackendPoolUpdateFailed` event with guidance to retrigger (see the Event Behavior section).
+
+## Tests
```
nilo19, line 245: the design should keep explicit tests for “retry requeue before delete/migration is later canceled by removeOperation or map deletion,” because the safety depends on requeue happening while serviceReconcileLock is still held.

content/en/development/design-docs/refining-local-service-backend-pool-updater.md
Liunardy: With serviceReconcileLock held during the count check, we have a guarantee that the count can't change due to removeOperation() between the count check and the drain. removeOperation() callers (which hold serviceReconcileLock) need to cancel queued operations before the drain.

content/en/development/design-docs/refining-local-service-backend-pool-updater.md
Liunardy: Experimented with triggering real 429s. Added ThrottleError section with pipeline flow.

content/en/development/design-docs/refining-local-service-backend-pool-updater.md
Liunardy: Replaced *azcore.ResponseError approach with ThrottleError type that enriches ErrTooManyRequest. See ThrottleError section.

content/en/development/design-docs/refining-local-service-backend-pool-updater.md
Liunardy: Added prepend logic (step 9) in Retry Model. This preserves chronological diff ordering so retried ops execute before fresh ops. For the retry exhaustion case, the stale state limitation is the same as today's no-retry model. A desired-state model is documented as a follow-up in Queue Growth and Stale Operations.

content/en/development/design-docs/refining-local-service-backend-pool-updater.md
Liunardy: Documented as a follow-up in Queue Growth and Stale Operations.

content/en/development/design-docs/refining-local-service-backend-pool-updater.md
Liunardy: The config field definition is in pkg/provider/config/azure.go (added in the design now). The normalization/defaulting happens in pkg/provider/azure.go in checkEnableMultipleStandardLoadBalancers(), next to the existing LoadBalancerBackendPoolUpdateIntervalInSeconds defaulting.

content/en/development/design-docs/refining-local-service-backend-pool-updater.md
Liunardy: Added.

content/en/development/design-docs/refining-local-service-backend-pool-updater.md
```diff
@@ -0,0 +1,287 @@
...
+
+### ThrottleError
+
+The `ThrottlingPolicy` (a per-retry policy in the SDK pipeline) intercepts 429 responses, reads the `Retry-After` header, sets an internal gate timer, and returns `ErrTooManyRequest`. The SDK pipeline executes per-retry policies inside the retry loop. When ARM returns 429, `ThrottlingPolicy` sets the gate and returns the error. The SDK retry loop makes up to 4 attempts (1 initial + `MaxRetries` 3). If the gate timer has not expired, subsequent retries are blocked by the local gate without reaching ARM. After all attempts are exhausted, the error passes through the generated client to the updater.
+
+Today `ErrTooManyRequest` is a plain sentinel (`errors.New(...)`) with no metadata. The updater cannot extract the `Retry-After` duration from it.
+
+Without enrichment, a long `Retry-After` can exhaust the updater's retry budget against the local gate without ever making a real ARM attempt after the throttle window.
+
+Enrich the existing sentinel with a `ThrottleError` type in `pkg/azclient/policy/retryrepectthrottled/throttle.go`:
```
nilo19, line 131: In this way we will have to bump azclient for all release branches, and the k8s.io/xx dependencies minor versions will have to be updated as well. Do we have a plan B, e.g., not supporting throttling retry?

Liunardy: As discussed, we'll proceed with the ThrottleError approach (azclient bump) for now.

Alternatives considered:
- Fall back to next tick (no azclient change): Each retry tick makes an SDK call that hits the ThrottlingPolicy gate unnecessarily.
- Treat 429 as terminal (no azclient change): Consistent with how SDK-retriable statuses (408, 500, 502, 503, 504) are already treated. No retry for throttling conditions.

If azclient bump becomes an issue, we can fall back to treating 429 as terminal.

content/en/development/design-docs/refining-local-service-backend-pool-updater.md
```diff
@@ -0,0 +1,287 @@
...
+8. A fresh operation merged with a requeued operation keeps its own retry counter; on group failure, all operations in the group consume one retry.
+9. A group with mixed retry budgets (one operation at max retries, one fresh): the exhausted operation is dropped and reported while the fresh operation is requeued.
+10. Retry budget exhaustion emits `LoadBalancerBackendPoolUpdateFailed` with guidance text and leaves the queue empty.
+11. Statuses from `retryrepectthrottled.GetRetriableStatusCode()` do not requeue and emit `LoadBalancerBackendPoolUpdateFailed`.
+12. Other non-retriable errors do not requeue and emit `LoadBalancerBackendPoolUpdateFailed`.
+13. ARM resource-not-found does not requeue and emits no event.
+14. Stale service or changed load balancer before requeue is dropped quietly.
+15. Retry requeue is serialized with service deletion under `serviceReconcileLock`: a retriable failure requeues the operation before the main reconcile path can call `removeOperation()`. After the lock is released, `removeOperation()` removes the requeued operation from the queue.
```
nilo19: That's for LB migration path, where removeOperation(svcName) is called. It is not true for the delete path: EnsureLoadBalancerDeleted deletes localServiceNameToServiceInfoMap but does not call removeOperation(). I think we should split this into two tests: migration removes queued retry ops via removeOperation(), deletion leaves them queued and they are dropped on the next groupOperations() relevance check.

Liunardy: Split this.

content/en/development/design-docs/refining-local-service-backend-pool-updater.md
```diff
@@ -0,0 +1,288 @@
...
+
+Rules:
+
+- ARM resource-not-found errors as detected by `errutils.CheckResourceExistsFromAzcoreError` are `stale`.
+- `429 TooManyRequests` is `retriable`. The error reaches the updater through two paths: ARM 429 responses intercepted by `ThrottlingPolicy`, or local gate blocks when the throttle timer is still active. Both paths return `ThrottleError` with `RetryAfter` (see ThrottleError below). If `RetryAfter` is in the future, set `nextEligibleAt` on the requeued operation. Otherwise, fall back to the next updater tick.
+- ARM `409 Conflict` and `412 PreconditionFailed` responses are `retriable`; no special payload handling is needed because the next updater tick already runs through a fresh `Get` before the next `CreateOrUpdate`.
+- ARM response statuses in `retryrepectthrottled.GetRetriableStatusCode()` are `terminal` for this updater. The Azure SDK already retries those statuses inside the individual ARM call, so if the updater sees one of those errors, the SDK retry budget has already been exhausted. The updater emits `LoadBalancerBackendPoolUpdateFailed` and drops the operation instead of adding a second retry layer for the same condition.
```
nilo19: We have an assumption here that azclient will retry those with GetRetriableStatusCode(). But if it is configured to disable retry, e.g., CloudProviderBackoffRetries = 0, we will have no retry anywhere. Better to call this out explicitly.

content/en/development/design-docs/refining-local-service-backend-pool-updater.md
```diff
@@ -0,0 +1,288 @@
...
+
+Parked operations waiting for `nextEligibleAt` live in `updater.operations`, so `removeOperation(serviceName)` can remove them while they are parked. If removal races with the short snapshot-to-requeue window, the next tick's relevance check drops the stale operation quietly.
+
+The relevance re-check depends on the Service cleanup path also removing the Service from `localServiceNameToServiceInfoMap`. The implementation must preserve that invariant: a deleted Service should either be removed from the live queue by `removeOperation(serviceName)` or be rejected by the map-based relevance check before a retry is requeued or reported.
+
+### Queue Growth And Stale Operations
+
+The current operation model is diff-based: each EndpointSlice event enqueues add/remove operations for the IPs that changed. This has two limitations when combined with retry:
+
+**Queue growth**: during a long `Retry-After` period, EndpointSlice events continue to arrive and `addOperation()` appends new operations to the queue. The queue grows with each EndpointSlice event.
```
nilo19, line 211: Can we have a follow-up github issue and link to the existing one?

content/en/development/design-docs/refining-local-service-backend-pool-updater.md
```diff
@@ -0,0 +1,288 @@
...
+## Scope
+
+In scope:
+
+- Retry failed local-service backend-pool update operations when the error is classified as retriable.
+- Add retry state to queued backend-pool update operations.
+- Add a new max-retries configuration field for this updater.
+- Emit retry and failure events with bounded event volume.
+- Preserve current stale `404` event behavior. Stop recording `404` as a failure metric.
+- Enrich `ErrTooManyRequest` with a `ThrottleError` type to expose `Retry-After` duration to callers. The `ThrottleError` change in `pkg/azclient/policy/retryrepectthrottled/` requires a separate `azclient` release and dependency bump before the retry implementation can use it.
```
nilo19, line 26: Typo retryrepectthrottled?

Liunardy: It's the actual package name https://github.com/kubernetes-sigs/cloud-provider-azure/blob/3f57aa450c856b9c5dbf0e2b5863fc220281602c/pkg/azclient/policy/retryrepectthrottled/throttle.go#L17

Should we fix this?

nilo19: No for this change, could add an issue to track.

content/en/development/design-docs/refining-local-service-backend-pool-updater.md
Liunardy: CloudProviderBackoffRetries = 0 does not disable SDK retries. 

The override only applies when CloudProviderBackoff && CloudProviderBackoffRetries > 0, so 0 or negative values are ignored and the default MaxRetries (3) stays in effect. 
https://github.com/kubernetes-sigs/cloud-provider-azure/blob/3f57aa450c856b9c5dbf0e2b5863fc220281602c/pkg/azclient/arm_conf.go#L84-L85

The SDK also treats MaxRetries = 0 as unset in its own setDefaults().
https://github.com/kubernetes-sigs/cloud-provider-azure/blob/3f57aa450c856b9c5dbf0e2b5863fc220281602c/vendor/github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime/policy_retry.go#L22-L31

Will add a note for this.

content/en/development/design-docs/refining-local-service-backend-pool-updater.md
Liunardy: Created https://github.com/kubernetes-sigs/cloud-provider-azure/issues/10556

content/en/development/design-docs/refining-local-service-backend-pool-updater.md
Liunardy: Added this note.

content/en/development/design-docs/refining-local-service-backend-pool-updater.md
Liunardy: Created https://github.com/kubernetes-sigs/cloud-provider-azure/issues/10558

content/en/development/design-docs/refining-local-service-backend-pool-updater.md
```diff
@@ -0,0 +1,288 @@
...
+
+The throttling delay is group-level. If one operation for `lbName/backendPoolName` is parked by `nextEligibleAt`, all operations in that same group are preserved until the group becomes eligible. This avoids splitting add/remove operations for the same backend pool and preserves batching order. Unrelated groups continue processing normally.
+
+Retried operations are prepended to `updater.operations` so they appear before fresh operations added during processing. This preserves chronological diff ordering: old transitions are applied first, then newer ones, producing the correct final state. Without this, stale retried operations appended after fresh operations can overwrite the desired state.
```
nilo19, line 100: the test should specifically cover: parked same-pool old ops + fresh same-pool ops added before requeue.

content/en/development/design-docs/refining-local-service-backend-pool-updater.md
```diff
@@ -0,0 +1,288 @@
...
+
+### Queue Growth And Stale Operations
+
+The current operation model is diff-based: each EndpointSlice event enqueues add/remove operations for the IPs that changed. This has two limitations when combined with retry:
+
+**Queue growth**: during a long `Retry-After` period, EndpointSlice events continue to arrive and `addOperation()` appends new operations to the queue. The queue grows with each EndpointSlice event.
+
+**Stale state after retry exhaustion**: the informer computes diffs between old and new EndpointSlice states. If a prior write failed and its operations were dropped, the changes from that write are lost. Subsequent diffs only capture newer EndpointSlice changes and do not replicate the lost operations. This is the same behavior as today's no-retry model. The main reconciliation loop corrects the drift on the next Service reconciliation triggered by `EnsureLoadBalancer`.
```
nilo19, line 213: The event message should clearly state the reason why it failed and why retry can't be helpful. We should also enrich our document to state those limitations as well. Better to add a link to that doc from the comment of our code somewhere.

content/en/development/design-docs/refining-local-service-backend-pool-updater.md
Liunardy: Added.

content/en/development/design-docs/refining-local-service-backend-pool-updater.md
Liunardy: Differentiated the event message for non-retriable and retriable errors.
Added multi-slb doc update in scope and for the code to link to this.

PR #10443, opened 2026-05-26
chore: pin release-notes cli in generate release note script to v0.21.1

anndono:
#### What type of PR is this?

/kind cleanup
Pin the `release-notes` CLI version used by `hack/generate-release-note.sh`.

Previously the script only installed `release-notes` when the binary was not already on `PATH`, so any pre-existing version would be reused, making release note generation non-deterministic across machines and CI runners.

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/hack/generate-release-note.sh
+++ b/hack/generate-release-note.sh
@@ -25,7 +25,10 @@ install_cli() {
      ! go version -m "$(command -v release-notes)" 2>/dev/null | \
        grep -qE "^[[:space:]]+mod[[:space:]]+k8s.io/release[[:space:]]+${RELEASE_NOTES_VERSION}([[:space:]]|$)"; then
     echo "CLI release-notes ${RELEASE_NOTES_VERSION} not found, installing..."
-    GO111MODULE=on go install "k8s.io/release/cmd/release-notes@${RELEASE_NOTES_VERSION}"
+    if ! GO111MODULE=on go install "k8s.io/release/cmd/release-notes@${RELEASE_NOTES_VERSION}"; then
+      echo "ERROR: failed to install release-notes ${RELEASE_NOTES_VERSION}" >&2
+      exit 1
+    fi
   else
     echo "CLI release-notes ${RELEASE_NOTES_VERSION} found, skip installing."
   fi
--- a/hack/generate-release-note.sh
+++ b/hack/generate-release-note.sh
@@ -20,12 +20,13 @@ UPDATE_SITE=${3:-false}
 RELEASE_NOTES_VERSION=v0.21.1
 
 install_cli() {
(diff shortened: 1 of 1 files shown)


hack/generate-release-note.sh
```diff
@@ -17,13 +17,17 @@ RELEASE=$1
...
 install_cli() {
   export PATH="$(go env GOPATH)/bin:${PATH}"
-  if ! [[ -x "$(command -v release-notes)" ]]; then
-    echo "CLI release-notes not found, installing..."
-    GO111MODULE=on go install k8s.io/release/cmd/release-notes@v0.21.1
+  if ! [[ -x "$(command -v release-notes)" ]] || \
+     ! go version -m "$(command -v release-notes)" 2>/dev/null | \
+       grep -qE "^[[:space:]]+mod[[:space:]]+k8s.io/release[[:space:]]+${RELEASE_NOTES_VERSION}([[:space:]]|$)"; then
+    echo "CLI release-notes ${RELEASE_NOTES_VERSION} not found, installing..."
+    GO111MODULE=on go install "k8s.io/release/cmd/release-notes@${RELEASE_NOTES_VERSION}"
```
nilo19: If the go install failed and a stale release-notes exists, it may still use the stale version. set -e, or go install xxx || exit 1, or install_cli || exit 1.

anndono: Added error handling after go install

hack/generate-release-note.sh
```diff
@@ -17,13 +17,20 @@ RELEASE=$1
...
 install_cli() {
   export PATH="$(go env GOPATH)/bin:${PATH}"
-  if ! [[ -x "$(command -v release-notes)" ]]; then
-    echo "CLI release-notes not found, installing..."
-    GO111MODULE=on go install k8s.io/release/cmd/release-notes@v0.21.1
+  if ! [[ -x "$(command -v release-notes)" ]] || \
+     ! go version -m "$(command -v release-notes)" 2>/dev/null | \
+       grep -qE "^[[:space:]]+mod[[:space:]]+k8s.io/release[[:space:]]+${RELEASE_NOTES_VERSION}([[:space:]]|$)"; then
+    echo "CLI release-notes ${RELEASE_NOTES_VERSION} not found, installing..."
+    if ! GO111MODULE=on go install "k8s.io/release/cmd/release-notes@${RELEASE_NOTES_VERSION}"; then
```
nilo19: not generated by this change, but if GOBIN is set elsewhere in the env, we could still use the stale version inside GOBIN, because the new version is installed in go env GOPATH/bin. We may set GOBIN=go env GOPATH/bin when performing go install.

nilo19:
ask copilot to not include itself into the commit message.

Issue #10083, opened 2026-03-23
Feature Request: Reduce NSG Destination Prefix Growth When Floating IP Is Disabled


craigktreasure:
## Summary

When `service.beta.kubernetes.io/azure-disable-load-balancer-floating-ip: "true"` is set on a `LoadBalancer` Service, the Azure cloud provider uses backend node IPs as NSG rule destinations instead of the load balancer frontend IPs. In larger clusters, this causes NSG destination prefix usage to grow with node count and can exhaust Azure NSG prefix limits.

We would like to request support for a more scalable destination model when floating IP is disabled, such as allowing destination CIDRs/prefixes or allowing operators to opt out of ACP-managed NSG rules for a Service.

## Current Behavior

The controller switches destination handling based on `azure-disable-load-balancer-floating-ip` in [pkg/provider/azure_loadbalancer.go](https://github.com/kubernetes-sigs/cloud-provider-azure/blob/master/pkg/provider/azure_loadbalancer.go#L3185-L3223).

When floating IP is disabled, `reconcileSecurityGroup()` uses backend node IPs:

```go
if disableFloatingIP {
    dstIPv4Addresses = append(dstIPv4Addresses, backendIPv4Addresses...)
    dstIPv6Addresses = append(dstIPv6Addresses, backendIPv6Addresses...)
} else {
    dstIPv4Addresses = append(dstIPv4Addresses, lbIPv4Addresses...)
    dstIPv6Addresses = append(dstIPv6Addresses, lbIPv6Addresses...)
}
```

The destination port also changes from `port.Port` to `port.NodePort` in [pkg/provider/loadbalancer/accesscontrol.go](https://github.com/kubernetes-sigs/cloud-provider-azure/blob/master/pkg/provider/loadbalancer/accesscontrol.go#L361-L373):

```go
if consts.IsK8sServiceDisableLoadBalancerFloatingIP(svc) {
    p = port.NodePort
} else {
    p = port.Port
}
```

Those destination addresses are then written into NSG rules via [pkg/provider/securitygroup/securitygroup.go](https://github.com/kubernetes-sigs/cloud-provider-azure/blob/master/pkg/provider/securitygroup/securitygroup.go#L166-L196) and [pkg/provider/securitygroup/securityrule.go](https://github.com/kubernetes-sigs/cloud-provider-azure/blob/master/pkg/provider/securitygroup/securityrule.go#L125-L133).

## Why This Becomes a Scaling Problem

[...]

PR #10475, opened 2026-06-04
Add service opt-out for load balancer NSG rules

craigktreasure:
Refs kubernetes-sigs/cloud-provider-azure#10083.

This PR implements the first, minimal option from the issue discussion: a per-Service break-glass annotation that lets operators opt out of cloud-provider-managed LoadBalancer NSG rule creation for Services where those rules need to be managed externally.

The new annotation is intentionally narrow in scope. When it is set, cloud-provider-azure still computes the Service's NSG destinations and cleans stale ACP-managed rules for that Service, but it does not recreate the managed allow/deny rules. Load balancer rules, frontend IPs, backend pool membership, probes, and backend selection are unchanged.

This is not intended to be the preferred long-term UX for the destination-prefix scaling problem described in the issue. It is one of the proposed options, and provides a small, explicit escape hatch while preserving ACP cleanup behavior. A follow-up option can still explore allowing operators to provide destination prefixes while keeping ACP as the NSG rule owner.

Operators using this annotation are responsible for ensuring equivalent NSG access is provided outside ACP-managed rules. Manual rules should stay outside ACP's managed rule naming and priority conventions.

### DIFF
--- a/internal/testutil/fixture/kubernetes.go
+++ b/internal/testutil/fixture/kubernetes.go
@@ -106,6 +106,11 @@ func (f *KubernetesServiceFixture) WithDisableFloatingIP() *KubernetesServiceFix
 	return f
 }
 
+func (f *KubernetesServiceFixture) WithDisableLoadBalancerSecurityGroupRules() *KubernetesServiceFixture {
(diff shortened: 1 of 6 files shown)


craigktreasure:
Superseded by #10508.

PR #10507, opened 2026-06-11
Avoid listPIP for non-public pinned LB IPs

### LINKED ISSUE #10226: Reduce listPIP 429
## What happened

`cloud-controller-manager` repeatedly logs `serviceOwnsFrontendIP` warnings while reconciling an internal LoadBalancer Service with a private pinned IP:

```text
serviceOwnsFrontendIP: unexpected error when finding match public IP of the service ingress-nginx-controller with loadBalancerIP 10.104.176.35:
findMatchedPIPByLoadBalancerIP: failed to listPIP force refresh: throttled due to too many requests
```

The Service is internal and uses:

```yaml
service.beta.kubernetes.io/azure-load-balancer-internal: "true"
service.beta.kubernetes.io/azure-load-balancer-ipv4: "10.104.176.35"
```

During reconcile/cleanup, the ownership check can enter the external/public-PIP branch with this private IP and try to find a Public IP resource whose address is `10.104.176.35`. That lookup is a guaranteed miss, and under ARM throttling it can amplify repeated `listPIP` / force-refresh calls.

## Symptoms

- Repeated `serviceOwnsFrontendIP` warnings for the same Service/IP.
- Errors alternate between:
  - `cannot find public IP with IP address <private-ip>`
  - `failed to listPIP`
  - `failed to listPIP force refresh`
- Reconcile latency increases while ARM is throttling.
- Logs become noisy around node churn and Service backend syncs.

## Suspected code path

Relevant paths:

- `pkg/provider/azure_loadbalancer.go`
  - `reconcileService`
  - `reconcileLoadBalancer`
  - `getServiceLoadBalancerStatus`
  - `reconcileFrontendIPConfigs`
  - `isFrontendIPChanged`
  - `serviceOwnsFrontendIP`
- `pkg/provider/azure_publicip_repo.go`
  - `findMatchedPIP`
  - `findMatchedPIPByLoadBalancerIP`
  - `listPIP`
- `pkg/cache/azure_cache.go`
  - `TimedCache.get`

Observed flow:

```text
Service reconcile reaches frontend ownership checks
-> serviceOwnsFrontendIP sees loadBalancerIP="10.104.176.35"
-> external/public-PIP branch calls findMatchedPIP
-> listPIP(default)
-> on miss, listPIP(force refresh)
-> 429 or guaranteed not found
```

## Why this is expensive

`serviceOwnsFrontendIP` is used like a cheap boolean predicate, but for external secondary Services it can trigger ARM-backed Public IP discovery.

One reconcile may call it multiple times:

- while scanning frontend IP configs for status
- while reconciling frontend IP configs
- while checking whether frontend IP changed
- while deriving owned frontend config IP version

When lookup fails, `serviceOwnsFrontendIP` logs and returns `false`. The error is not propagated as a reusable throttling/lookup state, so callers may continue scanning and re-enter the same PIP lookup.

`findMatchedPIPByLoadBalancerIP` force-refreshes the PIP cache on every miss. The cache does not memoize negative results or 429 failures, so repeated callers can repeatedly hit ARM.

## Expected behavior

For a private `loadBalancerIP`, the public-PIP ownership path should avoid repeated Public IP list/force-refresh calls.

During ARM throttling, repeated ownership checks should avoid re-triggering identical PIP lookups within the same reconcile.

## Proposed improvements

- Short-circuit impossible public-PIP lookups.

  If `serviceOwnsFrontendIP` is on the external/public-PIP branch and `loadBalancerIP` is private/RFC1918, skip `findMatchedPIP` during ownership scanning.

  For real external create/update paths where a user specifies a private IP, return a clear validation error instead of repeatedly listing PIPs.

- Add per-reconcile memoization for PIP lookup.

  Memoize PIP lookup results by:

  ```text
  (resourceGroup, loadBalancerIP or pipName)
  ```

  within one reconcile. Cache both success and failure/throttle state for the reconcile duration.

- Stop flattening lookup errors into `false`.

  Change ownership resolution to distinguish:

  ```text
  owned=false, err=nil        // definitely not owned
  owned=false, err=throttled  // lookup unknown due to throttling
  owned=false, err=notFound   // lookup completed, no match
  ```

  Then the top-level reconcile can log once and back off instead of continuing as if the frontend simply belongs to another Service.

- Reduce NSG retain log amplification.

  `listAvailableSecurityGroupDestinations` and `RetainDestinationFromRules` can log very large duplicated destination lists. De-duping retained destinations before retain/logging would reduce noise and CPU overhead, though it does not directly fix the PIP 429 loop.

## Acceptance criteria

- Private `loadBalancerIP` values do not trigger Public IP list/force-refresh calls from ownership scanning.
- Repeated `serviceOwnsFrontendIP` checks in one reconcile do not repeat identical PIP list calls after a miss or 429.
- 429 from PIP list is propagated or memoized as an unknown/throttled lookup state, not silently converted to `false`.
- Unit tests cover:
  - public-PIP ownership scan with private `loadBalancerIP` skips `findMatchedPIP`
  - external Service with private pinned IP fails/skips early as designed
  - repeated ownership checks reuse memoized PIP lookup result
  - throttled PIP lookup is not retried multiple times in one reconcile

nilo19:
#### What type of PR is this?

/kind bug

This PR rejects invalid or non-public pinned LoadBalancer IPs before `findMatchedPIP` lists Azure Public IPs.

For external Services with a pinned LoadBalancer IP that cannot identify an Azure Public IP, the provider now returns a clear validation error instead of doing a guaranteed-miss Public IP list. The validation covers non-global-unicast addresses plus Azure-documented private/reserved VNet ranges, including RFC1918, RFC6598 `100.64.0.0/10`, Azure internal DNS `168.63.129.16/32`, and IPv6 ULA.

It also moves the pinned LoadBalancer IP validation errors into `pkg/provider/errors` so the clean user-facing Service event message can be preserved while callers can still classify the root cause with `errors.Is`.

Partially addresses #10226

This intentionally does not close #10226. It addresses the non-public pinned IP path and error clarity, but leaves per-reconcile memoization and 429-state propagation for a follow-up.

Validation performed:
- `go test ./pkg/provider/... -count=1`
- From `tests/`: `go test ./e2e/network -run '^$'`
- Manual standalone replay confirmed the validation event and no ingress assignment.

```release-note
External LoadBalancer Services with invalid or non-public pinned LoadBalancer IPs now fail fast with a validation error instead of listing Azure Public IPs.
```

```docs
- [Other doc]: https://learn.microsoft.com/azure/virtual-network/virtual-networks-faq#what-address-ranges-can-i-use-in-my-virtual-networks
```

### DIFF
--- a/pkg/provider/azure_loadbalancer.go
+++ b/pkg/provider/azure_loadbalancer.go
@@ -47,6 +47,7 @@ import (
 	"sigs.k8s.io/cloud-provider-azure/pkg/log"
 	"sigs.k8s.io/cloud-provider-azure/pkg/metrics"
 	"sigs.k8s.io/cloud-provider-azure/pkg/provider/config"
+	providererrors "sigs.k8s.io/cloud-provider-azure/pkg/provider/errors"
 	"sigs.k8s.io/cloud-provider-azure/pkg/provider/loadbalancer"
 	"sigs.k8s.io/cloud-provider-azure/pkg/trace"
 	"sigs.k8s.io/cloud-provider-azure/pkg/trace/attributes"
@@ -1190,11 +1191,11 @@ func (az *Cloud) determinePublicIPName(ctx context.Context, clusterName string, 
 		return pipName, false, err
 	}
 
-	// For the services with loadBalancerIP set, an existing public IP is required, primary
-	// or secondary, or a public IP not found error would be reported.
+	// For services with loadBalancerIP set, validate the IP and require an existing
+	// public IP, primary or secondary.
 	pip, err := az.findMatchedPIP(ctx, loadBalancerIP, "", pipResourceGroup)
 	if err != nil {
-		return "", false, err
+		return "", false, providererrors.NewExternalServiceLoadBalancerIPError(getServiceName(service), loadBalancerIP, err)
 	}
 
 	if pip != nil && pip.Name != nil {
@@ -4549,6 +4550,9 @@ func (az *Cloud) getLoadBalancerNameByPublicIP(
 		}
 		pip, err := az.findMatchedPIP(ctx, ip, "", pipResourceGroup)
 		if err != nil {
+			if providererrors.IsLoadBalancerIPValidationError(err) {
+				return "", fmt.Errorf("find public IP by address %q: %w", ip, providererrors.NewExternalServiceLoadBalancerIPError(getServiceName(service), ip, err))
+			}
 			return "", fmt.Errorf("find public IP by address %q: %w", ip, err)
 		}
 		if lbName := az.getLoadBalancerNameFromPIP(pip, clusterName); lbName != "" {
@@ -4889,6 +4893,9 @@ func (az *Cloud) serviceOwnsFrontendIP(ctx context.Context, fip *armnetwork.Fron
 		for _, loadBalancerIP := range loadBalancerIPs {
 			pip, err := az.findMatchedPIP(ctx, loadBalancerIP, "", pipResourceGroup)
 			if err != nil {
+				if providererrors.IsLoadBalancerIPValidationError(err) {
+					return false, isPrimaryService, nil
+				}
 				klog.Warningf("serviceOwnsFrontendIP: unexpected error when finding match public IP of the service %s with loadBalancerIP %s: %v", service.Name, loadBalancerIP, err)
 				return false, isPrimaryService, nil
 			}
--- a/pkg/provider/azure_publicip_repo.go
+++ b/pkg/provider/azure_publicip_repo.go
@@ -22,6 +22,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"net/netip"
 	"regexp"
 	"strings"
 	"sync"
@@ -36,7 +37,15 @@ import (
 	azcache "sigs.k8s.io/cloud-provider-azure/pkg/cache"
 	"sigs.k8s.io/cloud-provider-azure/pkg/consts"
 	"sigs.k8s.io/cloud-provider-azure/pkg/log"
+	providererrors "sigs.k8s.io/cloud-provider-azure/pkg/provider/errors"
 	"sigs.k8s.io/cloud-provider-azure/pkg/util/deepcopy"
+)
+
+var (
+	azureReservedIPPrefixes = []netip.Prefix{
+		netip.MustParsePrefix("100.64.0.0/10"),
+		netip.MustParsePrefix("168.63.129.16/32"),
+	}
 )
 
 // CreateOrUpdatePIP invokes az.NetworkClientFactory.GetPublicIPAddressClient().CreateOrUpdate with exponential backoff retry
@@ -168,9 +177,15 @@ func (az *Cloud) listPIP(ctx context.Context, pipResourceGroup string, crt azcac
 }
 
 func (az *Cloud) findMatchedPIP(ctx context.Context, loadBalancerIP, pipName, pipResourceGroup string) (pip *armnetwork.PublicIPAddress, err error) {
+	if loadBalancerIP != "" {
+		if err := validatePublicIPCandidate(loadBalancerIP); err != nil {
+			return nil, err
+		}
+	}
+
 	pips, err := az.listPIP(ctx, pipResourceGroup, azcache.CacheReadTypeDefault)
 	if err != nil {
-		return nil, fmt.Errorf("findMatchedPIPByLoadBalancerIP: failed to listPIP: %w", err)
+		return nil, fmt.Errorf("findMatchedPIP: failed to listPIP: %w", err)
 	}
 
 	if loadBalancerIP != "" {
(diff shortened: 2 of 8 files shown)


pkg/provider/azure_publicip_repo.go
```diff
@@ -220,13 +239,36 @@ func (az *Cloud) findMatchedPIPByLoadBalancerIP(ctx context.Context, pips []*arm
...
+func validatePublicIPCandidate(ip string) error {
+	addr, err := netip.ParseAddr(ip)
+	if err != nil {
+		return providererrors.NewInvalidPinnedLoadBalancerIP(ip)
+	}
+	if !addr.IsGlobalUnicast() || isAzurePrivateIP(addr) {
+		return providererrors.NewNonPublicPinnedLoadBalancerIP(ip)
+	}
+	return nil
+}
```
Liunardy, line 255: Possible IPv4-in-IPv6 bypass, "::ffff:a00:5" is private IPv4-mapped IPv6 address. Probably can use netip.Addr.IsPrivate() instead of using azurePrivateIPPrefixes.

pkg/provider/azure_publicip_repo.go
```diff
@@ -220,13 +239,36 @@ func (az *Cloud) findMatchedPIPByLoadBalancerIP(ctx context.Context, pips []*arm
...
+
+// isAzurePrivateIP reports whether addr belongs to a private or reserved Azure
+// Virtual Network address range:
+// https://learn.microsoft.com/azure/virtual-network/virtual-networks-faq#what-address-ranges-can-i-use-in-my-virtual-networks
+func isAzurePrivateIP(addr netip.Addr) bool {
+	for _, prefix := range azurePrivateIPPrefixes {
```
Liunardy: The names don't really match behaviour since it covers reserved Azure address range as well. Also see if netip.Addr.IsPrivate() can be used instead, so these can become something like isAzureReservedIP.

pkg/provider/errors/pinned_loadbalancer_ip.go
```diff
@@ -0,0 +1,74 @@
...
+	return fmt.Errorf("%w %q", ErrInvalidPinnedLoadBalancerIP, ip)
+}
+
+// NewNonPublicPinnedLoadBalancerIP returns an error for a pinned LoadBalancer IP that is not a public IP candidate.
+func NewNonPublicPinnedLoadBalancerIP(ip string) error {
+	return fmt.Errorf("%w %q", ErrNonPublicPinnedLoadBalancerIP, ip)
+}
+
+// IsPinnedLoadBalancerIP reports whether err is a known pinned LoadBalancer IP validation error.
+func IsPinnedLoadBalancerIP(err error) bool {
```
Liunardy: Nit: the constructors and bool check are New…IP/Is...IP but they return errors. How about matching with the NewExternalServicePinnedLoadBalancerIPError so they end with Error suffix?

pkg/provider/errors/pinned_loadbalancer_ip_test.go
```diff
@@ -0,0 +1,71 @@
...
+	} {
+		t.Run(tc.description, func(t *testing.T) {
+			assert.ErrorIs(t, tc.err, tc.target)
+			assert.EqualError(t, tc.err, tc.expected)
+			assert.True(t, providererrors.IsPinnedLoadBalancerIP(tc.err))
+		})
+	}
+}
+
+func TestNewExternalServicePinnedLoadBalancerIPError(t *testing.T) {
```
Liunardy: Nit: ErrInvalidPinnedLoadBalancerIP is not tested for NewExternalServicePinnedLoadBalancerIPError. Probably can use table-driven for all 3 paths.

pkg/provider/errors/pinned_loadbalancer_ip.go
```diff
@@ -0,0 +1,74 @@
...
+package providererrors
+
+import (
+	"errors"
+	"fmt"
+)
+
+var (
+	// ErrInvalidPinnedLoadBalancerIP identifies a pinned LoadBalancer IP that is not a valid IP address.
+	ErrInvalidPinnedLoadBalancerIP = errors.New("invalid pinned LoadBalancer IP")
```
Liunardy: "pinned LoadBalancer IP" is "pinned" actually meaningful, or can we drop it?

pkg/provider/azure_loadbalancer.go
```diff
@@ -1194,7 +1195,7 @@ func (az *Cloud) determinePublicIPName(ctx context.Context, clusterName string,
 	// or secondary, or a public IP not found error would be reported.
```
anndono: nit: It looks like this comment may be slightly out of sync as this path can return errors other than “public IP not found.”

pkg/provider/azure_publicip_repo_test.go
```diff
@@ -330,6 +333,114 @@ func TestFindMatchedPIP(t *testing.T) {
...
+		{
+			description:    "IPv4-mapped Azure shared address space load balancer IP",
+			loadBalancerIP: "::ffff:100.100.0.202",
+			expectedErr:    providererrors.ErrNonPublicLoadBalancerIP,
+		},
+		{
+			description:    "IPv4-mapped Azure internal DNS load balancer IP",
+			loadBalancerIP: "::ffff:168.63.129.16",
+			expectedErr:    providererrors.ErrNonPublicLoadBalancerIP,
+		},
```
Liunardy, line 394: Add cases for RFC1918 172.16/12, RFC1918 192.168/16, and IPv6 unique local address.

nilo19: Done

pkg/provider/azure_publicip_repo.go
nilo19: Use IsPrivate now.

PR #10508, opened 2026-06-11
Add LoadBalancer NSG rule opt-out annotation

nilo19:
#### What type of PR is this?

/kind feature

Adds `service.beta.kubernetes.io/azure-disable-load-balancer-nsg-rule` as a per-Service opt-out from cloud-provider-managed LoadBalancer NSG rules.

When the annotation is set to `true`, NSG reconciliation still cleans stale cloud-provider-managed rules for the Service and retains managed destinations, but skips patching new allow rules. This lets operators intentionally manage the Service's NSG rules outside cloud-provider-azure.

The annotation is intentionally generic for LoadBalancer Services and is not coupled to `service.beta.kubernetes.io/azure-disable-load-balancer-floating-ip`. It is documented in code as not recommended for general use and intended as a break-glass option where operators own the NSG rule lifecycle.

Partially #10083

This implements option C from the issue discussion.

The key ordering is cleanup first, then skip patch, then retain/write back. That removes stale ACP-managed rules instead of orphaning them when a Service opts out.

Unit coverage includes both disabled-floating-IP backend-node rules and the floating-IP-enabled frontend-IP rules. The e2e coverage exercises the disabled-floating-IP case that motivated the issue.

```release-note
Added the `service.beta.kubernetes.io/azure-disable-load-balancer-nsg-rule` Service annotation to disable cloud-provider-managed LoadBalancer NSG rule creation for Services whose NSG rules are intentionally managed by operators.
```

```docs
Documentation should be added in the cloud-provider-azure docs site.
```

### DIFF
--- a/pkg/provider/azure_loadbalancer_accesscontrol_test.go
+++ b/pkg/provider/azure_loadbalancer_accesscontrol_test.go
@@ -499,6 +499,54 @@ func TestCloud_reconcileSecurityGroup(t *testing.T) {
 			assert.NoError(t, err)
 		})
 
+		t.Run("with disabled floating IP and NSG rule management disabled", func(t *testing.T) {
+			var (
+				ctrl                    = gomock.NewController(t)
+				az                      = GetTestCloud(ctrl)
+				securityGroupClient     = az.NetworkClientFactory.GetSecurityGroupClient().(*mock_securitygroupclient.MockInterface)
+				loadBalancerClient      = az.NetworkClientFactory.GetLoadBalancerClient().(*mock_loadbalancerclient.MockInterface)
+				loadBalancerBackendPool = az.LoadBalancerBackendPool.(*MockBackendPool)
+				svc                     = k8sFx.Service().
+							WithDisableFloatingIP().
+							WithDisableLoadBalancerNSGRule().
+							Build()
+				securityGroup = azureFx.SecurityGroup().Build()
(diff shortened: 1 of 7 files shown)


craigktreasure:
This PR supersedes #10475.

pkg/provider/azure_loadbalancer_accesscontrol_test.go
```diff
@@ -1982,6 +2030,171 @@ func TestCloud_reconcileSecurityGroup(t *testing.T) {
...
+			Return(
+				azureFx.LoadBalancer().BackendPoolIPv4Addresses(),
+				azureFx.LoadBalancer().BackendPoolIPv6Addresses(),
+			).
+			Times(1)
+
+		_, err := az.reconcileSecurityGroup(ctx, ClusterName, &svc, *loadBalancer.Name, azureFx.LoadBalancer().Addresses(), EnsureLB)
+		assert.NoError(t, err)
+	})
+
```
Liunardy, line 2197: Is it worth adding test for the ServiceAnnotationDisableLoadBalancerNSGRule removal (on then off) scenario?

nilo19: Hard to test it in unit test but let me test it manually.

nilo19: Manual test passed.

anndono (approved):

PR #10514, opened 2026-06-15
fix(multi-slb): select existing LB with no rules over creating a new one

### LINKED ISSUE #10051: multi-slb: lb selection logic upgrade

  In multi-SLB mode, an existing load balancer that does not have any Service placed on it should be treated the same as a non-existent load balancer during default LB selection.

  Today, getMostEligibleLBForService() treats any matching LB object as "existing" and therefore skips it in the "pick the first eligible LB that does not exist yet" step. This causes surprising placement when an
  external LB already exists for cluster plumbing but has no Service using it.

  Example:

  - multipleStandardLoadBalancerConfigurations = [kubernetes, lb-2]
  - kubernetes external LB object already exists
  - kubernetes has no Service rules / no Service using it. e.g., only contains external rules to host outbound traffic.
  - lb-2 does not exist yet

  Current behavior:

  - first internal Service picks kubernetes-internal
  - first external Service picks lb-2

  Requested behavior:

  - the empty existing kubernetes LB should be treated as "not exist" for placement purposes
  - the first Service should pick kubernetes regardless of whether it is internal or external
  - lb-2 should only be preferred once kubernetes is actually in use by Service traffic, or based on the normal fewest-rules balancing after both are meaningfully in service

  In other words, "LB object exists but has no Service in it" should not change the default placement outcome.

  #### Why is this needed:

  The current behavior is inconsistent and surprising:

  - internal and external Services can land on different default LB configs even with the same config order
  - the result depends on whether an empty external LB object already exists, not on actual Service usage
  - this makes default placement less deterministic and harder for operators to reason about

  Treating an empty LB as equivalent to "not exist yet" would make the behavior more intuitive:

  - config order remains authoritative for the first Service placement
  - internal and external placement become consistent
  - cluster-created but unused LB objects do not skew Service placement unexpectedly

  If useful, I can also turn this into a concrete proposal with acceptance criteria and test scenarios.

Liunardy:
#### What type of PR is this?
/kind bug

In multi-SLB mode, `getMostEligibleLBForService()` skips existing LBs with zero LoadBalancingRules, causing unexpected Service placement when an LB exists for outbound traffic but has no service rules.

This PR adds a rule-count check so an existing LB with zero rules is preferred over a non-existent LB for placement.

Fixes #10051

```release-note
fix(multi-slb): select existing LB with no rules over creating a new one
```

```docs

```

### DIFF
--- a/pkg/provider/azure_loadbalancer.go
+++ b/pkg/provider/azure_loadbalancer.go
@@ -4632,22 +4632,34 @@ func getMostEligibleLBForService(
 		return currentLBName
 	}
 
-	// 2. If the LB is eligible and not created yet, choose it because it has the fewest rules.
+	// 2. If the LB is eligible and has no load balancing rules or is not created yet,
+	// choose it because it has the fewest rules.
+	// Prefer existing LBs with no rules over non-existent ones.
+	var firstNonExistent string
 	for _, eligibleLB := range eligibleLBs {
-		var found bool
+		var found, hasRules bool
 		for i := range existingLBs {
 			existingLB := (existingLBs)[i]
 			if strings.EqualFold(trimSuffixIgnoreCase(ptr.Deref(existingLB.Name, ""), consts.InternalLoadBalancerNameSuffix), eligibleLB) &&
 				isInternalLoadBalancer(existingLB) == isInternal {
 				found = true
+				hasRules = existingLB.Properties != nil && len(existingLB.Properties.LoadBalancingRules) > 0
 				break
 			}
 		}
 
-		if !found {
-			logger.V(4).Info("choose LB as it is eligible and not existing", "eligibleLB", eligibleLB)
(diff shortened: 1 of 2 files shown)


pkg/provider/azure_loadbalancer.go
```diff
@@ -4648,6 +4650,10 @@ func getMostEligibleLBForService(
 			logger.V(4).Info("choose LB as it is eligible and not existing", "eligibleLB", eligibleLB)
 			return eligibleLB
 		}
+		if !hasRules {
```
nilo19: Why do we put this behind if !found? It seems reasonable for existing but empty LB to have higher priority than non-existing one?

Liunardy: Changed to prefer existing but empty LB.

PR #10515, opened 2026-06-15
fix: use network subscription for zone discovery when NetworkResourceSubscriptionID is set

### LINKED ISSUE #10513: LoadBalancer frontend zones are discovered from cluster subscription instead of network subscription
#### What happened:

When `NetworkResourceSubscriptionID` is configured, Azure LoadBalancer network resources are created or updated in the network subscription, but LoadBalancer zone discovery currently uses the cluster/compute subscription.

For internal LoadBalancer frontends, `getFrontendZones()` calls `getRegionZonesBackoff()` and then sets `fipConfig.Zones`. For public LoadBalancer services, the Public IP path calls the same helper and sets `pip.Zones`.

However, the zone repository is initialized from `az.ComputeClientFactory.GetProviderClient()`, and `ComputeClientFactory` is initialized with `az.SubscriptionID`. This means zone metadata is discovered from the cluster subscription even when the frontend IP configuration or Public IP resource belongs to `NetworkResourceSubscriptionID`.

Zone discovery for LoadBalancer frontend IP configurations and Public IP resources should use the network subscription, because these are network resources and are reconciled through the network client factory.

If `NetworkResourceSubscriptionID` is not configured, the behavior should continue to fall back to `az.SubscriptionID`.

1. Configure cloud-provider-azure with different values for `subscriptionId` and `networkResourceSubscriptionID`.
2. Reconcile a LoadBalancer Service that causes CCM to create or update a frontend IP configuration or Public IP.
3. Observe the zone discovery path:
   - `pkg/provider/azure_loadbalancer.go`: `getFrontendZones()` calls `getRegionZonesBackoff()` and sets `fipConfig.Zones`.
   - `pkg/provider/azure_loadbalancer.go`: the Public IP path calls `getRegionZonesBackoff()` and sets `pip.Zones`.
   - `pkg/provider/azure_zones.go`: `getRegionZonesBackoff()` uses `az.zoneRepo.ListZones()`.
   - `pkg/provider/azure.go`: `az.zoneRepo` is initialized with `zone.NewRepo(az.ComputeClientFactory.GetProviderClient())`.
   - `pkg/provider/azure.go`: `ComputeClientFactory` is initialized with `az.SubscriptionID`, while `NetworkClientFactory` is initialized with `az.getNetworkResourceSubscriptionID()`.

The network resource IDs already use the network subscription. For example, the frontend IP configuration ID is built with `az.getNetworkResourceSubscriptionID()`. The inconsistency is specifically in the provider metadata lookup used to populate zones.

Proposed fix: make zone discovery for LoadBalancer/Public IP network-resource zone assignment use the network client factory or otherwise use `az.getNetworkResourceSubscriptionID()`.

Suggested test coverage: initialize the cloud with different `subscriptionId` and `networkResourceSubscriptionID`, then verify the provider client used by the zone repository is created with the network subscription for network-resource zone discovery.

#### Environment:

- Kubernetes version (use `kubectl version`): N/A - code path issue
- Cloud provider or hardware configuration: Azure cloud-provider with `NetworkResourceSubscriptionID` configured
- Install tools: N/A
- Others: Observed in current `master` checkout at commit `5a7a7669c`

Liunardy:
#### What type of PR is this?
/kind bug

`zoneRepo` was initialized from `ComputeClientFactory` instead of `NetworkClientFactory`, causing LB frontend IP and public IP zones to be discovered from the cluster subscription instead of the network subscription.

No behavior change for single-subscription deployments.

Fixes #10513

```release-note
fix: use network subscription for zone discovery when NetworkResourceSubscriptionID is set
```

```docs

```

### DIFF
--- a/pkg/provider/azure.go
+++ b/pkg/provider/azure.go
@@ -460,7 +460,7 @@ func (az *Cloud) InitializeCloudFromConfig(ctx context.Context, config *azurecon
 	}
 
 	if az.zoneRepo == nil {
-		az.zoneRepo, err = zone.NewRepo(az.ComputeClientFactory.GetProviderClient())
+		az.zoneRepo, err = zone.NewRepo(az.NetworkClientFactory.GetProviderClient())
 		if err != nil {
 			return err
 		}
--- a/pkg/provider/azure_test.go
+++ b/pkg/provider/azure_test.go
@@ -49,6 +49,8 @@ import (
 	"sigs.k8s.io/cloud-provider-azure/pkg/azclient/configloader"
 	"sigs.k8s.io/cloud-provider-azure/pkg/azclient/interfaceclient/mock_interfaceclient"
 	"sigs.k8s.io/cloud-provider-azure/pkg/azclient/loadbalancerclient/mock_loadbalancerclient"
+	"sigs.k8s.io/cloud-provider-azure/pkg/azclient/mock_azclient"
+	"sigs.k8s.io/cloud-provider-azure/pkg/azclient/providerclient/mock_providerclient"
 	"sigs.k8s.io/cloud-provider-azure/pkg/azclient/publicipaddressclient/mock_publicipaddressclient"
 	"sigs.k8s.io/cloud-provider-azure/pkg/azclient/securitygroupclient/mock_securitygroupclient"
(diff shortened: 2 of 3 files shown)


nilo19:
Currently we only path zones for network resources, so it would be safe to use network subscription for all resources.
/lgtm

PR #10540, opened 2026-06-16
fix: getServicePIPName returns empty for IPv6 on IPv4-only service

### LINKED ISSUE #10169: getServicePIPNames returns duplicated PIP name for single-stack IPv4 services
#### What happened:

For a single-stack IPv4 service with a PIP name annotation, `getServicePIPNames` returns `["my-pip", "my-pip"]`, the same IPv4 PIP name twice. 
This is because it iterates over both IPv4 and IPv6, and `getServicePIPName(svc, true)` falls through to return the IPv4 annotation value for a single-stack IPv4 service.

https://github.com/kubernetes-sigs/cloud-provider-azure/blob/c3c63a912cd151d42e818110732d23348952b4cd/pkg/provider/azure_utils.go#L379-L403

The duplicate causes redundant PIP lookups
https://github.com/kubernetes-sigs/cloud-provider-azure/blob/c3c63a912cd151d42e818110732d23348952b4cd/pkg/provider/azure_loadbalancer.go#L4379-L4391

`getServicePIPNames` should not return duplicates for a single-stack IPv4 service.

Liunardy:
#### What type of PR is this?
/kind bug

`getServicePIPName` returned the IPv4 PIP annotation when called with `isIPv6=true` on a single-stack IPv4 service, causing `getServicePIPNames` to return `["pip", "pip"]`. This led to redundant `findMatchedPIP` API calls in `serviceOwnsFrontendIP` and `getLoadBalancerNameByPIPName`.

This PR returns `""` early when `isIPv6=true` and the service has no IPv6 family. The same guard is applied to `getServicePIPPrefixID` for consistency, though its callers are currently protected by `v6Enabled` gating.

Fixes #10169

```release-note
fix: IPv6 PIP name lookup on single-stack IPv4 services incorrectly returned the IPv4 annotation value, causing redundant public IP lookups.
```

```docs

```

### DIFF
--- a/pkg/provider/azure_utils.go
+++ b/pkg/provider/azure_utils.go
@@ -388,6 +388,9 @@ func getServicePIPName(service *v1.Service, isIPv6 bool) string {
 				return name
 			}
 		}
+		if isIPv6 && !v6Enabled {
+			return ""
+		}
 		return service.Annotations[consts.ServiceAnnotationPIPNameDualStack[false]]
 	}
 
--- a/pkg/provider/azure_utils.go
+++ b/pkg/provider/azure_utils.go
@@ -419,6 +419,9 @@ func getServicePIPPrefixID(service *v1.Service, isIPv6 bool) string {
 				return id
 			}
 		}
+		if isIPv6 && !v6Enabled {
+			return ""
+		}
 		return service.Annotations[consts.ServiceAnnotationPIPPrefixIDDualStack[false]]
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/provider/azure_utils.go
anndono, line 421: Should we add the same logic here so `getServicePIPPrefixID` also returns empty when called with isIPv6=true on a single-stack IPv4 service?

Liunardy: All callers of getServicePIPPrefixID gate on v6Enabled before passing isIPv6=true, so the wrong return value was never reached in practice.

That said, fixing it here makes the behavior consistent and defends against any future caller that might not observe the same gating convention. Updated with the guard and a matching test case.

PR #10543, opened 2026-06-16
Bump azclient to v0.20.9 for ServiceGateway client

georgeedward2000:
/kind cleanup

 #### What this PR does / why we need it:

 Rolls the ServiceGateway client (added to the pkg/azclient submodule in #10066) into the root module's vendored dependency so providercode can call ClientFactory.GetServiceGatewayClient(). First azclient release containing the client is v0.20.8.

 Bumps `pkg/azclient` v0.20.7 → v0.20.9 and re-vendors it.`go.mod`/`go.sum`/`vendor/` only; no source changes.

 #### Which issue(s) this PR fixes:

 Fixes #

 #### Special notes for your reviewer:

 Mechanical dependency bump: `go getsigs.k8s.io/cloud-provider-azure/pkg/azclient@v0.20.9` followed by`go mod tidy && go mod vendor`. The vendored client is byte-for-byteidentical to the merged submodule source — a follow-up to #10066, not a replacement.

 #### Does this PR introduce a user-facing change?

 NONE

Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.:

 NONE

Issue #10428, opened 2026-05-25
Track Container-Native SLB support


nilo19:
#### What would you like to be added:

Add Container-Native SLB support in `cloud-provider-azure`.

This is an umbrella tracking issue for the Cloud Provider Azure work needed to support Service SKU load balancers with pod IP backends. The intended configuration shape is:

- `loadBalancerSku: service`
- `loadBalancerBackendPoolConfigurationType: podIP`
- `serviceGatewayEnabled: true`

Cloud Provider Azure should reconcile Kubernetes `Service` load balancers through Service Gateway / Service SKU Load Balancer instead of relying on node or NIC backend pool programming.

Expected repository scope:

- Accept and validate the Service LB + PodIP configuration combination without changing behavior for existing `basic` and `standard` clusters.
- Reconcile Service SKU load balancers, frontend IP configuration, load-balancing rules, health behavior, NSG handling, and backend membership for Kubernetes `Service` objects.
- Add Service Gateway client/wrapper integration and use differential or batched update flows for locations and inbound/outbound services.
- Maintain pod address membership from Kubernetes service, pod, and endpoint state, including startup baseline reconciliation and retry/backoff behavior after API failures.
- Support the outbound integration path where Service Gateway and Standard v2 NAT Gateway are enabled.
- Define and document supported behavior for `externalTrafficPolicy`, `healthCheckNodePort`, IP stability across zero-endpoint windows, migration or rollback boundaries, feature gates, and unsupported scenarios.
- Add focused unit and e2e coverage for config validation, service reconciliation, pod churn, retry/recovery behavior, source IP preservation, IPv4/IPv6 behavior, and scale/limit cases.

This issue is intended as the public umbrella issue for future PRs and references. Related prior closed partial tracker: #8395.

#### Why is this needed:

The current Azure load balancer integration is node-oriented: traffic is sent to nodes and then routed to pods. That model adds extra forwarding hops, depends on node-level health/probe behavior, can create uneven distribution or stale routing during pod churn, and makes high-scale container workloads harder to operate.

Container-Native SLB lets Azure load balancing target pod addresses through Service Gateway mappings. This is intended to preserve source IP, avoid node-level forwarding in the ingress path, improve behavior during pod churn, and provide a foundation for per-service or per-pod egress through Standard v2 NAT Gateway.

A single umbrella issue also gives the CNLB PR stack a stable public reference for related code changes, review decisions, tests, and rollout follow-ups.

PR #10068, opened 2026-03-19
CLB 2nd PR: Add DiffTracker core types, state management, and sync operations

georgeedward2000:
#### What type of PR is this?

/kind feature

#### What this PR does / why we need it:

This PR introduces the **DiffTracker core package** -- the foundational types, state management, and diff computation logic for the Container Load Balancer (CLB) feature.

The DiffTracker maintains two parallel state representations:
- **K8s State**: Services, Egresses, and Nodes (with Pods and their inbound/outbound identities)
- **NRP State**: LoadBalancers, NATGateways, and Locations (with Addresses and service references)

It provides operations to mutate both states and compute the diff between them, determining what needs to be synced to Azure.

**Package contents (`pkg/provider/difftracker/`):**

| File | Content |
|------|---------|
| `types.go` | Core enums (`Operation`, `UpdateAction`, `SyncStatus`), K8s/NRP state model types, `DiffTracker` struct, input types, sync result types |
| `config.go` | `Config` struct with Azure subscription, resource group, location, and ServiceGateway fields |
| `difftracker.go` | `New()` constructor with validation |
| `k8s_state_updates.go` | K8s state mutations: services, egresses, endpoints, pods |
| `nrp_state_updates.go` | NRP state mutations: LoadBalancers, NATGateways, location/address updates |
| `sync_operations.go` | Diff computation between K8s and NRP states (`GetSyncOperations`, `GetSyncLoadBalancerServices`, `GetSyncNRPNATGateways`, `GetSyncLocationsAddresses`) |
| `util.go` | Enum serialization, `DeepEqual`, `Equals` comparison methods |

**Dependency changes (`pkg/util/sets/string.go`):**
- Added `Equals()` method to `IgnoreCaseSet` for set comparison
- Added `Difference()` method to `IgnoreCaseSet` for set difference

**Test coverage:**
- `difftracker_test.go` -- 14 test functions covering initialization, state updates, sync operations, and end-to-end scenarios
- `util_test.go` -- 11 test functions covering enum serialization, JSON round-trips, `HasPods`/`HasIdentities` helpers, `DeepEqual`, `Equals`, and `Config.Validate()`
- `string_test.go` -- Added `TestEquals` (8 sub-tests) for the new `IgnoreCaseSet.Equals()` method

[...]

PR #10549, opened 2026-06-17
CLB 3rd PR: ServiceGateway Azure operations layer

georgeedward2000:
**Draft - stacked PR.** Based on #10068 (CLB 2nd PR,DiffTracker core) and #10543 (azclient v0.20.9 bump), neither yet in master. **Please review only the top commit** `CP2: ServiceGatewayAzure operations layer` (`pkg/consts` +`pkg/provider/difftracker/*`). The vendored `azclient` churn and DiffTracker-core files belong to the base PRs and will disappear once they merge and this branch is rebased onto master.

 #### What type of PR is this?
 /kind feature

 Adds the ServiceGateway Azure operations layer for Container/Service Load Balancer: Public IP, Load Balancer, and NAT Gateway CRUD plus ServiceGateway service and address-location updates, along with the DTO types and mappers that translate DiffTracker sync output into Azure SDK requests.

 Part of #10428

 Stacked on #10068 and #10543. Will be rebased onto master (dropping the base commits) once those merge, leaving only the files of this operations layer.

 ```release-note
 NONE


nilo19:
Part of https://github.com/kubernetes-sigs/cloud-provider-azure/issues/10428

pkg/provider/servicegateway/difftracker/azure_operations.go
```diff
@@ -0,0 +1,478 @@
+// Package difftracker provides state tracking and synchronization between Kubernetes
```
david-kow, line 17: lets add license text

pkg/provider/servicegateway/difftracker/azure_operations.go
```diff
@@ -0,0 +1,478 @@
...
+package difftracker
+
+import (
+	"context"
+	"fmt"
+	"net"
+	"strings"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/network/armnetwork/v9"
+	v1 "k8s.io/api/core/v1"
```
david-kow, line 49: do we separate k8s.* and others?

pkg/provider/servicegateway/difftracker/azure_operations.go
```diff
@@ -0,0 +1,478 @@
...
+	servicehelper "k8s.io/cloud-provider/service/helpers"
+	"k8s.io/klog/v2"
+	"k8s.io/utils/ptr"
+
+	"sigs.k8s.io/cloud-provider-azure/pkg/util/errutils"
+)
+
+// createOrUpdatePIP creates or updates a public IP address
+func (dt *DiffTracker) createOrUpdatePIP(ctx context.Context, pipResourceGroup string, pip *armnetwork.PublicIPAddress) error {
+	_, err := dt.createOrUpdatePIPWithResponse(ctx, pipResourceGroup, pip)
```
david-kow, line 63: do we ever need the response?

pkg/provider/difftracker/azure_operations.go
```diff
@@ -0,0 +1,478 @@
...
+		return nil, fmt.Errorf("createOrUpdatePIPWithResponse: pip is nil")
+	}
+	pipName := ptr.Deref(pip.Name, "")
+	if pipName == "" {
+		return nil, fmt.Errorf("createOrUpdatePIPWithResponse: pip name is empty")
+	}
+	klog.Infof("createOrUpdatePIPWithResponse(%s): start", pipName)
+
+	response, err := dt.networkClientFactory.GetPublicIPAddressClient().CreateOrUpdate(ctx, pipResourceGroup, pipName, *pip)
+	klog.V(10).Infof("PublicIPAddressClient.CreateOrUpdate(%s, %s): end", pipResourceGroup, pipName)
```
david-kow: which log level we should have here?

pkg/provider/servicegateway/difftracker/azure_operations.go
```diff
@@ -0,0 +1,478 @@
...
+	if pip == nil {
+		return nil, fmt.Errorf("createOrUpdatePIPWithResponse: pip is nil")
+	}
+	pipName := ptr.Deref(pip.Name, "")
+	if pipName == "" {
+		return nil, fmt.Errorf("createOrUpdatePIPWithResponse: pip name is empty")
+	}
+	klog.Infof("createOrUpdatePIPWithResponse(%s): start", pipName)
+
+	response, err := dt.networkClientFactory.GetPublicIPAddressClient().CreateOrUpdate(ctx, pipResourceGroup, pipName, *pip)
```
david-kow, line 79: just making sure - it's ok to call this from multiple goroutines?

pkg/provider/servicegateway/difftracker/azure_operations.go
```diff
@@ -0,0 +1,478 @@
...
+	var serviceToBeUpdated *armnetwork.ServiceGatewayService
+	for _, service := range services {
+		if service.Name != nil && *service.Name == natGatewayName {
+			serviceToBeUpdated = service
+			break
+		}
+	}
+
+	if serviceToBeUpdated != nil && serviceToBeUpdated.Properties != nil && serviceToBeUpdated.Properties.PublicNatGatewayID != nil {
+		serviceToBeUpdated.Properties.PublicNatGatewayID = nil
```
david-kow, line 182: how we guarantee that this (and the rest of delete) doesn't race with a create? (I mean, it's ok if it's recreated, but they shouldn't interfere)

pkg/provider/difftracker/azure_operations.go
```diff
@@ -0,0 +1,478 @@
...
+}
+
+// updateNRPSGWServices updates services in the Service Gateway
+func (dt *DiffTracker) updateNRPSGWServices(ctx context.Context, serviceGatewayName string, updateServicesRequestDTO ServicesDataDTO) error {
+	if len(updateServicesRequestDTO.Services) == 0 && updateServicesRequestDTO.Action != FullUpdate {
+		klog.Infof("updateNRPSGWServices(%s): no services to update", serviceGatewayName)
+		return nil
+	}
+
+	klog.Infof("updateNRPSGWServices(%s): start", serviceGatewayName)
```
david-kow: why different logger invocation?

pkg/provider/servicegateway/difftracker/azure_operations.go
```diff
@@ -0,0 +1,478 @@
...
+
+	klog.V(10).Infof("ServiceGatewayClient.UpdateAddressLocations(%s): success", serviceGatewayName)
+	klog.Infof("updateNRPSGWAddressLocations(%s): end, error: nil", serviceGatewayName)
+	return nil
+}
+
+// getServiceByUID returns the Service whose UID matches the given uid
+func (dt *DiffTracker) getServiceByUID(ctx context.Context, uid string) (*v1.Service, error) {
+	// list via client (could be expensive; acceptable for initialization)
+	svcList, err := dt.kubeClient.CoreV1().Services(v1.NamespaceAll).List(ctx, metav1.ListOptions{})
```
david-kow: it is expensive - it's probably ok for now as this is once per service (+retries), maybe we should look into optimizing this (storing ns+name) in the future

pkg/provider/difftracker/azure_operations.go
```diff
@@ -0,0 +1,478 @@
...
+			return svc.DeepCopy(), nil
+		}
+	}
+	return nil, fmt.Errorf("service with uid %s not found", uid)
+}
+
+// updateServiceLoadBalancerStatus updates the K8s Service status with the LoadBalancer IP address.
+// This is called after the PIP is successfully created in ServiceGateway mode to populate
+// the Service.Status.LoadBalancer.Ingress field, which would otherwise be empty since
+// EnsureLoadBalancer returns immediately in async mode.
```
david-kow: what are these modes you are referring to here?

pkg/provider/servicegateway/difftracker/azure_operations.go
```diff
@@ -0,0 +1,478 @@
...
+
+	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		svc, err := dt.getServiceByUID(ctx, serviceUID)
+		if err != nil {
+			return fmt.Errorf("updateServiceLoadBalancerStatus: failed to get service: %w", err)
+		}
+
+		desired := make([]v1.LoadBalancerIngress, 0, len(svc.Status.LoadBalancer.Ingress)+1)
+		newPresent := false
+		for _, ingress := range svc.Status.LoadBalancer.Ingress {
```
david-kow, line 317: maybe add a comment for this loop

pkg/provider/difftracker/azure_operations.go
```diff
@@ -0,0 +1,478 @@
...
+				continue
+			}
+			desired = append(desired, ingress)
+		}
+		if !newPresent {
+			desired = append(desired, v1.LoadBalancerIngress{IP: ip})
+		}
+
+		if loadBalancerIngressEqual(svc.Status.LoadBalancer.Ingress, desired) {
+			klog.V(3).Infof("updateServiceLoadBalancerStatus: service %s/%s already has IP %s", svc.Namespace, svc.Name, ip)
```
david-kow: why this is guarded with this equal? isn't this always the same as !newPresent?

pkg/provider/difftracker/azure_operations.go
```diff
@@ -0,0 +1,478 @@
...
+		return nil
+	})
+}
+
+func loadBalancerIngressEqual(a, b []v1.LoadBalancerIngress) bool {
+	if len(a) != len(b) {
+		return false
+	}
+	for i := range a {
+		if a[i].IP != b[i].IP || a[i].Hostname != b[i].Hostname {
```
david-kow: what about other fields?

pkg/provider/servicegateway/difftracker/azure_operations.go
```diff
@@ -0,0 +1,478 @@
...
+
+// extractResourceChildName extracts a child resource name from an Azure resource ID.
+// Returns "" if the requested segment is not present.
+func extractResourceChildName(id, segment string) string {
+	if id == "" {
+		return ""
+	}
+	parts := strings.Split(id, "/")
+	for i := 0; i < len(parts)-1; i++ {
+		if strings.EqualFold(parts[i], segment) && parts[i+1] != "" {
```
david-kow, line 386: Is `&& parts[i+1] != ""` needed? if it's empty we'll return "" anyway outside of the loop. unless we care about iterating through entire parts.

pkg/provider/difftracker/dto_mappers.go
```diff
@@ -0,0 +1,80 @@
...
+			ServiceType: Inbound,
+			LoadBalancerBackendPools: []LoadBalancerBackendPoolDTO{
+				{
+					Id: fmt.Sprintf(
+						consts.BackendPoolIDTemplate,
+						subscriptionID,
+						resourceGroup,
+						service,
+						service,
+						// fmt.Sprintf("%s-backendpool", service),
```
david-kow: needed?

pkg/provider/difftracker/resource_helpers.go
```diff
@@ -0,0 +1,326 @@
...
+			Name:       to.Ptr(backendPoolName),
+			Properties: &armnetwork.BackendAddressPoolPropertiesFormat{
+				// Backend pool will be populated by ServiceGateway with pod IPs
+			},
+		},
+	}
+
+	// Build LB rules and probes from config
+	var lbRules []*armnetwork.LoadBalancingRule
+	var probes []*armnetwork.Probe
```
david-kow: I think it's confusing to have it here, can we move this closer to the usage (or just remove it) with the comment reg. not having support for that

pkg/provider/difftracker/resource_helpers.go
```diff
@@ -0,0 +1,326 @@
...
+			lbRules = append(lbRules, &armnetwork.LoadBalancingRule{
+				Name:       to.Ptr(ruleName),
+				Properties: ruleProps,
+			})
+
+			klog.V(4).Infof("buildInboundServiceResources: created LB rule %s: frontend=%d backend=%d protocol=%s for service %s",
+				ruleName, frontendPort.Port, backendPort, frontendPort.Protocol, serviceUID)
+		}
+	} else {
+		klog.V(2).Infof("buildInboundServiceResources: no port configuration provided for service %s, creating LB without rules", serviceUID)
```
david-kow: what would happen in this case? would we reconcile it later correctly?

pkg/provider/difftracker/resource_helpers.go
```diff
@@ -0,0 +1,326 @@
...
+		},
+		dtConfig.SubscriptionID,
+		dtConfig.ResourceGroup,
+	)
+
+	return pip, natGateway, servicesDTO
+}
+
+// newIgnoreCaseSetFromSlice creates an IgnoreCaseSet from a slice of strings
+func newIgnoreCaseSetFromSlice(items []string) *utilsets.IgnoreCaseSet {
```
david-kow: can we do `utilsets.NewString(items...)` instead?

pkg/provider/difftracker/resource_helpers.go
```diff
@@ -0,0 +1,326 @@
...
+		backendPort := port.Port
+		switch port.TargetPort.Type {
+		case intstr.Int:
+			if port.TargetPort.IntVal > 0 {
+				backendPort = port.TargetPort.IntVal
+			}
+		case intstr.String:
+			klog.Warningf("ExtractInboundConfigFromService: named targetPort %q is not supported for service %s/%s; rejecting",
+				port.TargetPort.StrVal, service.Namespace, service.Name)
+			return nil
```
david-kow: should this return nil or separate err? not sure how this will be used

pkg/provider/difftracker/sync_operations.go
```diff
@@ -0,0 +1,226 @@
...
+
+import (
+	"k8s.io/klog/v2"
+
+	utilsets "sigs.k8s.io/cloud-provider-azure/pkg/util/sets"
+)
+
+// GetServicesToSync handles the synchronization of services between K8s and NRP
+func GetServicesToSync(k8sServices, nrpServices *utilsets.IgnoreCaseSet) SyncServicesReturnType {
+	klog.V(2).Infof("GetServicesToSync: K8s services (%d): %v", k8sServices.Len(), k8sServices.UnsortedList())
```
david-kow: we should probably call this only once if needed, iiuc each invocation enumerates it

pkg/provider/difftracker/sync_operations.go
```diff
@@ -0,0 +1,226 @@
...
+	klog.V(2).Infof("GetServicesToSync: NRP services (%d): %v", nrpServices.Len(), nrpServices.UnsortedList())
+
+	syncServices := SyncServicesReturnType{
+		Additions: utilsets.NewString(),
+		Removals:  utilsets.NewString(),
+	}
+
+	for _, service := range k8sServices.UnsortedList() {
```
david-kow: can we use sets functions instead of those loops?

pkg/provider/servicegateway/difftracker/sync_operations.go
```diff
@@ -0,0 +1,226 @@
...
+	// Iterate over all locations in the NRP state
+	for location, nrpLocation := range dt.NRPResources.Locations {
+		node, exists := dt.K8sResources.Nodes[location]
+		if !exists {
+			result.Locations[location] = Location{
+				AddressUpdateAction: PartialUpdate,
+				Addresses:           make(map[string]Address),
+			}
+		} else {
+			locationData := findLocationData(result, location)
```
david-kow, line 126: is this helper really saving much here?

pkg/provider/difftracker/resource_helpers.go
georgeedward2000: Changed it to not return nil. Returning nil here would silently drop the port and break the service. Instead, I fall back to  backendPort = port.Port  (same default as when TargetPort is unset) and log a warning. We can switch to a hard error later if we decide named ports shouldn't be allowed.

pkg/provider/servicegateway/difftracker/azure_operations.go
georgeedward2000: Done. Added the Apache license header to azure_operations.go, and also to resource_helpers.go which was missing it too.

pkg/provider/servicegateway/difftracker/azure_operations.go
georgeedward2000: Well, based on the other files, we don’t separate them. Our goimports config sets local-prefixes to sigs.k8s.io/cloud-provider-azure, so there are three groups: stdlib, third party (github.com and k8s.io together), and our local module. This matches azure_loadbalancer.go and the rest of pkg/provider, so the current grouping is already correct and lint clean.

pkg/provider/servicegateway/difftracker/azure_operations.go
georgeedward2000: Yes, the inbound path (in the engine PR) uses the response to read the allocated PIP IP and patch it into the Service status. The thin createOrUpdatePIP wrapper is used for the outbound path, which doesn’t require the IP. Both behaviors are intentional.

pkg/provider/servicegateway/difftracker/azure_operations.go
georgeedward2000: Yes. The function keeps no shared mutable state: it only reads networkClientFactory, which is set once at construction, plus local variables. The Azure SDK clients are azcore-based and safe for concurrent use. On top of that, ServiceUpdater (engine PR) serializes per service (via the activeOperations map under a mutex), and each service has its own PIP name (uid-pip), so two goroutines never call CreateOrUpdate on the same PIP. Concurrency is also capped at 10 (currently) by the semaphore.

pkg/provider/difftracker/azure_operations.go
georgeedward2000: Good catch, the levels here are inconsistent. V(10) is effectively never printed, while the matching start log is at Infof (always on), so the pair is mismatched. I’ll standardize the per‑SDK call begin/end traces at V(4) to align with the rest of the package, keep failures at Warningf/Errorf, and reserve Infof/V(2) only for meaningful state changes.

pkg/provider/servicegateway/difftracker/azure_operations.go
georgeedward2000: The engine serializes per service, so a delete and a create of the same service never run concurrently. processBatch sets activeOps[uid] under a mutex before dispatching and skips anything already in flight, and each service has a single state machine entry in pendingServiceOps, so create and delete are mutually exclusive states rather than concurrent operations. Each worker clears activeOps[uid] and requeues in its defer, so if a create is pending after a delete, it runs strictly after the delete finishes (recreate is fine, with no interleaving).

Across different services there is no shared resource: every PIP, LB, and NAT gateway is named by the service UID, and the SGW updates use PartialUpdate, touching only that service’s own entry. These guarantees live in the engine (service_updater.go), which is the real caller, not in this PR.

pkg/provider/difftracker/azure_operations.go
georgeedward2000: Same point as the V(10) comment above. The file mixed klog.Infof (always on) with klog.V(N).Infof. I’ve normalized the whole file: routine begin/end traces are now at V(4), meaningful state changes at V(2), and failures at Warningf/Errorf. The updateNRPSGWServices lines are now at V(4), so the invocations are consistent.

pkg/provider/servicegateway/difftracker/azure_operations.go
georgeedward2000: Agreed, it is a full list and not cheap, but it’s bounded: it runs once per service operation plus conflict retries, not in any hot path. For a future optimization, I’ll track a UID → namespace/name map maintained from Service informer events, so this becomes an O(1) lookup followed by a direct Get (or a lister cache read) instead of a NamespaceAll list. I’ll also leave a TODO here pointing to that.

pkg/provider/difftracker/azure_operations.go
georgeedward2000: You are right, that referred to the old sync vs. async split, which no longer exists, so I’ve removed the “mode” wording. The comment now simply states that EnsureLoadBalancer returns an empty status immediately while the engine provisions the PIP, LB, and SGW registration in the background, and this function backfills the ingress IP.

pkg/provider/servicegateway/difftracker/azure_operations.go
georgeedward2000: Added a comment. The loop rebuilds the ingress list in a dual-stack-safe way: it preserves hostname-only entries and entries from the other IP family, replaces any stale IP from the same family with the new one, and skips appending when the new IP is already present.

pkg/provider/difftracker/azure_operations.go
georgeedward2000: No, they are not the same. !newPresent only answers “do we need to add the new IP,” while the equality check answers “is the recomputed list identical to the current one,” which also covers removing a stale same‑family IP.

For example, if the status is [10.0.0.1, 10.0.0.9] and we set 10.0.0.1, newPresent is true, so a !newPresent guard would skip the patch and 10.0.0.9 would leak. However, the desired state becomes [10.0.0.1] because the stale same‑family entry is dropped, so the equality check is false and we correctly patch to remove it.

I verified this with a quick test. The equality guard also avoids unnecessary writes and conflict retries when nothing actually changed.

pkg/provider/difftracker/azure_operations.go
georgeedward2000: Done. Switched the guard to apiequality.Semantic.DeepEqual on the full ingress slices and removed the hand‑rolled comparison, so IPMode and Ports are now covered as well. This is the same helper we already use in nodemanager.

pkg/provider/servicegateway/difftracker/azure_operations.go
georgeedward2000: It is not strictly needed for our inputs. The IDs are template-generated and always well-formed, so the next part is never empty. The guard only matters for malformed IDs where the same segment appears twice and the first value is empty; it allows the loop to skip the empty match and still find a later valid one. It’s just a cheap defensive check, and I’m happy to drop it if you prefer the simpler version.

pkg/provider/difftracker/dto_mappers.go
georgeedward2000: No, not needed. Removed it, it was a leftover comment. The template takes the service name twice (LB name and backend pool name), which is already covered by the two service arguments.

pkg/provider/difftracker/resource_helpers.go
georgeedward2000: Good point. Removed the probes variable entirely since it was always nil, and the Probes field now defaults to nil. I kept a short comment explaining that PodIP backend pools do not support health probes, which is why we do not create any.

pkg/provider/difftracker/resource_helpers.go
georgeedward2000: This is defensive now. After the named targetPort fix, a service with ports always yields frontend ports, and a LoadBalancer service always has at least one, so we no longer build a rule-less LB for real services. If this path were ever hit and ports were added later, UpdateService would diff the config and updateInboundService would rebuild the LB with rules, so it still reconciles correctly.

pkg/provider/difftracker/resource_helpers.go
georgeedward2000: Done. Replaced the helper with utilsets.NewString(serviceUID) at all call sites, since NewString is variadic, and removed newIgnoreCaseSetFromSlice. I also dropped its two tests, as NewString is already covered in pkg/util/sets.

pkg/provider/difftracker/sync_operations.go
georgeedward2000: This file is part of the base DiffTracker PR #10068, not the ServiceGateway Azure ops layer in this PR, so the fix should land there. The concern is valid though: the UnsortedList args are evaluated before the klog call regardless of verbosity, so the enumeration runs even when v=2 is off. The fix is to wrap these logs in if klog.V(2).Enabled() so we only enumerate when the line will actually print. I will track it against #10068.

pkg/provider/servicegateway/difftracker/sync_operations.go
georgeedward2000: This file is part of the base DiffTracker PR #10068, not the ServiceGateway ops layer in this PR, so this should be addressed there. On the point itself, I agree: the helper is a one-line map lookup used in a single place, so it doesn’t add much value. I would inline it as loc, ok := result.Locations[location] and use ok directly. As a side note, it currently returns the address of the range copy rather than the map entry, so a comma‑ok lookup is also cleaner and safer. I will track this against #10068.

georgeedward2000 (commented):
addressed comments

pkg/provider/servicegateway/difftracker/azure_operations.go
robbiezhang, line 1: please move the files (and difftracker folder) into pkg/provider/servicegateway folder.

pkg/provider/servicegateway/difftracker/azure_operations.go
```diff
@@ -0,0 +1,475 @@
...
+	"k8s.io/client-go/util/retry"
+	servicehelper "k8s.io/cloud-provider/service/helpers"
+	"k8s.io/utils/ptr"
+
+	"sigs.k8s.io/cloud-provider-azure/pkg/log"
+	"sigs.k8s.io/cloud-provider-azure/pkg/util/errutils"
+)
+
+// createOrUpdatePIP creates or updates a public IP address
+func (dt *DiffTracker) createOrUpdatePIP(ctx context.Context, pipResourceGroup string, pip *armnetwork.PublicIPAddress) error {
```
robbiezhang, line 62: why this is a function on DiffTracker? I think you need a higher controller to conduct the operations

pkg/provider/difftracker/azure_operations.go
```diff
@@ -0,0 +1,475 @@
...
+// containing the allocated IP address. Use this when you need the IP address after PIP creation.
+func (dt *DiffTracker) createOrUpdatePIPWithResponse(ctx context.Context, pipResourceGroup string, pip *armnetwork.PublicIPAddress) (*armnetwork.PublicIPAddress, error) {
+	if pip == nil {
+		return nil, fmt.Errorf("createOrUpdatePIPWithResponse: pip is nil")
+	}
+	pipName := ptr.Deref(pip.Name, "")
+	if pipName == "" {
+		return nil, fmt.Errorf("createOrUpdatePIPWithResponse: pip name is empty")
+	}
+	dt.logger.V(5).Info("Creating or updating public IP", "pip", pipName, "resourceGroup", pipResourceGroup)
```
robbiezhang: it's an important log, should be logged at level 2

pkg/provider/difftracker/azure_operations.go
```diff
@@ -0,0 +1,475 @@
...
+
+	return response, nil
+}
+
+// deletePublicIP deletes a public IP address
+func (dt *DiffTracker) deletePublicIP(ctx context.Context, pipResourceGroup string, pipName string) error {
+	if pipName == "" {
+		return fmt.Errorf("deletePublicIP: pipName is empty")
+	}
+	dt.logger.V(5).Info("Deleting public IP", "pip", pipName, "resourceGroup", pipResourceGroup)
```
robbiezhang: please log at level 2

pkg/provider/difftracker/azure_operations.go
```diff
@@ -0,0 +1,475 @@
...
+		return fmt.Errorf("creating or updating LoadBalancer %q: %w", lbName, err)
+	}
+
+	return nil
+}
+
+// deleteLB deletes a load balancer by service UID
+func (dt *DiffTracker) deleteLB(ctx context.Context, uid string) error {
+	uid = strings.ToLower(uid)
+	dt.logger.V(5).Info("Deleting LoadBalancer", "loadBalancer", uid)
```
robbiezhang: please log at v2

pkg/provider/difftracker/azure_operations.go
```diff
@@ -0,0 +1,475 @@
...
+
+	return nil
+}
+
+// deleteNatGateway deletes a NAT gateway
+func (dt *DiffTracker) deleteNatGateway(ctx context.Context, natGatewayResourceGroup string, natGatewayName string) error {
+	if natGatewayName == "" {
+		return fmt.Errorf("deleteNatGateway: NAT gateway name is empty")
+	}
+	dt.logger.V(5).Info("Deleting NAT Gateway", "natGateway", natGatewayName, "resourceGroup", natGatewayResourceGroup)
```
robbiezhang: please log at v2

pkg/provider/servicegateway/difftracker/azure_operations.go
```diff
@@ -0,0 +1,475 @@
...
+	}
+	return nil, fmt.Errorf("service with uid %s not found", uid)
+}
+
+// updateServiceLoadBalancerStatus updates the K8s Service status with the LoadBalancer IP address.
+// EnsureLoadBalancer returns an empty LoadBalancer status immediately while the difftracker
+// engine provisions the PIP, LB and ServiceGateway registration asynchronously in the
+// background. This function backfills Service.Status.LoadBalancer.Ingress once the PIP is
+// created, since it would otherwise stay empty.
+func (dt *DiffTracker) updateServiceLoadBalancerStatus(ctx context.Context, serviceUID string, ip string) error {
```
robbiezhang, line 295: where is this function called? why does the caller only know the serviceUID instead of namespace/name?

pkg/provider/servicegateway/difftracker/azure_operations.go
georgeedward2000: Done. Moved the difftracker folder and all its files under pkg/provider/servicegateway/difftracker , consistent with the same move in PR2.

pkg/provider/servicegateway/difftracker/azure_operations.go
georgeedward2000: DiffTracker is the reconciliation engine, so these are its internal Azure primitives (all unexported and never called from outside the package). The same engine conducts them via the updater triggers (`triggerServiceUpdater` / `triggerLocationsUpdater`) and the public entry points (`AddService`, `UpdateService`, `DeleteService`), which land in the next PRs. Above that, the CCM provider (`EnsureLoadBalancer` plus the pod and endpoint informers) acts as the controller feeding events into the engine. This PR only introduces the operation primitives, DTO mappers, and resource builders, so nothing sequences them yet.

pkg/provider/difftracker/azure_operations.go
georgeedward2000: Done. Raised it to V(2) and applied the same change to the sibling resource mutations (LoadBalancer, NAT Gateway, the SGW service and address-location updates, and the Service status patch), since the convention places system state changes at V(2). Left no-op and “already in sync” paths at trace level.

pkg/provider/difftracker/azure_operations.go
georgeedward2000: Done, now at V(2)

pkg/provider/difftracker/azure_operations.go
georgeedward2000: Done, now at V(2)

pkg/provider/difftracker/azure_operations.go
georgeedward2000: Done, now at V(2)

pkg/provider/servicegateway/difftracker/azure_operations.go
georgeedward2000: It isn’t wired yet in this PR. In the full design, it is called from the service updater (to backfill Service.Status.LoadBalancer.Ingress after the PIP is provisioned, since EnsureLoadBalancer returns an empty status) and from the init and crash-recovery path (to re-patch Services whose status was lost across a restart).

It takes serviceUID because the engine is UID-keyed throughout (Kubernetes state, NRP LoadBalancer and NAT Gateway sets, pod identities, and the LoadBalancer resource name is the lowercased UID). UID is also the safer key: it is immutable and unique, so it cannot patch a deleted-and-recreated Service of the same namespace and name with a stale IP. It resolves the UID to the live Service via getServiceByUID to obtain namespace and name for the patch.

robbiezhang: my main concern is that getServiceByUID is not efficient. It lists all services and compares the UID. I'm wondering if you can add hint in the resource caller consumes to effectively correlate the service and pip? as I know, we set tag "k8s-azure-service" to the service namespace/name on the pip created for SLB for fast correlation.

can you do the same?

pkg/provider/servicegateway/difftracker/azure_operations.go
robbiezhang: `DiffTracker is the reconciliation engine`

I cannot tell from the name "DiffTracker". impressed from the 2nd PR, I thought DiffTracker is an in-memory state cache. However, this PR indicates that it's also responsible to call ARM. Such mixed implementation make it easy to make mistake (call a private function which lock the mt while calling ARM). How do you prevent that?

reading these functions to operate on the Azure network resource, I think it deserves to have a controller, which hold the DiffTracker, NetworkClient, K8sClient, to reconcile the resources.

pkg/provider/servicegateway/difftracker/azure_operations.go
```diff
@@ -0,0 +1,475 @@
...
+
+	// Step 1: clear the ServiceGateway-side reference if it still has one.
+	services, err := dt.networkClientFactory.GetServiceGatewayClient().GetServices(ctx, dt.config.ResourceGroup, serviceGatewayName)
+	if err != nil {
+		return fmt.Errorf("getting Service Gateway services: %w", err)
+	}
+
+	var serviceToBeUpdated *armnetwork.ServiceGatewayService
+	for _, service := range services {
+		if service.Name != nil && *service.Name == natGatewayName {
```
robbiezhang: is the service name case sensitive?

pkg/provider/servicegateway/difftracker/azure_operations.go
georgeedward2000: I’d keep it as is but I am open for further discussions.

Locking and ARM are already split by file. Every method that takes dt.mu is in the in-memory files (k8s_state_updates.go, nrp_state_updates.go, sync_operations.go, util.go) and makes no ARM calls, while every ARM call is in azure_operations.go and takes no lock. The lock is never held across I/O, so the issue you’re concerned about cannot occur.

A controller would not remove the coupling, only relay it. Reconciliation requires a consistent snapshot of the in-memory state to compute the diff and act on it. A separate controller would still go through DiffTracker’s lock for each decision, adding indirection without improving isolation.

DiffTracker is already that engine. Its components are clearly separated by file: in-memory state, sync/diff logic, lock-guarded state updates, and lock-free Azure primitives. It already holds networkClientFactory and kubeClient and owns the reconciliation entry points. A Controller wrapping a state-only DiffTracker would duplicate those fields, introduce an additional layer without a meaningful boundary, and churn follow-up PRs and the umbrella for no behavior change.

pkg/provider/servicegateway/difftracker/azure_operations.go
georgeedward2000: Good catch. It was case-sensitive (==), which is inconsistent with how these identities are handled elsewhere: NRP service sets use IgnoreCaseSet, and deleteLB lowercases the UID. If NRP returned the name in a different case, the disassociation would silently no-op. Changed it to strings.EqualFold.

pkg/provider/servicegateway/difftracker/azure_operations.go
georgeedward2000: Agreed, the full List does not scale.

One nuance: the k8s-azure-service tag correlates a PIP to a Service (used by the reverse-lookup SLB during PIP GC), but our bottleneck is the forward direction, resolving a service UID to the Kubernetes Service object in order to patch its status. A PIP tag does not help with that, so it does not remove the List in this path.

The scalable fix is to reuse infrastructure we already have. Carry the Service namespace and name in ServiceConfig (EnsureLoadBalancer already has the *v1.Service) and resolve it through the provider’s existing serviceLister (backed by a SharedInformer), using serviceLister.Services(ns).Get(name), and keeping the UID only as a consistency check. This turns the lookup into an O(1), zero-apiserver, zero-extra-memory cached read instead of a NamespaceAll List.

getServiceByUID is not wired to a caller in this PR, and the engine here only has the raw clientset, not the lister. I have updated the TODO to reflect this plan and will implement it in the PR that wires the lister and ServiceConfig into the engine.

david-kow: See https://github.com/kubernetes-sigs/cloud-provider-azure/pull/10549#discussion_r3465794570 - this is expected to run once per lifetime of the Service resource. I think we should be fine unless we have tens of thousands Services + high churn (not sure if Services high churn is an actual scenario though)

pkg/provider/servicegateway/difftracker/azure_operations.go
```diff
@@ -0,0 +1,478 @@
...
+			},
+		}
+		if err := dt.networkClientFactory.GetServiceGatewayClient().UpdateServices(ctx, dt.config.ResourceGroup, serviceGatewayName, updateServicesRequest); err != nil {
+			return fmt.Errorf("updating Service Gateway to disassociate NAT Gateway %q: %w", natGatewayName, err)
+		}
+	}
+
+	// Step 2: clear the NAT-gateway-side reference if it still points back at the
+	// Service Gateway. This runs independently of Step 1 so a retry after a
+	// partial failure still reconciles the NAT gateway.
```
robbiezhang, line 199: what if the application exit before calling update? this can happen during deployment rolling upgrade. will the new pod continues the reconciliation?

georgeedward2000: Yes, the new pod continues it. This is not a linear in-memory script; it's a level-triggered reconciliation loop, so a crash or rolling upgrade at any point (including immediately before the Step 2 update) is recovered when the next pod starts.

On every CCM initialization, we diff the desired state (K8s egresses) against the actual state (Azure/NRP NAT gateways). Any NAT gateway that still exists in Azure but is no longer desired by Kubernetes is detected as orphaned (!K8sResources.Egresses.Has(natName)) and routed back through the same deleteOutboundService teardown path.

Each step is idempotent, so retries simply complete whatever work remains. disassociate... re-GETs both sides and only clears references that still exist (Step 1 is gated on PublicNatGatewayID != nil, Step 2 on ServiceGateway != nil). Similarly, UpdateServices and deleteNatGateway treat NotFound as success. In-memory progress is intentionally not trusted across restarts; instead, intent is rebuilt from Kubernetes and Azure state on every reconciliation cycle.

Validated end-to-end: a test deliberately terminates the CCM mid-lifecycle and verifies that the replacement pod successfully completes the reconciliation.

pkg/provider/servicegateway/difftracker/azure_operations.go
robbiezhang: Even though the function/property is private, it’s still accessible from other methods on the same struct, so it would be easy for future changes to accidentally hold the lock while making external requests.

I think encapsulating the lock-protected state in a dedicated DiffTracker and introducing a higher-level controller (SGW-reconcilier) would help reduce that risk. The controller can fetch the full data/diff from DiffTracker through public methods, call networkClient to update the SGW, and then, based on the result, call k8sClient to update the Kubernetes resource (on success) or back off reconciliation (on failure). This keeps the lock boundary local to DiffTracker and avoids mixing state tracking with external side effects.

pkg/provider/servicegateway/difftracker/azure_operations.go
robbiezhang: can you use field selector to filter services with LoadBalancer type?

Issue #10054, opened 2026-03-16
Agent-fist development workflow


nilo19:
Build the resources that instruct agents to better handle everything related to this project.

1. Agents.md instructions.
2. Skills specific to this repo
    - releases, including pushing tags, handle release notes, publish builds...
    - run any e2e case
    - unblock dependabot prs
    - building and pushing images
    - cherry picks
    - rebase-based confliction resolver.

PR #10603, opened 2026-06-25
Add shared build-images agent skill

nilo19:
#### What type of PR is this?

/kind documentation

Adds a repo-owned shared `build-images` skill under `.agents/skills/` for building cloud-provider-azure images through the existing Makefile targets. The skill includes a Python helper that requires explicit `IMAGE_TAG` and `IMAGE_REGISTRY`, supports CCM/CNM/HPP aliases, dry-runs shell-quoted commands, allows controlled `--set` / `--unset` flag changes, and scrubs make-control environment overrides before real execution.

Also updates the shared skill catalog and build guidance, including the actual Linux CNM target name.

Related: #10054

Validation performed:
- `PYTHONDONTWRITEBYTECODE=1 python3 .agents/skills/build-images/scripts/test_build_image.py`
- helper `--help` and dry-runs for `ccm`, `cnm --set ARCH=arm64 --unset GOEXPERIMENT`, `hpp`, `all`, and `cnm-all`
- reserved `--set MAKEFLAGS=...` rejection and inherited `MAKEFLAGS` dry-run cleanup
- `quick_validate.py .agents/skills/build-images` with temporary PyYAML under `/tmp`
- `git diff --check`
- targeted boilerplate check for the new Python files

```release-note
NONE
```

```docs
NONE
```

### DIFF
--- a/.agents/skills/build-images/scripts/build_image.py
+++ b/.agents/skills/build-images/scripts/build_image.py
@@ -1,0 +1,256 @@
+#!/usr/bin/env python3
+# Copyright 2026 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import annotations
+
+import argparse
+import os
+import re
+import shlex
+import subprocess
+import sys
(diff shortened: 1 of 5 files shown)


.agents/skills/build-images/scripts/build_image.py
```diff
@@ -0,0 +1,255 @@
...
+            inherited_env=os.environ,
+        )
+    except ValueError as exc:
+        print(f"error: {exc}", file=sys.stderr)
+        return 2
+
+    if args.dry_run:
+        print(format_dry_run(plan))
+        return 0
+    return run_plan(plan)
```
Liunardy, line 252: Nit: maybe always print the resolved command before invoking the real run.

nilo19: Done. Real executions now print the same resolved cwd/command block before invoking `make`, while `--dry-run` still prints without running. I also updated the unit test and skill text to cover/document this behavior.

nilo19: done

PR #10605, opened 2026-06-26
Add toleration on VM shown up delay in ARM/ARG for new nodes

### LINKED ISSUE #10604: New Node is Deleted because the VM/VMSS instance is not shown up in ARM
#### What happened:
New node is deleted right after joined the cluster.
node_lifecycle_controller logged a "Instance not found" error and deleted the node.
Suspect the VM is bootstrapped too fast before the instance is not propagated into the ARM.
Suggest to add some grace period to tolerate such situation.

New nodes should not be deleted in case of VM state not synced in ARM.

No constant repro due to dependencies.

ARM seems to switching to ARG in querying the VM from the VMSS in some region. There might be additional delays. Because node registered much earlier than the operation completed graph was delayed and CCM couldn't find the resource.

#### Environment:

- Kubernetes version (use `kubectl version`): 1.34 (any version using CCM)
- Install tools:
- Others:

robbiezhang:
#### What type of PR is this?

/kind flake

Add a configurable grace period for newly registered Kubernetes nodes whose backing Azure VM/VMSS instance is not immediately visible in ARM.

When a new node bootstraps and joins the cluster faster than its VM/VMSS instance becomes visible in ARM (notably in regions where ARM resolves VMSS instances via ARG, which can add latency), InstanceExists returns false. The cloud-node-lifecycle controller interprets this as a non-existent instance and deletes the freshly joined node.

The new nodeInstanceNotFoundGracePeriodInSeconds config option allows cloud-provider-azure to tolerate this transient “instance not found” state for recently created nodes. During the grace period, InstanceExists reports the node as existing instead of allowing premature deletion.

Fixes #10604

```release-note
Add a configurable grace period (nodeInstanceNotFoundGracePeriodInSeconds) in the cloud config for newly registered Kubernetes nodes whose backing Azure VM/VMSS instance is not immediately visible in ARM.
```

### DIFF
--- a/cmd/cloud-controller-manager/app/config/config.go
+++ b/cmd/cloud-controller-manager/app/config/config.go
@@ -63,6 +63,12 @@ type Config struct {
 
 	// Node filtering configuration
 	NodeFilterRequirements string
+
+	// NodeInstanceNotFoundGracePeriodInSeconds is the period, measured from a node's
+	// creation timestamp, during which a node whose backing VM/VMSS instance is not
+	// yet visible in ARM is still reported as existing, preventing premature node
+	// deletion by the cloud-node-lifecycle controller.
+	NodeInstanceNotFoundGracePeriodInSeconds int
 }
 
 type DynamicReloadingConfig struct {
--- a/cmd/cloud-controller-manager/app/config/config.go
+++ b/cmd/cloud-controller-manager/app/config/config.go
@@ -63,12 +63,6 @@ type Config struct {
 
 	// Node filtering configuration
 	NodeFilterRequirements string
-
-	// NodeInstanceNotFoundGracePeriodInSeconds is the period, measured from a node's
-	// creation timestamp, during which a node whose backing VM/VMSS instance is not
-	// yet visible in ARM is still reported as existing, preventing premature node
-	// deletion by the cloud-node-lifecycle controller.
-	NodeInstanceNotFoundGracePeriodInSeconds int
 }
 
 type DynamicReloadingConfig struct {
--- a/pkg/provider/azure.go
+++ b/pkg/provider/azure.go
@@ -94,6 +94,16 @@ type Cloud struct {
 	VMSet                   VMSet
 	LoadBalancerBackendPool BackendPool
 
+	// NodeInstanceNotFoundGracePeriodInSeconds is the period, measured from a node's
+	// creation timestamp, during which a node whose VM/VMSS instance is not yet
+	// visible in ARM is still reported as existing. This tolerates the delay
+	// between a node registering itself in the cluster and its backing instance
+	// propagating into ARM, preventing the cloud-node-lifecycle controller from
+	// deleting the node prematurely. It is configured via the
+	// --node-instance-not-found-grace-period-in-seconds command-line flag and
+	// defaults to 0 (disabled).
+	NodeInstanceNotFoundGracePeriodInSeconds int
+
 	// ipv6DualStack allows overriding for unit testing.  It's normally initialized from featuregates
 	ipv6DualStackEnabled bool
 	// Lock for access to node caches, includes nodeZones, nodeResourceGroups, and unmanagedNodes.
--- a/pkg/provider/azure.go
+++ b/pkg/provider/azure.go
@@ -94,16 +94,6 @@ type Cloud struct {
 	VMSet                   VMSet
 	LoadBalancerBackendPool BackendPool
 
-	// NodeInstanceNotFoundGracePeriodInSeconds is the period, measured from a node's
-	// creation timestamp, during which a node whose VM/VMSS instance is not yet
-	// visible in ARM is still reported as existing. This tolerates the delay
-	// between a node registering itself in the cluster and its backing instance
-	// propagating into ARM, preventing the cloud-node-lifecycle controller from
-	// deleting the node prematurely. It is configured via the
-	// --node-instance-not-found-grace-period-in-seconds command-line flag and
-	// defaults to 0 (disabled).
-	NodeInstanceNotFoundGracePeriodInSeconds int
-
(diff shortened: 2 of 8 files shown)


pkg/provider/azure_instances_v2.go
```diff
@@ -51,15 +54,50 @@ func (az *Cloud) InstanceExists(ctx context.Context, node *v1.Node) (bool, error
...
+	if node.CreationTimestamp.IsZero() {
+		return false
+	}
+
+	nodeAge := time.Since(node.CreationTimestamp.Time)
+	if nodeAge >= gracePeriod {
+		return false
+	}
+
+	logger.Info("Instance not found in ARM but the node is still within the grace period; reporting it as existing to avoid premature deletion",
```
nilo19: Suggest to have a V(2) here.

pkg/provider/azure_instances_v2.go
```diff
@@ -51,15 +54,50 @@ func (az *Cloud) InstanceExists(ctx context.Context, node *v1.Node) (bool, error
...
+				return az.tolerateInstanceNotFoundForNewNode(logger, node), nil
 			}
 
 			logger.Error(err, "failed to get the provider ID by node name", "node", node.Name)
 			return false, err
 		}
 	}
 
-	return az.InstanceExistsByProviderID(ctx, providerID)
+	exists, err := az.InstanceExistsByProviderID(ctx, providerID)
```
nilo19, line 65: The providerID-exist path seems doesn't have UT coverage, could you help add tests?

pkg/provider/azure_instances_v2.go
```diff
@@ -51,15 +54,50 @@ func (az *Cloud) InstanceExists(ctx context.Context, node *v1.Node) (bool, error
...
+// NodeInstanceNotFoundGracePeriodInSeconds (measured from the node's creation
+// timestamp) the node is reported as existing.
+// See https://github.com/kubernetes-sigs/cloud-provider-azure/issues/10604.
+func (az *Cloud) tolerateInstanceNotFoundForNewNode(logger logr.Logger, node *v1.Node) bool {
+	gracePeriod := time.Duration(az.NodeInstanceNotFoundGracePeriodInSeconds) * time.Second
+	if gracePeriod <= 0 {
+		return false
+	}
+
+	if node.CreationTimestamp.IsZero() {
```
nilo19, line 89: It would be cheap to add UT cover for this if block as well.

cmd/cloud-controller-manager/app/config/config.go
```diff
@@ -63,6 +63,12 @@ type Config struct {
 
 	// Node filtering configuration
 	NodeFilterRequirements string
+
+	// NodeInstanceNotFoundGracePeriodInSeconds is the period, measured from a node's
+	// creation timestamp, during which a node whose backing VM/VMSS instance is not
+	// yet visible in ARM is still reported as existing, preventing premature node
+	// deletion by the cloud-node-lifecycle controller.
+	NodeInstanceNotFoundGracePeriodInSeconds int
```
nilo19: Why a cli config is needed here? I think configure from azure.json would be enough.

pkg/provider/azure.go
```diff
@@ -94,6 +94,16 @@ type Cloud struct {
...
 
+	// NodeInstanceNotFoundGracePeriodInSeconds is the period, measured from a node's
+	// creation timestamp, during which a node whose VM/VMSS instance is not yet
+	// visible in ARM is still reported as existing. This tolerates the delay
+	// between a node registering itself in the cluster and its backing instance
+	// propagating into ARM, preventing the cloud-node-lifecycle controller from
+	// deleting the node prematurely. It is configured via the
+	// --node-instance-not-found-grace-period-in-seconds command-line flag and
+	// defaults to 0 (disabled).
+	NodeInstanceNotFoundGracePeriodInSeconds int
```
nilo19: This is a behavior change and may hide genuine ARM issues where instance not found is returned by mistake. The change is acceptable but worth stating clearly the tradeoff in the comments, as well as in our doc (from documentation branch).

cmd/cloud-controller-manager/app/config/config.go
robbiezhang: cli flag would be easier to be configured via chart, while azure.json is composed and delivered by other services in AKS.

pkg/provider/azure.go
robbiezhang: happy to update the document as well. can you point me to the doc branch?

pkg/provider/azure_instances_v2.go
robbiezhang: UT added: 	t.Run("should report a node as not existing when its creation timestamp is zero", func(t *testing.T) {

pkg/provider/azure_instances_v2.go
robbiezhang: UTs added

pkg/provider/azure_instances_v2.go
robbiezhang: updated

pkg/provider/azure.go
nilo19: https://github.com/kubernetes-sigs/cloud-provider-azure/blob/documentation/content/en/install/configs.md

robbiezhang: #10622

nilo19:
Could you please remove the merge commits? Or the PR will be blocked.

nilo19:
Could you help update the release note + pr description as well? nodeInstanceNotFoundGracePeriodInSeconds not node-instance-not-found-grace-period-in-seconds

robbiezhang:
> Could you please remove the merge commits? Or the PR will be blocked.

removed

robbiezhang:
> Could you help update the release note + pr description as well? nodeInstanceNotFoundGracePeriodInSeconds not node-instance-not-found-grace-period-in-seconds

updated.

PR #10622, opened 2026-06-30
Add node instance grace period flag to azure-ccm.md

### LINKED ISSUE #10604: New Node is Deleted because the VM/VMSS instance is not shown up in ARM

robbiezhang:
Added new flag '--node-instance-not-found-grace-period-in-seconds' with details on its usage and default value.

#### What type of PR is this?

/kind documentation

Update the ccm document for the new flag: node-instance-not-found-grace-period-in-seconds

Fixes #10604

functional change is in #10605
```release-note
NONE
```

### DIFF
--- a/content/en/install/azure-ccm.md
+++ b/content/en/install/azure-ccm.md
@@ -49,6 +49,7 @@ azure-cloud-controller-manager should be run as Deployment with multiple replica
 |`--cloud-config`|/etc/kubernetes/cloud-config/azure.json| Path for [cloud provider config](../configs) |
 |`--controllers`|*,-cloud-node | cloud node controller should be disabled     |
 |`--configure-cloud-routes`| "false" for Azure CNI and "true" for other network plugins| Used for non-AzureCNI clusters               |
+|`--node-instance-not-found-grace-period-in-seconds`| 0 | Grace period in seconds, measured from a node's creation timestamp, during which a node whose backing VM/VMSS instance is not yet visible in ARM is still reported as existing to avoid premature deletion. Defaults to 0 (disabled). |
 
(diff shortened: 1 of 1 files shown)


content/en/install/azure-ccm.md
```diff
@@ -49,6 +49,7 @@ azure-cloud-controller-manager should be run as Deployment with multiple replica
 |`--cloud-config`|/etc/kubernetes/cloud-config/azure.json| Path for [cloud provider config](../configs) |
 |`--controllers`|*,-cloud-node | cloud node controller should be disabled     |
 |`--configure-cloud-routes`| "false" for Azure CNI and "true" for other network plugins| Used for non-AzureCNI clusters               |
+|`--node-instance-not-found-grace-period-in-seconds`| 0 | Grace period in seconds, measured from a node's creation timestamp, during which a node whose backing VM/VMSS instance is not yet visible in ARM is still reported as existing to avoid premature deletion. Defaults to 0 (disabled). |
```
nilo19, line 52: Could you help update the doc here? as well as the PR description.

PR #10627, opened 2026-07-01
chore: fix cves for release-1.36

anndono:
#### What type of PR is this?

/kind cleanup
Bump dependencies and distroless to fix cves

Fixes #

```release-note
NONE
```

```docs

```


health-probe-proxy/go.mod
```diff
@@ -1,6 +1,6 @@
 module health-probe-proxy
 
-go 1.26.0
```
Liunardy, line 3: Why is the patch bumped? Pretty sure 1.25.0 is the minimum.

nilo19: If this is unnecessary how do we improve the skill?

anndono: Looks like my local `GOTOOLCHAIN` was `go1.25.11+auto` so for go v1.26.x it will use the go directive in go.mod. Already reverted the changes.

Issue #10270, opened 2026-05-06
Refine loadBalancerBackendPoolUpdater


nilo19:
#### What would you like to be added:

https://github.com/kubernetes-sigs/cloud-provider-azure/pull/10253

#### Why is this needed:

related: https://github.com/kubernetes-sigs/cloud-provider-azure/issues/10252

nilo19:
related: https://github.com/kubernetes-sigs/cloud-provider-azure/issues/10307

PR #10635, opened 2026-07-03
feat(multi-slb): add retry for local service backend pool updater

Liunardy:
#### What type of PR is this?

/kind feature

Adds bounded retry for retriable errors in the local service backend pool updater. Previously, any error dropped the batch permanently with a failure metric. Non-404 errors also emitted a failure event.

Now errors are classified into three categories. Retriable errors (409 Conflict, 412 Precondition Failed, ThrottleError) are requeued with a per-op retry counter up to `loadBalancerBackendPoolUpdateMaxRetries` (default 3, configurable, 0 disables). ThrottleError with a future RetryAfter parks the group until eligible. Terminal errors (non-retriable: SDK-retried statuses, non-ARM errors, and unrecognized status codes) are reported and dropped as before. Stale 404 errors are dropped silently without event or metric (previously recorded a failure metric).

Events are deduplicated per service per tick, fixing the previous notify() that only reported the first operation in a batch due to an unconditional break. Retried operations are prepended to the queue to preserve chronological diff ordering. Operations are re-checked for relevance (service still exists, LB mapping unchanged) before requeuing, and all operations are dropped silently on shutdown.

Part of #10270

Design doc: #10417
Documentation update for `topics/multislb` #10648

Deferred to follow-up PR (#10730):
- Desired-state operation model (one op per service/pool, eliminating duplicate events for mixed-budget groups)

```release-note
feat(multi-slb): add bounded retry for local service backend pool updates. Retriable errors (409, 412, throttle) are retried up to 3 times by default. Configure with `loadBalancerBackendPoolUpdateMaxRetries` (0 disables). New event `LoadBalancerBackendPoolUpdateRetrying` signals a retry is pending. 404 errors no longer count as failures in metrics.
```

```docs

```

### DIFF
--- a/pkg/provider/azure_local_services.go
+++ b/pkg/provider/azure_local_services.go
@@ -18,11 +18,15 @@ package provider
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"net/http"
+	"slices"
 	"strings"
 	"sync"
 	"time"
 
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/network/armnetwork/v9"
 	v1 "k8s.io/api/core/v1"
 	discovery_v1 "k8s.io/api/discovery/v1"
@@ -34,6 +38,7 @@ import (
 	utilnet "k8s.io/utils/net"
 	"k8s.io/utils/ptr"
 
+	"sigs.k8s.io/cloud-provider-azure/pkg/azclient/policy/retryrepectthrottled"
 	"sigs.k8s.io/cloud-provider-azure/pkg/consts"
 	"sigs.k8s.io/cloud-provider-azure/pkg/log"
 	"sigs.k8s.io/cloud-provider-azure/pkg/metrics"
@@ -65,6 +70,8 @@ type loadBalancerBackendPoolUpdateOperation struct {
 	backendPoolName  string
 	kind             consts.LoadBalancerBackendPoolUpdateOperation
 	nodeIPs          []string
+	retryAttempts    int
+	nextEligibleAt   time.Time
 }
 
 func (op *loadBalancerBackendPoolUpdateOperation) wait() batchOperationResult {
@@ -203,11 +210,37 @@ func (updater *loadBalancerBackendPoolUpdater) groupOperations(ctx context.Conte
 			continue
 		}
 
+		// Check if the Service still exists in the informer. Catches the case where
+		// EnsureLoadBalancerDeleted failed partway, leaving a stale map entry.
+		if _, exists, err := updater.az.getLatestService(lbOp.serviceName, false); err == nil && !exists {
+			logger.V(4).Info("Service not found in informer, skip the operation", "service", lbOp.serviceName)
+			continue
+		}
+
 		key := fmt.Sprintf("%s:%s", lbOp.loadBalancerName, lbOp.backendPoolName)
 		groups[key] = append(groups[key], op)
 	}
 
 	return groups
+}
+
+// hasParkedOperations returns true if any operation has a nextEligibleAt in the future.
+func (updater *loadBalancerBackendPoolUpdater) hasParkedOperations(ops []batchOperation) bool {
+	now := time.Now()
+	for _, op := range ops {
+		lbOp := op.(*loadBalancerBackendPoolUpdateOperation)
+		if lbOp.nextEligibleAt.After(now) {
+			return true
+		}
+	}
+	return false
+}
+
+// requeueOperations prepends operations back to the front of the queue.
+func (updater *loadBalancerBackendPoolUpdater) requeueOperations(ops []batchOperation) {
+	updater.lock.Lock()
+	defer updater.lock.Unlock()
+	updater.operations = append(ops, updater.operations...)
 }
 
 // process processes all operations in the loadBalancerBackendPoolUpdater.
@@ -242,9 +275,20 @@ func (updater *loadBalancerBackendPoolUpdater) process(ctx context.Context) {
 	groups := updater.groupOperations(ctx, ops)
 
 	for key, ops := range groups {
+		if ctx.Err() != nil {
+			break
+		}
+
 		parts := strings.Split(key, ":")
 		lbName, poolName := parts[0], parts[1]
 		operationName := fmt.Sprintf("%s/%s", lbName, poolName)
+
+		// If any op in the group has nextEligibleAt in the future,
+		// requeue the entire group without making ARM calls or emitting events.
+		if updater.hasParkedOperations(ops) {
+			updater.requeueOperations(ops)
+			continue
+		}
 
 		mc := metrics.NewMetricContext(
 			"services_local",      // prefix name, differ from "services" in main reconciliation loop
@@ -256,8 +300,7 @@ func (updater *loadBalancerBackendPoolUpdater) process(ctx context.Context) {
 
 		bp, err := updater.az.NetworkClientFactory.GetBackendAddressPoolClient().Get(ctx, updater.az.ResourceGroup, lbName, poolName)
 		if err != nil {
-			mc.ObserveOperationWithResult(false)
-			updater.processError(err, operationName, ops...)
+			updater.handleGroupError(ctx, mc, err, operationName, ops)
 			continue
 		}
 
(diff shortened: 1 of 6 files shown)


pkg/provider/azure_local_services.go
```diff
@@ -281,39 +317,132 @@ func (updater *loadBalancerBackendPoolUpdater) process(ctx context.Context) {
...
+			return errClassTerminal
+		}
+	}
+
+	return errClassTerminal
+}
+
+// handleGroupError classifies the error and either drops, retries, or fails the group.
+// For user-facing documentation of retry behavior, see
+// https://cloud-provider-azure.sigs.k8s.io/topics/multislb/
```
nilo19: Could link to "retry behavior" title.

pkg/provider/azure_local_services.go
```diff
@@ -281,39 +317,132 @@ func (updater *loadBalancerBackendPoolUpdater) process(ctx context.Context) {
...
+func (updater *loadBalancerBackendPoolUpdater) handleGroupError(ctx context.Context, mc *metrics.MetricContext, err error, operationName string, ops []batchOperation) {
+	logger := log.FromContextOrBackground(ctx).WithName("loadBalancerBackendPoolUpdater.handleGroupError")
+
+	// On shutdown, drop silently.
+	if ctx.Err() != nil {
```
nilo19, line 377: add a log when ctx is canceled?

pkg/provider/azure_local_services.go
```diff
@@ -281,39 +317,132 @@ func (updater *loadBalancerBackendPoolUpdater) process(ctx context.Context) {
...
+		var throttleErr *retryrepectthrottled.ThrottleError
+		errors.As(err, &throttleErr)
+
+		var opsToRequeue []batchOperation
+		var exhaustedOps []batchOperation
+		for _, op := range ops {
+			lbOp := op.(*loadBalancerBackendPoolUpdateOperation)
+
+			// Re-check relevance before requeue.
+			si, found := updater.az.getLocalServiceInfo(strings.ToLower(lbOp.serviceName))
```
nilo19, line 407: How do we make sure the service info map is not stale here? For example, during a failed svc deletion, EnsureLoadBalancerDeleted returns early before clearing the map, so an operation for a Service already absent from the informer can still call ARM and requeue on 409/412/throttle.

pkg/provider/azure_local_services.go
```diff
@@ -281,39 +317,132 @@ func (updater *loadBalancerBackendPoolUpdater) process(ctx context.Context) {
...
+		}
 
+		// TODO: The desired-state model (one op per service/pool) would eliminate
+		// the case where a single Service receives both Failed and Retrying events
+		// in the same tick.
+
+		if len(exhaustedOps) > 0 {
+			mc.ObserveOperationWithResult(false)
+			updater.notify(ctx, v1.EventTypeWarning, consts.LoadBalancerBackendPoolUpdateFailed,
+				fmt.Sprintf("Backend pool update failed after %d retries: %v. To retrigger, cause an endpoint change for the Service (e.g., restart or scale a backing pod).", *updater.az.LoadBalancerBackendPoolUpdateMaxRetries, err), exhaustedOps...)
```
nilo19: The error message is confusing. To trigger the operation we need to change the node/ipSet. To scale/restart pods on a same node won't affect the updater. Please check the documentation PR as well.

pkg/provider/azure_local_services.go
```diff
@@ -281,39 +317,132 @@ func (updater *loadBalancerBackendPoolUpdater) process(ctx context.Context) {
...
+	}
+
+	var respErr *azcore.ResponseError
+	if errors.As(err, &respErr) {
+		if respErr.StatusCode == http.StatusConflict || respErr.StatusCode == http.StatusPreconditionFailed {
+			return errClassRetriable
+		}
+		// SDK-retriable statuses are terminal because the SDK already retried.
+		if slices.Contains(retryrepectthrottled.GetRetriableStatusCode(), respErr.StatusCode) {
+			return errClassTerminal
```
nilo19: Will this confuse the user? Saying that the SDK retry budget was exhausted instead and expose this information to the user would be more helpful?

pkg/provider/azure_local_services.go
Liunardy: I think we can use getLatestService in groupOperations after checking getLocalServiceInfo and service lb to check if service still exists in the informer.

Liunardy: Added the check.

pkg/provider/azure_local_services.go
Liunardy: Created separate err class for SDK retried with "Backend pool update failed (SDK retries exhausted):" event.

pkg/provider/azure_local_services.go
Liunardy: Updated to rely on the main reconcile path "cause a reconciliation for the Service (e.g., add or update any annotation on the Service)." for the guidance here. Also updated the doc PR for the node IP set change.

pkg/provider/azure_local_services.go
Liunardy: Added.

pkg/provider/azure_local_services.go
Liunardy: Added the expected anchor.

pkg/provider/azure_local_services.go
```diff
@@ -203,13 +210,39 @@ func (updater *loadBalancerBackendPoolUpdater) groupOperations(ctx context.Conte
 			continue
 		}
 
+		// Check if the Service still exists in the informer. Catches the case where
+		// EnsureLoadBalancerDeleted failed partway, leaving a stale map entry.
+		if _, exists, err := updater.az.getLatestService(lbOp.serviceName, false); err == nil && !exists {
```
nilo19: I'm afraid the services being deleted still can be found in the informer, because the cleanup hooked by the service finalizer hasn't been finished. Have you tested this to be effective? To me a clearer way is to check if a svc has a non nil deletion timestamp. Should check both here and ~L407-412.

Liunardy: Updated to check if service is gone or has deletion timestamp. Tested by building an image that return error on EnsureLoadBalancerDeleted, groupOperations skips it correctly. Updated the handleGroupErr requeue relevance checks too.

pkg/provider/azure_local_services.go
nilo19, line 69: I'm considering the following scenario, not sure if it's a potential bug because I haven't tested.
  1. An operation is queued for old Service A.
  2. Azure returns a conflict, so the operation is retained for retry.
  3. Service A is deleted.
  4. New Service B is created with the same name and load balancer.
  5. The retry checks the name, load balancer, and deletion status, all match B, so it mistakenly applies A’s operation to B.
If this is exactly a bug, maybe we can fix by using service uid instead of name to identify a service. We can target it in a following PR.

pkg/provider/azure_local_services.go
```diff
@@ -282,39 +324,151 @@ func (updater *loadBalancerBackendPoolUpdater) process(ctx context.Context) {
...
+
+		var opsToRequeue []batchOperation
+		var exhaustedOps []batchOperation
+		for _, op := range ops {
+			lbOp := op.(*loadBalancerBackendPoolUpdateOperation)
+
+			// Re-check relevance before requeue.
+			si, found := updater.az.getLocalServiceInfo(strings.ToLower(lbOp.serviceName))
+			if !found {
+				logger.V(4).Info("Service is not a local service, dropped before requeue",
```
nilo19, line 409: If the service has been deleted, do we still hit this not found block? If so do we need to update the error message?

nilo19:
Not directly related to this change: have we tested converting an existing local service to cluster? In this case will the local svc pool be removed correctly?

pkg/provider/azure_local_services.go
```diff
@@ -282,39 +324,151 @@ func (updater *loadBalancerBackendPoolUpdater) process(ctx context.Context) {
...
+	switch class {
+	case errClassStale:
+		logger.V(4).Info("Backend pool not found, skipped update", "operation", operationName)
 
+	case errClassSDKExhausted:
+		mc.ObserveOperationWithResult(false)
+		updater.notify(ctx, v1.EventTypeWarning, consts.LoadBalancerBackendPoolUpdateFailed,
+			fmt.Sprintf("Backend pool update failed (SDK retries exhausted): %v.", err), ops...)
+
+	case errClassTerminal:
```
nilo19, line 392: If a terminal error is returned but the service was deleted before that, should we ignore the error?

PR #10641, opened 2026-07-03
ServiceGateway (SLB) common-ground provider integration

georgeedward2000:
#### What type of PR is this?

/kind feature

Wires ServiceGateway (Service Load Balancer) support into the shared provider layer. When
`ServiceGatewayEnabled` is set, load balancers are managed per-service (PodIP backend pools)
via an async diff tracker instead of the synchronous reconcile path.

- Route `GetLoadBalancer`/`EnsureLoadBalancer`/`UpdateLoadBalancer`/`EnsureLoadBalancerDeleted`
  through the diff tracker; per-service LB/PIP naming and `Service`/`StandardV2` SKUs.
- Per-service backend pools and endpoint-slice → diff-tracker endpoint sync.
- Inbound-config validation (rejects internal, dual-stack, named/duplicate ports, non-TCP/UDP).
- Adds the `servicegateway/difftracker` package as the API surface the integration compiles
  against; the engine implementation is delivered separately.

All new behavior is gated on `ServiceGatewayEnabled` (default false), so existing flows are unchanged.

Fixes #

`pkg/provider/servicegateway/difftracker` is the API surface only (no-op stubs); the engine lands
in a follow-up. Review focuses on the common-ground hooks in the shared provider files.

```release-note
NONE

### CODE REVIEW

File: pkg/provider/azure.go [hunk 1]
```diff
@@ -505,6 +523,40 @@ func (az *Cloud) InitializeCloudFromConfig(ctx context.Context, config *azurecon
...
+			dtConfig := difftracker.Config{
+				SubscriptionID:             az.SubscriptionID,
+				ResourceGroup:              az.ResourceGroup,
+				Location:                   az.Location,
+				VNetName:                   az.VnetName,
+				VNetResourceGroup:          az.VnetResourceGroup,
+				ServiceGatewayResourceName: sgwName,
+				ServiceGatewayID:           az.GetServiceGatewayID(),
+			}
+			az.diffTracker, err = difftracker.InitializeFromCluster(ctx, dtConfig, az.NetworkClientFactory, az.KubeClient)
```
david-kow, line 548: is kubeclient already initialized here?

pkg/provider/azure_test.go
```diff
@@ -2665,6 +2665,68 @@ func TestInitializeCloudFromConfig(t *testing.T) {
...
+		ctrl := gomock.NewController(t)
+		defer ctrl.Finish()
+		az := GetTestCloud(ctrl)
+		zoneMock := az.zoneRepo.(*zone.MockRepository)
+		zoneMock.EXPECT().ListZones(gomock.Any()).Return(map[string][]string{"eastus": {"1", "2", "3"}}, nil).AnyTimes()
+
+		// serviceGatewayEnabled is only effective with a service SKU and podIP backend pool. The raw
+		// flag alone (here a standard SKU / NodeIPConfiguration pool) must be refined to false from the
+		// incoming config. callFromCCM=false keeps the assertion on the config-derived flag.
+		azureconfig := providerconfig.Config{
```
david-kow, line 2854: shouldn't this return error instead?

georgeedward2000: Yes, silently disabling an explicitly requested ServiceGateway hides an invalid configuration. I changed this so serviceGatewayEnabled=true, loadBalancerSku=service, and backendPoolType=podIP must all be configured together. Any incomplete combination now returns an error. Added coverage for all six incomplete combinations.

pkg/provider/azure.go
georgeedward2000: Good catch. It was not initialized at this point. NewCloud assigned az.KubeClient only after InitializeCloudFromConfig returned, while InitializeFromCluster runs inside it. The stub masked this because it ignores the client, but the real engine rejects a nil KubeClient. I moved the client initialization before InitializeCloudFromConfig.

pkg/provider/azure.go
```diff
@@ -505,6 +520,40 @@ func (az *Cloud) InitializeCloudFromConfig(ctx context.Context, config *azurecon
...
+			if err != nil {
+				klog.Errorf("InitializeCloudFromConfig: failed to initialize difftracker: %s", err.Error())
+				return err
+			}
+
+			// Publish the deps the difftracker needs (event recorder and shared EndpointSlice cache).
+			az.diffTracker.SetEventRecorder(az.eventRecorder)
```
david-kow: doesn't this have the same issue as the kubeclient?

pkg/provider/azure_loadbalancer.go
```diff
@@ -188,7 +205,7 @@ func (az *Cloud) reconcileService(ctx context.Context, clusterName string, servi
 
 	lbName := strings.ToLower(ptr.Deref(lb.Name, ""))
 	key := strings.ToLower(getServiceName(service))
-	if az.UseMultipleStandardLoadBalancers() && isLocalService(service) {
+	if az.UseMultipleStandardLoadBalancers() && (isLocalService(service) || az.ServiceGatewayEnabled) {
 		az.localServiceNameToServiceInfoMap.Store(key, newServiceInfo(getServiceIPFamily(service), lbName))
```
david-kow, line 202: are we using that map?

pkg/provider/azure_local_services.go
```diff
@@ -367,7 +386,7 @@ func (az *Cloud) setUpEndpointSlicesInformer(informerFactory informers.SharedInf
 
 				key := strings.ToLower(fmt.Sprintf("%s/%s", newES.Namespace, svcName))
 				si, found := az.getLocalServiceInfo(key)
-				if !found {
+				if !found && !az.ServiceGatewayEnabled {
```
david-kow: I think thath if the sgw is enabled but the info was not found, si will be nil but and in 393 we will panic

pkg/provider/azure.go
```diff
@@ -735,14 +797,28 @@ func (az *Cloud) SetInformers(informerFactory informers.SharedInformerFactory) {
...
+	// Publish the Service lister so the difftracker can resolve UIDs via cached reads.
+	if az.diffTracker != nil {
+		az.diffTracker.SetServiceLister(az.serviceLister)
+	}
+
 	az.setUpEndpointSlicesInformer(informerFactory)
+
+	if az.ServiceGatewayEnabled && az.diffTracker != nil {
+		az.diffTracker.SetUpPodInformer()
+	}
```
david-kow, line 829: we should check for sgw enabled in both if or don't check it at all. also, is there a dependency between those calls? can we wrap them into one if?

pkg/provider/azure.go
```diff
@@ -708,12 +757,25 @@ func (az *Cloud) SetInformers(informerFactory informers.SharedInformerFactory) {
...
 			prevNode := prev.(*v1.Node)
 			newNode := obj.(*v1.Node)
 			az.updateNodeCaches(prevNode, newNode)
 			az.updateNodeTaint(newNode)
+			if az.ServiceGatewayEnabled && az.diffTracker != nil {
+				// A node IP change fires no slice event; move resident pods off the old IP.
+				oldIPs := getNodePrivateIPAddresses(prevNode)
+				newIPs := getNodePrivateIPAddresses(newNode)
+				oldSet, newSet := utilsets.NewString(oldIPs...), utilsets.NewString(newIPs...)
+				if oldSet.Difference(newSet).Len() != 0 || newSet.Difference(oldSet).Len() != 0 {
```
david-kow: I think the underlying set has a single func that does that, we probably should expose it and use it to avoid two enumerations

pkg/provider/azure_loadbalancer.go
```diff
@@ -279,6 +296,55 @@ func (az *Cloud) EnsureLoadBalancer(ctx context.Context, clusterName string, ser
...
+			az.diffTracker.UpdateService(config)
+			// A type flip during an in-flight delete recreates without a slice event; re-seed endpoints.
+			if az.diffTracker.IsServiceRecreating(serviceUID) {
+				az.seedInboundEndpointsFromCache(serviceUID)
+			}
+		} else {
+			az.diffTracker.AddService(config)
+			// Seed current endpoints so a re-registered service without a slice event still gets its backend.
+			az.seedInboundEndpointsFromCache(serviceUID)
+		}
```
david-kow, line 301: why we need non-difftracker struct?

pkg/provider/azure_loadbalancer.go
```diff
@@ -983,7 +1065,26 @@ func (az *Cloud) getServiceLoadBalancer(
 			Location:   &az.Location,
 			Properties: &armnetwork.LoadBalancerPropertiesFormat{},
 		}
-		if az.UseStandardLoadBalancer() {
+		if az.ServiceGatewayEnabled {
```
david-kow: I'm not sure I get why we need default lb - there is no default lb with sgw. is that needed to satisfy some conditions from this functions callers?

pkg/provider/azure_loadbalancer.go
david-kow: do we need to check that with sgw enabled?

pkg/provider/azure_loadbalancer_backendpool.go
```diff
@@ -790,3 +799,96 @@ func removeNodeIPAddressesFromBackendPool(
...
+
+func newBackendPoolTypePodIP(c *Cloud) BackendPool {
+	return &backendPoolTypePodIP{c}
+}
+
+func (bpi *backendPoolTypePodIP) EnsureHostsInPool(_ context.Context, _ *v1.Service, _ []*v1.Node, _ string, _ string, _ string, _ string, _ *armnetwork.BackendAddressPool) error {
+	return nil
+}
+
+func (bpi *backendPoolTypePodIP) GetBackendPrivateIPs(_ context.Context, _ string, service *v1.Service, lb *armnetwork.LoadBalancer) ([]string, []string) {
```
david-kow: where this is used?

pkg/provider/azure_local_services.go
```diff
@@ -413,6 +432,32 @@ func (az *Cloud) setUpEndpointSlicesInformer(informerFactory informers.SharedInf
 					}
 					az.applyIPChangesAmongLocalServiceBackendPoolsByIPFamily(lbName, key, currentIPsInBackendPools, currentIPs)
 				}
+
+				if az.ServiceGatewayEnabled {
```
david-kow: do we need anything from the code above? why don't we start with this condition and bail after we are done?

pkg/provider/azure.go
georgeedward2000: Yes. The recorder is created later in Cloud.Initialize, so this path was publishing a nil recorder. I moved SetEventRecorder to run immediately after recorder creation and before SetInformers starts the SGW pod informer. I also added regression coverage to verify the initialization order.

pkg/provider/azure.go
georgeedward2000: Yes, there is an ordering dependency: the Service lister must be published before the engine-owned pod informer starts. I wrapped both calls under a single SGW/difftracker check and kept the lister assignment first to preserve the required initialization order.

pkg/provider/azure.go
georgeedward2000: Good point. IgnoreCaseSet already exposes Equals, so I switched this comparison to use that directly and updated the wrapper to delegate to the underlying set equality operation. This avoids creating and scanning two difference sets while keeping the intent clearer.

pkg/provider/azure_loadbalancer.go
georgeedward2000: No, SGW does not use this map through this path. UseMultipleStandardLoadBalancers() cannot be true when ServiceGateway is enabled because the configurations are mutually exclusive. I removed the redundant SGW condition.

pkg/provider/azure_loadbalancer.go
georgeedward2000: This helper intentionally remains on Cloud because it adapts provider-owned EndpointSlice and node-IP caches into the difftracker.UpdateEndpoints API. The difftracker owns the resulting state, but it does not own the provider's live node-IP cache that is required to resolve each pod IP to its same-family node location.

The explicit seeding is required for Service type flips and re-registration scenarios where no EndpointSlice event is generated. Without it, the recreated load balancer ends up with an empty backend pool.

The configuration itself is already represented as a difftracker-owned ServiceConfig.

pkg/provider/azure_local_services.go
georgeedward2000: You're right. SGW does not require localServiceInfo, so si can be nil in this path. I moved the dereference into the non-SGW backend pool block, where localServiceInfo is actually needed, and added regression coverage for this case.

pkg/provider/azure_loadbalancer.go
georgeedward2000: Correct, SGW does not have a default LB. This shared function still returns a placeholder to preserve its legacy contract, but SGW does not use it for provisioning. I removed the SGW-specific SKU, scope, and backend-pool setup and restored the original Standard Load Balancer-only behavior.

pkg/provider/azure_loadbalancer.go
georgeedward2000: No, SGW does not require this lock because it only queues work into difftracker, which manages its own synchronization. I restricted the Azure resource lock to the legacy non-SGW reconciliation path and added regression coverage to ensure SGW continues to operate correctly without it.

pkg/provider/azure_loadbalancer_backendpool.go
georgeedward2000: This adapter is no longer used. SGW bypasses the legacy load-balancer/backend-pool reconciliation path, and difftracker owns PodIP backend pools. I removed the PodIP BackendPool implementation and its tests; the PodIP configuration value remains as the SGW mode selector.

pkg/provider/azure_local_services.go
georgeedward2000: Agreed. SGW does not require the local-service lookup or the legacy node-IP aggregation in the preceding logic. I moved the SGW path immediately after the cache update and return as soon as SGW processing completes.

I also added coverage to verify that SGW updates are tracked by Service UID and do not depend on the legacy service-name label.

pkg/provider/azure_loadbalancer.go
david-kow: but why it's needed? why diff tracker needs additional input?

pkg/consts/consts.go
```diff
@@ -243,6 +243,12 @@ const (
 	LoadBalancerSKUStandard = "standard"
 	// LoadBalancerSKUService is the load balancer service SKU
 	LoadBalancerSKUService = "service"
+	// LoadBalancerSKUNameService is the case-sensitive ARM SKU name ("Service") for the
+	// ServiceGateway inbound load balancer.
+	LoadBalancerSKUNameService = "Service"
```
wonderyl: nit: "Name" is very generic and doesn't really distinguish it with `LoadBalancerSKUService`. How about "LoadBalancerARMSKUService "

pkg/provider/azure.go
```diff
@@ -172,16 +175,20 @@ func NewCloud(ctx context.Context, clientBuilder cloudprovider.ControllerClientB
...
+	}
+
 	err := az.InitializeCloudFromConfig(ctx, config, false, callFromCCM)
 	if err != nil {
 		return nil, err
 	}
 
 	az.ipv6DualStackEnabled = true
 
-	if clientBuilder != nil {
```
nilo19, line 182: Any reason we change the logic to not init az.KubeClient if service gateway is not enabled?

wonderyl, line 182: Will removing this affect the code path that doesn't use service gateway?

pkg/provider/azure.go
```diff
@@ -355,6 +349,28 @@ func (az *Cloud) InitializeCloudFromConfig(ctx context.Context, config *azurecon
 		return err
 	}
 
+	serviceGatewayConfigured := config.ServiceGatewayEnabled ||
```
nilo19, line 355: Why do we change this from `if az.ServiceGatewayEnabled && az.IsLBBackendPoolTypePodIPAndUseServiceLoadBalancer()`?

pkg/provider/azure.go
```diff
@@ -708,12 +765,25 @@ func (az *Cloud) SetInformers(informerFactory informers.SharedInformerFactory) {
 			node := obj.(*v1.Node)
 			az.updateNodeCaches(nil, node)
 			az.updateNodeTaint(node)
+			if az.ServiceGatewayEnabled && az.diffTracker != nil {
```
nilo19: How would az.ServiceGatewayEnabled=true but az.diffTracker=nil? In L528 you don't set az.diffTracker=nil check.

nilo19: same below.

pkg/provider/azure_loadbalancer.go
```diff
@@ -109,10 +110,22 @@ func (az *Cloud) GetLoadBalancer(ctx context.Context, clusterName string, servic
 		return status, existsLb || az.existsPip(ctx, clusterName, service), err
 	}
 
-	flippedService := flipServiceInternalAnnotation(service)
-	_, _, status, _, existsLb, _, err = az.getServiceLoadBalancer(ctx, flippedService, clusterName, nil, false, existingLBs)
-	if err != nil || existsLb {
-		return status, existsLb || az.existsPip(ctx, clusterName, service), err
+	// ServiceGateway uses a single per-service LB, so skip the flipped-service check.
```
nilo19: If we don't need the service reconciliation loop beause we have our own loop, why not just bypass every function in the very beginning?
function list: https://github.com/kubernetes/cloud-provider/blob/master/cloud.go#L138

nilo19: Or, maybe a better option, build a new controller other than "service controller", and register it here? https://github.com/kubernetes-sigs/cloud-provider-azure/blob/e989199d59979119d42927600b577ada8908cc89/cmd/cloud-controller-manager/app/controllermanager.go#L490

pkg/provider/azure_local_services.go
```diff
@@ -350,11 +350,19 @@ func (az *Cloud) setUpEndpointSlicesInformer(informerFactory informers.SharedInf
 		cache.ResourceEventHandlerFuncs{
 			AddFunc: func(obj interface{}) {
 				es := obj.(*discovery_v1.EndpointSlice)
+				if az.ServiceGatewayEnabled {
```
nilo19: Please do not bypass each and every function here, you may just disable the informer when the service gateway is enabled. Or, consider create a new controller option I mentioned above.

nilo19 (dismissed):
Consider add e2e test coverages.

pkg/consts/consts.go
georgeedward2000: Agreed. Renamed it to LoadBalancerARMSKUService to distinguish the case-sensitive ARM SKU value from LoadBalancerSKUService.

pkg/provider/azure.go
georgeedward2000: ServiceGateway needs the Kubernetes client before InitializeCloudFromConfig because difftracker is initialized there. I restored the original fallback initialization for non-ServiceGateway paths while ensuring the client is created only once.

georgeedward2000: I restored the original post-configuration fallback, while ServiceGateway initializes the client earlier because difftracker requires it during InitializeCloudFromConfig.

pkg/provider/azure.go
georgeedward2000: These three settings define a single atomic ServiceGateway mode. The previous condition silently disabled ServiceGatewayEnabled for incomplete configurations, potentially leaving the Service SKU or PodIP backend configured while legacy paths continued to run. This now fails initialization explicitly unless all three settings are configured together.

pkg/provider/azure.go
georgeedward2000: You’re right. Successful ServiceGateway CCM initialization always assigns diffTracker; otherwise, initialization returns an error. I removed the redundant nil checks consistently across the provider paths.

pkg/provider/azure_loadbalancer.go
georgeedward2000: Good point. The difftracker has its own asynchronous processing loop, but it does not independently watch the Service lifecycle. EnsureLoadBalancer and EnsureLoadBalancerDeleted are the ingestion points for Service create, update, and delete events.

Bypassing these methods would prevent the engine from receiving those events, and returning exists=false would stop the service controller from invoking deletion, potentially leaving the ServiceGateway finalizer stranded. A separate controller would also need to duplicate the existing workqueue, retry logic, finalizer handling, status updates, and event processing.

I did move the ServiceGateway handling to the beginning of GetLoadBalancer, so it now serves state directly from difftracker and avoids all legacy LB/PIP API calls. The legacy path remains unchanged.

pkg/provider/azure_local_services.go
georgeedward2000: Disabling the EndpointSlice informer entirely would prevent difftracker from receiving backend endpoint changes. Instead, I separated the paths at registration: ServiceGateway now registers only its forwarding handlers, while non-ServiceGateway mode registers only the existing legacy local-service handlers. This removes the per-callback branching without requiring a separate controller.

pkg/provider/azure_loadbalancer.go
georgeedward2000: I looked at two approaches:

1. LoadBalancer implementation only

Keep the current SGW initialization and Node/EndpointSlice/Pod informer setup as-is. Add a dedicated difftracker.LoadBalancer implementation and return it from Cloud.LoadBalancer() when SGW is enabled.

2. Move SGW runtime ownership into the SGW package

Move DiffTracker initialization, informer registration, and lifecycle management into the SGW package as well.

This would not replace or reimplement Kubernetes' Service controller, and it would not introduce another Service-controller goroutine. The existing servicecontroller.New(...) and startup flow would remain unchanged. We would simply plug our SGW-specific LoadBalancer implementation into the existing controller.

david-kow: I think the second option is more in the spirit of your suggestion @nilo19. @wonderyl - any thoughts on this?

david-kow:
> Consider add e2e test coverages.

@nilo19 What kind of tests you have in mind? I believe current e2e tests cover non-sgw scenarios and for sgw scenarios there is no behavior to be tested. E2E tests for sgw will be introduced in subsequent prs (when the behavior to be tested is there).

pkg/provider/azure_loadbalancer.go
wonderyl: IIUC, you can still run a second ServiceController, and only does the work for ServiceLB. And let the original ServiceController respond to other LBs. In this case you can have most of the logic decoupled. The only downside is you have more watchers, but the very head is pretty small IMHO.

PR #10668, opened 2026-07-08
fix: match ACR regional (geo-replica) login server endpoints in credential provider

### LINKED ISSUE #10667: ACR credential provider does not match regional (geo-replica) login server endpoints
#### What happened:

The ACR credential provider (`pkg/credentialprovider/azure_credentials.go`) does not recognize Azure Container Registry **regional endpoints** — the per-geo-replica login servers of the form `myregistry.<region>.geo.azurecr.io` (for example, `myregistry.eastus2.geo.azurecr.io`).

Regional endpoints are login servers for a specific geo-replica. They let clients pull an image from a chosen geo-replica for predictable routing, in-region affinity, and client-side failover, and they can be referenced directly in image references and Kubernetes deployment manifests. See [How regional endpoints work](https://learn.microsoft.com/en-us/azure/container-registry/container-registry-geo-replication#how-regional-endpoints-work).

On the provider's primary matching path, `parseACRLoginServerFromImage` does not match a regional host via `acrRE` (see Root cause) and returns an empty login server. See the **Update** below for how this actually manifests at AKS runtime.

#### Update (2026-07-09): empirical clarification of scope

Live testing on AKS refined the picture:

- Regional-endpoint pulls **do** fail on a default AKS cluster (`401` / ImagePullBackOff) — reconfirmed on a fresh cluster.
- The gate that actually causes the failure is the **node kubelet `matchImages` config** (managed by AgentBaker), not this provider. The regional host is 5 dot-parts and does not match the 3-part `*.azurecr.io`, so kubelet never invokes this credential provider at all. That node-side fix is [Azure/AgentBaker#8862](https://github.com/Azure/AgentBaker/issues/8862).
- Once kubelet does invoke the provider, the **current** provider already resolves the regional host at runtime: `parseACRLoginServerFromImage` falls through `acrRE` into a secondary branch that uses `environment.ContainerRegistryDNSSuffix` (`"azurecr.io"` on public cloud) with a substring match, which happens to catch the regional host. Verified: with only the `matchImages` fix and the **stock** provider binary, a fresh regional pull succeeds.

So this change is a **correctness/robustness improvement to the primary `acrRE` path** — regional hosts match explicitly instead of relying on the imprecise sovereign-cloud substring fallback (which also loosely matches strings like `xazurecr.io`) — plus regression tests. It is not, by itself, what unblocks AKS regional pulls. The "returns empty → anonymous → 401" phrasing in the sections below describes the unit-level (`environment == nil`) path, not the AKS runtime path.

#### Root cause:

The `acrRE` regular expression only matches a single DNS label before the ACR suffix:

```go
acrRE = regexp.MustCompile(`^[a-zA-Z0-9]+\.(azurecr\.io|azurecr\.cn|azurecr\.de|azurecr\.us)`)
```

The `[a-zA-Z0-9]+` sub-pattern matches exactly one DNS label — a single run of alphanumeric characters with no dot. Because it is anchored at `^` and immediately followed by `\.(azurecr\....)`, the expression only accepts a **single-label** registry name directly in front of the suffix; it cannot match a host that has any additional dot-separated labels before the suffix. A regional endpoint such as `myregistry.eastus2.geo.azurecr.io` carries two extra labels (`eastus2` and `geo`) ahead of the suffix, so the anchored single-label pattern does not match. `acrRE.FindString(...)` returns `""` in `parseACRLoginServerFromImage`, and the provider treats the image as non-ACR.

This was validated with actual AKS testing against regional endpoints, not just observed in code. On a real AKS cluster with a correct `AcrPull` role assignment, a pod that references a regional endpoint host fails to pull with `401 Unauthorized` (ImagePullBackOff): the provider returns no credential for the regional host, so the pull falls back to anonymous. At the unit level, `parseACRLoginServerFromImage("myregistry.eastus2.geo.azurecr.io/app:v1")` returns `""` instead of the regional host, confirming the same root cause.

The credential provider should recognize regional login server endpoints (`myregistry.<region>.geo.azurecr.io`) as ACR registries and return credentials for them, exactly as it does for the global endpoint (`myregistry.azurecr.io`).

Only the **login server** forms (global and regional `.geo`) should match. **Dedicated data endpoints** (`myregistry.<region>.data.azurecr.io`) should not be matched: they are layer-blob-download redirect targets (a registry issues a 307 redirect to them during a pull) and never appear as the registry host in an image reference, so they never require a credential lookup. See the [endpoint reference](https://learn.microsoft.com/en-us/azure/container-registry/container-registry-endpoint-reference#endpoint-types) for the distinction between regional `.geo` login servers and dedicated `.data` endpoints.

1. Create a Premium ACR with regional endpoints enabled (`az acr update -n myregistry --regional-endpoints enabled`).
2. Reference an image via the regional endpoint, for example `myregistry.eastus2.geo.azurecr.io/myapp:v1`.
3. Attempt a pull on a node that uses the ACR credential provider. The provider returns no credential, so the pull is anonymous and returns `401 Unauthorized`.

Unit-level equivalent: `parseACRLoginServerFromImage("myregistry.eastus2.geo.azurecr.io/myapp:v1")` returns `""` instead of `myregistry.eastus2.geo.azurecr.io`.

- The fix is limited to the `acrRE` pattern. `GetCredentials` already inserts an exact-host auth entry for the parsed login server, so the existing `*.azurecr.*` response wildcards need no change.
- For end-to-end pulls on AKS specifically, the node's kubelet credential-provider `matchImages` list (managed by AgentBaker, a separate repo) must also include the regional host pattern so kubelet invokes this provider. That companion node-side change is tracked in **Azure/AgentBaker#8862**. This issue tracks the provider-side matching bug, which is valid on its own.

#### Environment:

- cloud-provider-azure: `master` (ACR credential provider component)
- Cloud provider or hardware configuration: Azure — Azure Container Registry (Premium tier with regional endpoints enabled)

johnsonshi:
#### What type of PR is this?

/kind bug

The ACR credential provider did not recognize Azure Container Registry **regional endpoints** — the per-geo-replica login servers of the form `myregistry.<region>.geo.azurecr.io` (for example, `myregistry.eastus2.geo.azurecr.io`).

Regional endpoints are login servers for a specific geo-replica. They let clients pull an image from a chosen geo-replica for predictable routing, in-region affinity, and client-side failover, and they can be referenced directly in image references and deployment manifests. See [How regional endpoints work](https://learn.microsoft.com/en-us/azure/container-registry/container-registry-geo-replication#how-regional-endpoints-work).

The `acrRE` regex only matched a single DNS label before the ACR suffix:

```
^[a-zA-Z0-9]+\.(azurecr\.io|azurecr\.cn|azurecr\.de|azurecr\.us)
```

The `[a-zA-Z0-9]+` sub-pattern matches exactly one DNS label — a single run of alphanumeric characters with no dot — so the anchored expression only accepts a **single-label** registry name directly in front of the suffix and cannot match a host with any additional dot-separated labels before it. `myregistry.eastus2.geo.azurecr.io` carries two extra labels (`eastus2` and `geo`), so it does not match `acrRE`; on the primary matching path `parseACRLoginServerFromImage` returns `""` for a regional host (the behavior the unit tests capture, where `environment == nil`). See the note under **Special notes** for exactly what this does and does not change at AKS runtime.

This PR updates `acrRE` to accept an optional `.<region>.geo` segment:

```
^[a-zA-Z0-9]+(\.[a-zA-Z0-9]+\.geo)?\.(azurecr\.io|azurecr\.cn|azurecr\.de|azurecr\.us)
```

Only the `.geo` login server form is matched. **Dedicated data endpoints** (`myregistry.<region>.data.azurecr.io`) are intentionally not matched: they are layer-blob-download redirect targets (the registry issues a 307 redirect to them during a pull) and never appear as the registry host in an image reference, so they never need a credential lookup. See the [endpoint reference](https://learn.microsoft.com/en-us/azure/container-registry/container-registry-endpoint-reference#endpoint-types) for the distinction between regional `.geo` login servers and dedicated `.data` endpoints.

The change is minimal and self-contained: `GetCredentials` already inserts an exact-host auth entry for the parsed login server, so the existing `*.azurecr.*` response wildcards need no change. The downstream guard in `parseACRLoginServerFromImage` (trim the matched host, then require the remaining string to start with `/`) still correctly rejects lookalikes such as `foo.eastus2.geo.azurecr.io.example`.

Fixes #10667

This PR has three parts:
- **Provider fix:** the `acrRE` pattern in `pkg/credentialprovider/azure_credentials.go`, plus table-driven unit tests in `azure_credentials_test.go` (regional `.geo` for a public and a sovereign cloud, a negative lookalike case, and negative cases asserting dedicated data endpoints `<registry>.<region>.data.azurecr.io` do not match).
- **Recommended config:** added the `*.*.geo.azurecr.*` patterns to the example out-of-tree `matchImages` config (`examples/out-of-tree/credential-provider-config.yaml` and `-win.yaml`) so kubelet invokes the provider for 5-part regional login servers. The CI credential-provider lanes consume this config.
- **E2E coverage:** a new test in `tests/e2e/auth/cred.go` (suite `Credential`) that creates a Premium registry with regional endpoints enabled, then pulls an image via `<registry>.<region>.geo.azurecr.io` with no image pull secrets. The existing credential e2e only exercised the global endpoint. This test skips on the AKS pipeline, where the node `matchImages` is managed by AgentBaker (tracked in **Azure/AgentBaker#8862**).

`go test ./pkg/credentialprovider/...`, `gofmt -l`, and `go vet ./pkg/credentialprovider/` are clean; the `tests` module builds and vets clean.

**Manual end-to-end validation on live AKS nodes** (Premium ACR, RBAC / non-ABAC, anonymous pull disabled, regional endpoints enabled; kubelet identity holds `AcrPull` via `az aks --attach-acr`):
- **Default AKS cluster:** a pod pulling from the regional endpoint `<registry>.<region>.geo.azurecr.io` gets `ImagePullBackOff` / `401`, while the global-endpoint control pod runs. Reconfirmed on a fresh cluster.
- **Isolating the gate:** adding only the geo `matchImages` patterns to the node config and keeping the **stock (unpatched)** provider binary (sha verified unchanged), then pulling a distinct never-cached image via the regional endpoint with `imagePullPolicy: Always`, **succeeds**. So the node `matchImages` config is the functional gate for AKS regional pulls; once kubelet invokes the provider, the current provider already resolves the regional host via its `environment.ContainerRegistryDNSSuffix` substring fallback.
- This PR makes the provider's **primary `acrRE` path** match regional `.geo` hosts explicitly (instead of relying on that fallback) and adds regression tests. It is a correctness/robustness improvement rather than the functional unblock, which is the node `matchImages` change ([Azure/AgentBaker#8862](https://github.com/Azure/AgentBaker/issues/8862)).

The functional fix for AKS regional pulls is the node `matchImages` change ([Azure/AgentBaker#8862](https://github.com/Azure/AgentBaker/issues/8862)); this PR is the provider-side correctness improvement and is correct independently.

```release-note
Match Azure Container Registry regional (geo-replica) login server endpoints of the form `<registry>.<region>.geo.azurecr.io` explicitly in the ACR credential provider's `acrRE` pattern, instead of relying on the fallback substring match, and add regression tests.
```

```docs
- [ACR regional endpoints (login servers for geo-replicas)]: https://learn.microsoft.com/en-us/azure/container-registry/container-registry-geo-replication#how-regional-endpoints-work
- [ACR endpoint reference (regional `.geo` vs dedicated data `.data`)]: https://learn.microsoft.com/en-us/azure/container-registry/container-registry-endpoint-reference#endpoint-types
```

### DIFF
--- a/pkg/credentialprovider/azure_credentials.go
+++ b/pkg/credentialprovider/azure_credentials.go
@@ -49,8 +49,12 @@ const (
 
 var (
 	containerRegistryUrls = []string{"*.azurecr.io", "*.azurecr.cn", "*.azurecr.de", "*.azurecr.us"}
-	// a valid acr image starts with alphanumerics, followed by corresponding acr domain name.
-	acrRE = regexp.MustCompile(`^[a-zA-Z0-9]+\.(azurecr\.io|azurecr\.cn|azurecr\.de|azurecr\.us)`)
+	// a valid acr image starts with alphanumerics, followed by an optional
+	// regional login server segment (.<region>.geo), then the corresponding
+	// acr domain name. Examples:
+	//   foo.azurecr.io                 (global login server)
+	//   foo.eastus2.geo.azurecr.io     (regional login server)
+	acrRE = regexp.MustCompile(`^[a-zA-Z0-9]+(\.[a-zA-Z0-9]+\.geo)?\.(azurecr\.io|azurecr\.cn|azurecr\.de|azurecr\.us)`)
 )
 
 // CredentialProvider is an interface implemented by the kubelet credential provider plugin to fetch
--- a/examples/out-of-tree/credential-provider-config-win.yaml
+++ b/examples/out-of-tree/credential-provider-config-win.yaml
@@ -9,5 +9,9 @@ providers:
   - "*.azurecr.cn"
   - "*.azurecr.de"
   - "*.azurecr.us"
+  - "*.*.geo.azurecr.io"
+  - "*.*.geo.azurecr.cn"
+  - "*.*.geo.azurecr.de"
+  - "*.*.geo.azurecr.us"
   args:
   - c:\k\azure.json
--- a/examples/out-of-tree/credential-provider-config.yaml
+++ b/examples/out-of-tree/credential-provider-config.yaml
@@ -9,6 +9,10 @@ providers:
   - "*.azurecr.cn"
   - "*.azurecr.de"
   - "*.azurecr.us"
+  - "*.*.geo.azurecr.io"
+  - "*.*.geo.azurecr.cn"
+  - "*.*.geo.azurecr.de"
+  - "*.*.geo.azurecr.us"
   - "mcr.microsoft.com"
   args:
   - /etc/kubernetes/azure.json
--- a/pkg/credentialprovider/azure_credentials_test.go
+++ b/pkg/credentialprovider/azure_credentials_test.go
@@ -307,6 +307,31 @@ func TestParseACRLoginServerFromImage(t *testing.T) {
 			expected: "foo.azurecr.us",
 		},
 		{
+			// regional endpoint (geo-replicated Premium registry)
+			image:    "foo.eastus2.geo.azurecr.io/bar/image:version",
+			expected: "foo.eastus2.geo.azurecr.io",
+		},
+		{
+			// regional endpoint in a sovereign cloud
+			image:    "foo.chinaeast2.geo.azurecr.cn/bar/image:version",
+			expected: "foo.chinaeast2.geo.azurecr.cn",
+		},
+		{
+			// regional-looking host with a trailing suffix must not match
+			image:    "foo.eastus2.geo.azurecr.io.example/bar/image:version",
+			expected: "",
+		},
+		{
+			// dedicated data endpoints are not login servers and must not match
+			image:    "foo.eastus2.data.azurecr.io/bar/image:version",
+			expected: "",
+		},
+		{
+			// dedicated data endpoint in a sovereign cloud must not match either
+			image:    "foo.chinaeast2.data.azurecr.cn/bar/image:version",
+			expected: "",
+		},
+		{
 			image:    "foo.azurecr.my.cloud/bar/image:version",
 			expected: "foo.azurecr.my.cloud",
 		},
--- a/tests/e2e/auth/cred.go
+++ b/tests/e2e/auth/cred.go
@@ -91,6 +91,51 @@ var _ = Describe("Azure Credential Provider", Label(utils.TestSuiteLabelCredenti
 		Expect(err).NotTo(HaveOccurred())
 	})
 
+	It("should be able to pull private images from an acr regional (geo) endpoint without docker secrets set explicitly", func() {
+		if os.Getenv(utils.AKSTestCCM) != "" {
+			Skip("Skip the regional endpoint test on the AKS pipeline: the node credential-provider matchImages config is managed by AgentBaker (tracked in Azure/AgentBaker#8862), not by this repo")
+		}
+
+		// Regional endpoints are per-geo-replica login servers of the form
+		// <registry>.<region>.geo.azurecr.io. They require the Premium SKU and can be
+		// enabled even without geo-replication. This verifies that the ACR credential
+		// provider authenticates pulls from a regional endpoint, not only the global one.
+		err = utils.AZACRLogin()
+		Expect(err).NotTo(HaveOccurred())
+
+		registry, err := tc.CreateContainerRegistry()
+		Expect(err).NotTo(HaveOccurred())
+
+		defer func() {
+			err = tc.DeleteContainerRegistry(*registry.Name)
+			if err != nil {
+				utils.Logf("failed to cleanup registry with error: %w", err)
+			}
+		}()
+
+		// Enable regional (geo) endpoints on the Premium registry.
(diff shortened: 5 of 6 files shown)


pkg/credentialprovider/azure_credentials.go
lizMSFT, line 239: this SP path still uses `*.azurecr.io`  style wildcards

johnsonshi:
I am closing this PR in favor of https://github.com/kubernetes-sigs/cloud-provider-azure/pull/10720 opened by @lizMSFT

Issue #10273, opened 2026-05-07
Track cloud-controller-manager entrypoint drift from upstream


nilo19:
/kind cleanup

#### What would you like to be added:

Track and close the known gaps between Azure's cloud-controller-manager entrypoint and upstream `k8s.io/cloud-provider/app/controllermanager.go`, while preserving Azure's dynamic reloading behavior.

Azure currently pins `k8s.io/cloud-provider v0.35.0`, but `cmd/cloud-controller-manager/app/controllermanager.go` carries a local fork of upstream command construction, serving, leader election, controller startup, and controller context setup. The goal is to migrate toward upstream-owned code where possible and make the remaining Azure-owned boundary explicit.

Required Azure behavior that must be preserved:

- Dynamic cloud config reloading must remain supported.
- A reload must restart the active controller generation without releasing the current leader lease.
- File-backed and secret-backed cloud config loading must continue to work.
- Azure-specific node IPAM, node filtering, tracing, Azure SDK metrics, and Azure cloud initialization must continue to work.

Observed gaps:

1. Command construction is locally forked instead of using upstream builder.
   - Upstream has `app.CommandBuilder` with provider hooks for options, cloud initializer, controller registration, webhooks, additional flags, provider defaults, command name, and stop channel.
   - Azure directly constructs the Cobra command in `cmd/cloud-controller-manager/app/controllermanager.go`.
   - Migration gap: Azure cannot call upstream `BuildCommand()` directly until dynamic reload and Azure-specific setup have a defined hook/boundary.

2. Dynamic reload exists only in Azure.
   - Azure has `RunWrapper`, `runAsync`, file/secret watchers, and `shouldDisableCloudProvider`.
   - Upstream starts controllers once per process/leader lifecycle and has no reload supervisor.
   - Migration gap: direct use of upstream `Run()` would drop dynamic reloading.

3. Reload currently depends on staying inside `OnStartedLeading`.
   - Azure invokes `RunWrapper` from the leader elector's `OnStartedLeading` callback.
   - On config update, Azure cancels the active controller context and starts a new generation without returning from `OnStartedLeading`.
   - Migration gap: the reusable upstream lifecycle does not expose a generation-level restart hook under the same leader callback.

4. Static mode has a latent fallthrough bug.
   - In Azure `RunWrapper`, the non-dynamic path calls `Run(ctx, c.Complete(), h)` but does not return afterward.
   - That means static mode can fall through into dynamic watcher setup.
[...]

PR #10714, opened 2026-07-20
[Replacement for #10641] ServiceGateway (SLB) common-ground provider integration

david-kow:
> **Replacement for #10641**
>
> Edward (@georgeedward2000) is currently unavailable, so this PR recreates #10641 from the same commit stack. It includes all commits from #10641 plus one additional commit addressing the latest review comments.

#### What type of PR is this?

/kind feature

Wires ServiceGateway (Service Load Balancer) support into the shared provider layer. When
`ServiceGatewayEnabled` is set, load balancers are managed per-service (PodIP backend pools)
via an async diff tracker instead of the synchronous reconcile path.

- Route `GetLoadBalancer`/`EnsureLoadBalancer`/`UpdateLoadBalancer`/`EnsureLoadBalancerDeleted`
  through the diff tracker; per-service LB/PIP naming and `Service`/`StandardV2` SKUs.
- Per-service backend pools and endpoint-slice → diff-tracker endpoint sync.
- Inbound-config validation (rejects internal, dual-stack, named/duplicate ports, non-TCP/UDP).
- Adds the `servicegateway/difftracker` package as the API surface the integration compiles
  against; the engine implementation is delivered separately.

All new behavior is gated on `ServiceGatewayEnabled` (default false), so existing flows are unchanged.

Fixes #

`pkg/provider/servicegateway/difftracker` is the API surface only (no-op stubs); the engine lands
in a follow-up. Review focuses on the common-ground hooks in the shared provider files.

```release-note
NONE

### CODE REVIEW

**Comment by nilo19** [comment]
@david-kow could you please sign CLA?

File: pkg/provider/azure_loadbalancer.go [hunk 1]
```diff
@@ -429,6 +429,7 @@ func (az *Cloud) EnsureLoadBalancerDeleted(ctx context.Context, clusterName stri
 		isOperationSucceeded = false
 	)
 	ctx = log.NewContext(ctx, logger)
+
```
wonderyl: nit, avoids touching unnecessary files

david-kow: ofc, fixed

david-kow:
> @david-kow could you please sign CLA?

done

pkg/util/sets/string.go
```diff
@@ -125,3 +111,24 @@ func (s *IgnoreCaseSet) Difference(other *IgnoreCaseSet) *IgnoreCaseSet {
...
+	if s.Len() != other.Len() {
+		return false
+	}
+	if s.Len() == 0 {
+		return true
+	}
+	return s.set.Equal(other.set)
+}
+
+func (s *IgnoreCaseSet) MarshalJSON() ([]byte, error) {
```
wonderyl, line 126: please add tests

david-kow: added

pkg/provider/azure.go
```diff
@@ -702,13 +669,20 @@ func (az *Cloud) Initialize(clientBuilder cloudprovider.ControllerClientBuilder,
 	az.eventBroadcaster = record.NewBroadcaster()
 	az.eventBroadcaster.StartRecordingToSink(&v1core.EventSinkImpl{Interface: az.KubeClient.CoreV1().Events("")})
 	az.eventRecorder = az.eventBroadcaster.NewRecorder(scheme.Scheme, v1.EventSource{Component: "azure-cloud-provider"})
-	if az.ServiceGatewayEnabled {
-		az.diffTracker.SetEventRecorder(az.eventRecorder)
-	}
+	az.serviceGatewayRuntime.SetKubeClient(az.KubeClient)
```
wonderyl: It's safer to have null check for serviceGatewayRuntime here, instead of in SetKubeClient.

pkg/provider/azure.go
```diff
@@ -355,6 +348,31 @@ func (az *Cloud) InitializeCloudFromConfig(ctx context.Context, config *azurecon
...
+
+	if serviceGatewayPartiallyEnabled && !serviceGatewayFullyEnabled {
+		return fmt.Errorf("InitializeCloudFromConfig: ServiceGateway requires serviceGatewayEnabled=true, loadBalancerBackendPoolConfigurationType=podIP, and loadBalancerSku=service to be configured together (actual: serviceGatewayEnabled=%t, loadBalancerBackendPoolConfigurationType=%s, loadBalancerSku=%s)",
+			config.ServiceGatewayEnabled,
+			config.LoadBalancerBackendPoolConfigurationType,
+			config.LoadBalancerSKU)
+	}
+
+	if serviceGatewayFullyEnabled {
+		config.ServiceGatewayEnabled = true
```
wonderyl: ServiceGatewayEnabled should be true already.

pkg/provider/azure.go
```diff
@@ -642,14 +663,26 @@ func (az *Cloud) setCloudProviderBackoffDefaults(config *azureconfig.Config) wai
...
+
+// ServiceGatewayRuntime returns the ServiceGateway runtime used by the Service controller.
+func (az *Cloud) ServiceGatewayRuntime() *servicegateway.Runtime {
+	return az.serviceGatewayRuntime
 }
 
 // LoadBalancer returns a balancer interface. Also returns true if the interface is supported, false otherwise.
 func (az *Cloud) LoadBalancer() (cloudprovider.LoadBalancer, bool) {
+	if az.serviceGatewayRuntime != nil {
+		return az.serviceGatewayRuntime.LoadBalancer()
```
wonderyl, line 685: should return true as well?

pkg/provider/azure_local_services.go
```diff
@@ -341,8 +341,22 @@ func (az *Cloud) getLocalServiceInfo(serviceName string) (*serviceInfo, bool) {
 	return data.(*serviceInfo), true
 }
 
-// setUpEndpointSlicesInformer creates an informer for EndpointSlices of local services.
-// It watches the update events and send backend pool update operations to the batch updater.
+func endpointSliceFromDeleteEvent(obj interface{}) (*discovery_v1.EndpointSlice, error) {
```
wonderyl: Is this change related to ServiceLB?

pkg/util/sets/string.go
wonderyl, line 126: Doesn't seem MarshalJSON is used in this PR, is that expected?

pkg/provider/azure_fakes.go
```diff
@@ -186,3 +188,13 @@ func GetTestCloudWithExtendedLocation(ctrl *gomock.Controller) (az *Cloud) {
 	az.ExtendedLocationType = "EdgeZone"
 	return az
 }
+
+// GetTestCloudWithContainerLoadBalancer returns a fake azure cloud for unit tests in Azure supporting container load balancer.
+func GetTestCloudWithContainerLoadBalancer(ctrl *gomock.Controller) (az *Cloud) {
```
wonderyl: nit: the concept of ContainerLoadBalancer is only used here. In other places, you call it ServiceGateway. I think sticking with one concept would help to reduce confusion

pkg/provider/azure.go
```diff
@@ -355,6 +348,31 @@ func (az *Cloud) InitializeCloudFromConfig(ctx context.Context, config *azurecon
 		return err
 	}
 
+	serviceGatewayPartiallyEnabled := config.ServiceGatewayEnabled || config.IsLBBackendPoolTypePodIP() || config.UseServiceLoadBalancer()
+	serviceGatewayFullyEnabled := config.ServiceGatewayEnabled && config.IsLBBackendPoolTypePodIP() && config.UseServiceLoadBalancer()
```
wonderyl, line 352: Do you need to check `LoadBalancerSKU == consts.LoadBalancerSKUService` as well?

pkg/provider/azure_test.go
```diff
@@ -1632,6 +1633,83 @@ func TestProtocolTranslationUDP(t *testing.T) {
...
+func serviceGatewayConfig() *providerconfig.Config {
+	return &providerconfig.Config{
+		ServiceGatewayEnabled:                    true,
+		LoadBalancerSKU:                          consts.LoadBalancerSKUService,
+		LoadBalancerBackendPoolConfigurationType: consts.LoadBalancerBackendPoolConfigurationTypePodIP,
+	}
+}
+
+func TestNewCloudInitializesKubeClientAtRequiredStage(t *testing.T) {
+	t.Run("ServiceGateway CCM", func(t *testing.T) {
```
wonderyl: test name doesn't match with content.

pkg/provider/azure_test.go
```diff
@@ -1632,6 +1633,83 @@ func TestProtocolTranslationUDP(t *testing.T) {
...
+		config.MultipleStandardLoadBalancerConfigurations = []providerconfig.MultipleStandardLoadBalancerConfiguration{{}}
+
+		_, err := NewCloud(context.Background(), builder, config, true)
+
+		assert.ErrorContains(t, err, "mutually exclusive")
+		assert.Zero(t, builder.calls)
+	})
+
+	t.Run("invalid non-ServiceGateway CCM", func(t *testing.T) {
+		builder := &trackingControllerClientBuilder{client: fake.NewSimpleClientset()}
```
wonderyl, line 1669: I would argue the value of this test. It's not testing anything related to ServiceGateway. I assume wrong backend pool config type is covered by existing test, but I could be wrong.

pkg/provider/azure_test.go
```diff
@@ -1632,6 +1633,83 @@ func TestProtocolTranslationUDP(t *testing.T) {
...
+		config := &providerconfig.Config{LoadBalancerBackendPoolConfigurationType: "invalid"}
+
+		_, err := NewCloud(context.Background(), builder, config, true)
+
+		assert.ErrorContains(t, err, "is not supported")
+		assert.Zero(t, builder.calls)
+	})
+
+	t.Run("non-ServiceGateway non-CCM", func(t *testing.T) {
```
wonderyl, line 1678: why this is non-ccm?

pkg/provider/azure_test.go
```diff
@@ -1632,6 +1633,83 @@ func TestProtocolTranslationUDP(t *testing.T) {
...
+		if err != nil {
+			return
+		}
+		assert.Same(t, kubeClient, cloud.(*Cloud).KubeClient)
+		assert.Nil(t, cloud.(*Cloud).ServiceGatewayRuntime())
+		assert.Equal(t, 1, builder.calls)
+	})
+
+	t.Run("ServiceGateway non-CCM", func(t *testing.T) {
+		builder := &trackingControllerClientBuilder{client: fake.NewSimpleClientset()}
```
wonderyl, line 1694: The code seems to be exactly the same as line 1656 "ServiceGateway CCM"

pkg/provider/azure_test.go
```diff
@@ -1632,6 +1633,83 @@ func TestProtocolTranslationUDP(t *testing.T) {
 	}
 }
 
+type trackingControllerClientBuilder struct {
```
wonderyl, line 1637: What is "tracking controller?"

pkg/provider/azure_test.go
```diff
@@ -1632,6 +1633,83 @@ func TestProtocolTranslationUDP(t *testing.T) {
...
+		config := serviceGatewayConfig()
+		config.MultipleStandardLoadBalancerConfigurations = []providerconfig.MultipleStandardLoadBalancerConfiguration{{}}
+
+		_, err := NewCloud(context.Background(), builder, config, false)
+
+		assert.ErrorContains(t, err, "mutually exclusive")
+		assert.Zero(t, builder.calls)
+	})
+
+	t.Run("ServiceGateway CCM defers builder requirement", func(t *testing.T) {
```
wonderyl, line 1704: Is this case related to ServiceGateway.

pkg/provider/azure.go
david-kow: actually, I'll do it in both places, as this could be called from different place in the future

pkg/provider/azure.go
david-kow: that's in `config.UseServiceLoadBalancer()`:

```
func (az *Config) UseServiceLoadBalancer() bool {
	return strings.EqualFold(az.LoadBalancerSKU, consts.LoadBalancerSKUService)
}
```

pkg/provider/azure.go
david-kow: This invokes a function that returns it:

```
// LoadBalancer returns the ServiceGateway LoadBalancer when enabled.
func (r *Runtime) LoadBalancer() (cloudprovider.LoadBalancer, bool) {
	if r == nil || !r.enabled || r.loadBalancer == nil {
		return nil, false
	}
	return r.loadBalancer, true
}
```

pkg/provider/azure_local_services.go
david-kow: Helper was needed previously as we wanted same behavior in service lb and standard lb, so this code was moved out to be called from two places. Now it's only called from standard lb, but I thought it makes setting up DeleteFunc more readable:

```go
DeleteFunc: func(obj interface{}) {
	endpointSlice, err := endpointSliceFromDeleteEvent(obj)
	if err != nil {
		logger.Error(err, "Cannot process EndpointSlice deletion")
		return
	}
	az.endpointSlicesCache.Delete(strings.ToLower(fmt.Sprintf("%s/%s", endpointSlice.Namespace, endpointSlice.Name)))
},
```

I'll resolve this for now, if you want me to remove the helper and do it inline, please reopen 🙂

pkg/util/sets/string.go
david-kow: yes, I believe we need it later in servicegateway dir, but wanted to do all non-servicegateway changes in this pr

pkg/provider/azure_fakes.go
david-kow: makes sense

pkg/provider/azure_test.go
david-kow: it's because false is passed to NewCloud which indicates non-ccm path:

`cloud, err := NewCloud(context.Background(), builder, &providerconfig.Config{}, >>>false<<<)`

I don't have full context on the scenario, but to be safe we want to make sure that no servicegateway related initialization takes place in such case

pkg/provider/azure_test.go
david-kow: a controller that tracks calls made to it 🙂 I'll add a comment

pkg/provider/azure_test.go
david-kow: they differ in `NewCloud()`s last arg:

first one has:
`_, err := NewCloud(context.Background(), builder, config, true)`

this one has:
`_, err := NewCloud(context.Background(), builder, config, false)`

pkg/provider/azure_test.go
david-kow: ah, yes, prepending with "invalid'

pkg/provider/azure_test.go
david-kow: yes - this tests whether enabling servicegateway with invalid config allows to get to using clientbuilder

pkg/provider/azure_test.go
david-kow: existing test is there but only testing directly `InitializeCloudFromConfig` not via `NewCloud`. as we've made sgw-related changes in both of those, here we validate whether enabling sgw doesn't change newcloud in a way that would allow invalid config to proceed to create kubeclient. creating kubeclient can panic (`.ClientOrDie`), so we can be more careful here. I'll resolve for now - if you want to have this test removed, please reopen 🙂

pkg/provider/azure_local_services.go
```diff
@@ -341,8 +341,22 @@ func (az *Cloud) getLocalServiceInfo(serviceName string) (*serviceInfo, bool) {
...
+	switch value := obj.(type) {
+	case *discovery_v1.EndpointSlice:
+		return value, nil
+	case cache.DeletedFinalStateUnknown:
+		endpointSlice, ok := value.Obj.(*discovery_v1.EndpointSlice)
+		if !ok {
+			return nil, fmt.Errorf("cannot convert tombstone object %T to *discovery_v1.EndpointSlice", value.Obj)
```
nilo19: Please avoid using %t.

pkg/provider/servicegateway/difftracker/loadbalancer.go
```diff
@@ -0,0 +1,144 @@
...
+	return service.Status.LoadBalancer.DeepCopy(), true, nil
+}
+
+func (lb *LoadBalancer) GetLoadBalancerName(_ context.Context, _ string, service *v1.Service) string {
+	return cloudprovider.DefaultLoadBalancerName(service)
+}
+
+func (lb *LoadBalancer) EnsureLoadBalancer(ctx context.Context, _ string, service *v1.Service, _ []*v1.Node) (status *v1.LoadBalancerStatus, err error) {
+	const operation = "EnsureLoadBalancer"
+	ctx, span := trace.BeginReconcile(ctx, trace.DefaultTracer(), operation)
```
nilo19, line 95: Do we need a new name for container LB operations? e.g., EnsureLoadBalancerServiceGateway?

nilo19:
One concern: this implementation enlarge the gaps between upstream cloud-provider.
https://github.com/kubernetes-sigs/cloud-provider-azure/issues/10273

pkg/provider/servicegateway/difftracker/loadbalancer.go
david-kow: as in, the func name? that's the impl. for the interface:

```go
type LoadBalancer interface {
	...
	EnsureLoadBalancer(ctx context.Context, clusterName string, service *v1.Service, nodes []*v1.Node) (*v1.LoadBalancerStatus, error)
    ...
}
```

for operation name, shouldn't these be 1:1 with func names?

david-kow:
> One concern: this implementation enlarge the gaps between upstream cloud-provider. #10273

Would that be something that has to be addressed in this PR or should it be done separately?

pkg/provider/azure_local_services.go
david-kow: brought back previous logging

pkg/provider/servicegateway/difftracker/loadbalancer.go
david-kow: I'll resolve for now, please reopen if I'm missing something

pkg/provider/azure_local_services.go
```diff
@@ -341,18 +342,20 @@ func (az *Cloud) getLocalServiceInfo(serviceName string) (*serviceInfo, bool) {
 	return data.(*serviceInfo), true
 }
 
-func endpointSliceFromDeleteEvent(obj interface{}) (*discovery_v1.EndpointSlice, error) {
+func endpointSliceFromDeleteEvent(logger logr.Logger, obj interface{}) *discovery_v1.EndpointSlice {
```
wonderyl, line 345: Is this change related to CNLB?

nilo19:
Manual merged, e2e failures not related.

Issue #9975, opened 2026-02-24
ACR credential provider regex doesn't handle regional endpoints or registries w/ DNL enabled


juliusl:
This isn't tied to an active incident, I just happened to be reviewing the code and noticed the regex here:

https://github.com/kubernetes-sigs/cloud-provider-azure/blob/17dbc506a6bed9d0127ce045891fb55af012b84c/pkg/credentialprovider/azure_credentials.go#L51C1-L52C1

Needs to be updated to support ACR's regional endpoints as well as registries w/ DNL enabled. 

https://github.com/azure/acr/issues/842

https://learn.microsoft.com/en-us/azure/container-registry/container-registry-get-started-portal?tabs=azure-cli#configure-domain-name-label-dnl-option

The current regex wouldn't be able to capture DNL enabled because they include a dash, and for regional endpoints there would be extra dots.

#### Environment:

- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Install tools:
- Network plugin and version (if this is a network-related bug):
- Others:

PR #10720, opened 2026-07-21
fix: match ACR regional (geo-replica) login server endpoints in credential provider

### LINKED ISSUE #10667: ACR credential provider does not match regional (geo-replica) login server endpoints

lizMSFT:
#### What type of PR is this?

/kind bug

The ACR credential provider did not recognize Azure Container Registry **regional endpoints** — the per-geo-replica login servers of the form `myregistry.<region>.geo.azurecr.io` (for example, `myregistry.eastus2.geo.azurecr.io`).

Regional endpoints are login servers for a specific geo-replica. They let clients pull an image from a chosen geo-replica for predictable routing, in-region affinity, and client-side failover, and they can be referenced directly in image references and deployment manifests. See [How regional endpoints work](https://learn.microsoft.com/en-us/azure/container-registry/container-registry-geo-replication#how-regional-endpoints-work).

The `acrRE` regex only matched a single DNS label before the ACR suffix:

```
^[a-zA-Z0-9]+\.(azurecr\.io|azurecr\.cn|azurecr\.de|azurecr\.us)
```

The `[a-zA-Z0-9]+` sub-pattern matches exactly one DNS label — a single run of alphanumeric characters with no dot — so the anchored expression only accepts a **single-label** registry name directly in front of the suffix and cannot match a host with any additional dot-separated labels before it. `myregistry.eastus2.geo.azurecr.io` carries two extra labels (`eastus2` and `geo`), so it does not match `acrRE`; on the primary matching path `parseACRLoginServerFromImage` returns `""` for a regional host (the behavior the unit tests capture, where `environment == nil`). See the note under **Special notes** for exactly what this does and does not change at AKS runtime.

This PR updates `acrRE` to accept an optional `.<region>.geo` segment:

```
^[a-zA-Z0-9]+(\.[a-zA-Z0-9]+\.geo)?\.(azurecr\.io|azurecr\.cn|azurecr\.de|azurecr\.us)
```

Only the `.geo` login server form is matched. **Dedicated data endpoints** (`myregistry.<region>.data.azurecr.io`) are intentionally not matched: they are layer-blob-download redirect targets (the registry issues a 307 redirect to them during a pull) and never appear as the registry host in an image reference, so they never need a credential lookup. See the [endpoint reference](https://learn.microsoft.com/en-us/azure/container-registry/container-registry-endpoint-reference#endpoint-types) for the distinction between regional `.geo` login servers and dedicated `.data` endpoints.

The change is minimal and self-contained: `GetCredentials` already inserts an exact-host auth entry for the parsed login server, so the existing `*.azurecr.*` response wildcards need no change. The downstream guard in `parseACRLoginServerFromImage` (trim the matched host, then require the remaining string to start with `/`) still correctly rejects lookalikes such as `foo.eastus2.geo.azurecr.io.example`.

Fixes #10667

This PR has three parts:
- **Provider fix:** the `acrRE` pattern in `pkg/credentialprovider/azure_credentials.go`, plus table-driven unit tests in `azure_credentials_test.go` (regional `.geo` for a public and a sovereign cloud, a negative lookalike case, and negative cases asserting dedicated data endpoints `<registry>.<region>.data.azurecr.io` do not match).
- **Recommended config:** added the `*.*.geo.azurecr.*` patterns to the example out-of-tree `matchImages` config (`examples/out-of-tree/credential-provider-config.yaml` and `-win.yaml`) so kubelet invokes the provider for 5-part regional login servers. The CI credential-provider lanes consume this config.
- **E2E coverage:** a new test in `tests/e2e/auth/cred.go` (suite `Credential`) that creates a Premium registry with regional endpoints enabled, then pulls an image via `<registry>.<region>.geo.azurecr.io` with no image pull secrets. The existing credential e2e only exercised the global endpoint. This test skips on the AKS pipeline, where the node `matchImages` is managed by AgentBaker (tracked in **Azure/AgentBaker#8862**).

`go test ./pkg/credentialprovider/...`, `gofmt -l`, and `go vet ./pkg/credentialprovider/` are clean; the `tests` module builds and vets clean.

**Manual end-to-end validation on live AKS nodes** (Premium ACR, RBAC / non-ABAC, anonymous pull disabled, regional endpoints enabled; kubelet identity holds `AcrPull` via `az aks --attach-acr`):
- **Default AKS cluster:** a pod pulling from the regional endpoint `<registry>.<region>.geo.azurecr.io` gets `ImagePullBackOff` / `401`, while the global-endpoint control pod runs. Reconfirmed on a fresh cluster.
- **Isolating the gate:** adding only the geo `matchImages` patterns to the node config and keeping the **stock (unpatched)** provider binary (sha verified unchanged), then pulling a distinct never-cached image via the regional endpoint with `imagePullPolicy: Always`, **succeeds**. So the node `matchImages` config is the functional gate for AKS regional pulls; once kubelet invokes the provider, the current provider already resolves the regional host via its `environment.ContainerRegistryDNSSuffix` substring fallback.
- This PR makes the provider's **primary `acrRE` path** match regional `.geo` hosts explicitly (instead of relying on that fallback) and adds regression tests. It is a correctness/robustness improvement rather than the functional unblock, which is the node `matchImages` change ([Azure/AgentBaker#8862](https://github.com/Azure/AgentBaker/issues/8862)).

The functional fix for AKS regional pulls is the node `matchImages` change ([Azure/AgentBaker#8862](https://github.com/Azure/AgentBaker/issues/8862)); this PR is the provider-side correctness improvement and is correct independently.

```release-note
Match Azure Container Registry regional (geo-replica) login server endpoints of the form `<registry>.<region>.geo.azurecr.io` explicitly in the ACR credential provider's `acrRE` pattern, instead of relying on the fallback substring match, and add regression tests.
```

```docs
- [ACR regional endpoints (login servers for geo-replicas)]: https://learn.microsoft.com/en-us/azure/container-registry/container-registry-geo-replication#how-regional-endpoints-work
- [ACR endpoint reference (regional `.geo` vs dedicated data `.data`)]: https://learn.microsoft.com/en-us/azure/container-registry/container-registry-endpoint-reference#endpoint-types
```

### DIFF
--- a/pkg/credentialprovider/azure_credentials.go
+++ b/pkg/credentialprovider/azure_credentials.go
@@ -244,6 +244,21 @@ func (a *acrProvider) GetCredentials(ctx context.Context, image string, _ []stri
 			response.Auth[url] = cred
 		}
 
+		// The static wildcard patterns above (e.g. *.azurecr.io) only match a
+		// single label before the ACR suffix, so they do not cover regional
+		// (geo-replica) login servers such as foo.eastus2.geo.azurecr.io. Add an
+		// explicit entry for the exact parsed login server(s) so service-principal
+		// pulls from regional endpoints are authenticated, mirroring the managed
+		// identity path above.
+		spCred := v1.AuthConfig{
+			Username: a.config.AADClientID,
+			Password: a.config.AADClientSecret,
+		}
+		response.Auth[targetloginServer] = spCred
+		if sourceloginServer != "" {
+			response.Auth[sourceloginServer] = spCred
+		}
+
 		// Handle the custom cloud case
 		// In clouds where ACR is not yet deployed, the string will be empty
 		if a.environment != nil && strings.Contains(a.environment.ContainerRegistryDNSSuffix, ".azurecr.") {
--- a/pkg/credentialprovider/azure_credentials.go
+++ b/pkg/credentialprovider/azure_credentials.go
@@ -49,8 +49,12 @@ const (
 
 var (
 	containerRegistryUrls = []string{"*.azurecr.io", "*.azurecr.cn", "*.azurecr.de", "*.azurecr.us"}
-	// a valid acr image starts with alphanumerics, followed by corresponding acr domain name.
-	acrRE = regexp.MustCompile(`^[a-zA-Z0-9]+\.(azurecr\.io|azurecr\.cn|azurecr\.de|azurecr\.us)`)
+	// a valid acr image starts with alphanumerics, followed by an optional
+	// regional login server segment (.<region>.geo), then the corresponding
+	// acr domain name. Examples:
+	//   foo.azurecr.io                 (global login server)
+	//   foo.eastus2.geo.azurecr.io     (regional login server)
+	acrRE = regexp.MustCompile(`^[a-zA-Z0-9]+(\.[a-zA-Z0-9]+\.geo)?\.(azurecr\.io|azurecr\.cn|azurecr\.de|azurecr\.us)`)
 )
 
 // CredentialProvider is an interface implemented by the kubelet credential provider plugin to fetch
--- a/pkg/credentialprovider/azure_credentials.go
+++ b/pkg/credentialprovider/azure_credentials.go
@@ -247,7 +247,7 @@ func (a *acrProvider) GetCredentials(ctx context.Context, image string, _ []stri
 		// The static wildcard patterns above (e.g. *.azurecr.io) only match a
 		// single label before the ACR suffix, so they do not cover regional
 		// (geo-replica) login servers such as foo.eastus2.geo.azurecr.io. Add an
-		// explicit entry for the exact parsed login server(s) so service-principal
+		// explicit entry for the exact parsed target login server so service-principal
 		// pulls from regional endpoints are authenticated, mirroring the managed
 		// identity path above.
 		spCred := v1.AuthConfig{
@@ -255,9 +255,6 @@ func (a *acrProvider) GetCredentials(ctx context.Context, image string, _ []stri
 			Password: a.config.AADClientSecret,
 		}
 		response.Auth[targetloginServer] = spCred
-		if sourceloginServer != "" {
-			response.Auth[sourceloginServer] = spCred
-		}
 
 		// Handle the custom cloud case
 		// In clouds where ACR is not yet deployed, the string will be empty
--- a/pkg/credentialprovider/azure_credentials.go
+++ b/pkg/credentialprovider/azure_credentials.go
@@ -49,13 +49,25 @@ const (
 
 var (
 	containerRegistryUrls = []string{"*.azurecr.io", "*.azurecr.cn", "*.azurecr.de", "*.azurecr.us"}
-	// a valid acr image starts with alphanumerics, followed by an optional
-	// regional login server segment (.<region>.geo), then the corresponding
-	// acr domain name. Examples:
+	// a valid acr image starts with a registry domain name label (alphanumerics,
+	// optionally followed by a single hyphen and a suffix for domain name label
+	// (DNL) registries, e.g. demo-abc123), followed by an optional regional login
+	// server segment (.<region>.geo), then the corresponding acr domain name.
+	// Examples:
 	//   foo.azurecr.io                 (global login server)
+	//   demo-abc123.azurecr.io         (domain name label (DNL) registry)
 	//   foo.eastus2.geo.azurecr.io     (regional login server)
-	acrRE = regexp.MustCompile(`^[a-zA-Z0-9]+(\.[a-zA-Z0-9]+\.geo)?\.(azurecr\.io|azurecr\.cn|azurecr\.de|azurecr\.us)`)
+	acrRE = regexp.MustCompile(`^[a-zA-Z0-9]+(-[a-zA-Z0-9]+)?(\.[a-zA-Z0-9]+\.geo)?\.(azurecr\.io|azurecr\.cn|azurecr\.de|azurecr\.us)`)
 )
+
+// isRegionalACRLoginServer reports whether loginServer is an ACR regional
+// (geo-replica) login server such as foo.eastus2.geo.azurecr.io. Such hosts have
+// an extra ".<region>.geo" segment that the static "*.azurecr.*" wildcard
+// patterns do not match, so they need an explicit auth entry. Global login
+// servers (e.g. foo.azurecr.io) are already covered by the wildcards.
+func isRegionalACRLoginServer(loginServer string) bool {
+	return strings.Contains(loginServer, ".geo.")
+}
 
 // CredentialProvider is an interface implemented by the kubelet credential provider plugin to fetch
 // the username/password based on the provided image name.
@@ -235,26 +247,27 @@ func (a *acrProvider) GetCredentials(ctx context.Context, image string, _ []stri
 			response.Auth[sourceloginServer] = authConfig
 		}
 	} else {
+		// The service principal credentials are identical for every ACR entry,
+		// so build them once and reuse.
+		spCred := v1.AuthConfig{
+			Username: a.config.AADClientID,
+			Password: a.config.AADClientSecret,
+		}
+
 		// Add our entry for each of the supported container registry URLs
 		for _, url := range containerRegistryUrls {
-			cred := v1.AuthConfig{
-				Username: a.config.AADClientID,
-				Password: a.config.AADClientSecret,
-			}
-			response.Auth[url] = cred
+			response.Auth[url] = spCred
 		}
 
 		// The static wildcard patterns above (e.g. *.azurecr.io) only match a
 		// single label before the ACR suffix, so they do not cover regional
 		// (geo-replica) login servers such as foo.eastus2.geo.azurecr.io. Add an
-		// explicit entry for the exact parsed target login server so service-principal
-		// pulls from regional endpoints are authenticated, mirroring the managed
-		// identity path above.
-		spCred := v1.AuthConfig{
-			Username: a.config.AADClientID,
-			Password: a.config.AADClientSecret,
+		// explicit entry for such regional targets so service-principal pulls from
+		// regional endpoints are authenticated. Global targets are already covered
+		// by the wildcards above, so we avoid adding a duplicate entry for them.
+		if isRegionalACRLoginServer(targetloginServer) {
+			response.Auth[targetloginServer] = spCred
 		}
-		response.Auth[targetloginServer] = spCred
 
 		// Handle the custom cloud case
 		// In clouds where ACR is not yet deployed, the string will be empty
@@ -269,11 +282,7 @@ func (a *acrProvider) GetCredentials(ctx context.Context, image string, _ []stri
 			}
 
 			if !hasBeenAdded {
-				cred := v1.AuthConfig{
-					Username: a.config.AADClientID,
-					Password: a.config.AADClientSecret,
-				}
-				response.Auth[customAcrSuffix] = cred
+				response.Auth[customAcrSuffix] = spCred
 			}
 		}
 	}
--- a/pkg/credentialprovider/azure_credentials_test.go
+++ b/pkg/credentialprovider/azure_credentials_test.go
@@ -77,8 +77,11 @@ func TestGetCredentials(t *testing.T) {
 		t.Fatalf("Unexpected error when fetching acr credentials: %v", err)
 	}
 
-	if credResponse == nil || len(credResponse.Auth) != len(result)+1 {
-		t.Errorf("Unexpected credential response: %v, expected length %d", credResponse, len(result)+1)
+	// Expected entries: the anonymous "*.azurecr.*" wildcard, one entry per
+	// containerRegistryUrls wildcard (len(result)), and one explicit entry for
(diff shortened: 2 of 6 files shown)


nilo19:
Fix: https://github.com/kubernetes-sigs/cloud-provider-azure/issues/9975

lizMSFT:
Hello @nilo19,  the PR has some CI failures, but they look unrelated to my change. Could you take a look and advise on next steps? Thanks!

pkg/credentialprovider/azure_credentials_test.go
```diff
@@ -306,6 +363,31 @@ func TestParseACRLoginServerFromImage(t *testing.T) {
...
+		{
+			// dedicated data endpoints are not login servers and must not match
+			image:    "foo.eastus2.data.azurecr.io/bar/image:version",
+			expected: "",
+		},
+		{
+			// dedicated data endpoint in a sovereign cloud must not match either
+			image:    "foo.chinaeast2.data.azurecr.cn/bar/image:version",
+			expected: "",
+		},
```
Liunardy, line 462: Add .de and .us cases?

tests/e2e/utils/container_registry_utils.go
```diff
@@ -33,15 +33,17 @@ import (
...
 func (tc *AzureTestClient) CreateContainerRegistry() (acr.Registry, error) {
 	acrClient := tc.createACRClient()
 	rgName, location := tc.GetResourceGroup(), tc.GetLocation()
 	acrName := "e2eacr" + string(uuid.NewUUID())[0:4]
 	template := acr.Registry{
 		SKU: &acr.SKU{
-			Name: to.Ptr(acr.SKUNameStandard),
-			Tier: to.Ptr(acr.SKUTierStandard),
+			Name: to.Ptr(acr.SKUNamePremium),
+			Tier: to.Ptr(acr.SKUTierPremium),
```
Liunardy: Could we create premium only for the E2E test that needs it?

pkg/credentialprovider/azure_credentials_test.go
```diff
@@ -93,6 +96,60 @@ func TestGetCredentials(t *testing.T) {
...
+		t.Fatalf("Unexpected error when fetching acr credentials: %v", err)
+	}
+
+	cred, found := credResponse.Auth["foo.eastus2.geo.azurecr.io"]
+	if !found {
+		t.Fatalf("Missing explicit entry for regional login server foo.eastus2.geo.azurecr.io; auth keys: %v", credResponse.Auth)
+	}
+	if cred.Username != "foo" || cred.Password != "bar" {
+		t.Errorf("Unexpected credentials for regional login server: %v", cred)
+	}
```
Liunardy, line 151: Can we have another test to cover registry mirror mapping host to a geo target?

pkg/credentialprovider/azure_credentials.go
```diff
@@ -240,6 +244,21 @@ func (a *acrProvider) GetCredentials(ctx context.Context, image string, _ []stri
...
+		// pulls from regional endpoints are authenticated, mirroring the managed
+		// identity path above.
+		spCred := v1.AuthConfig{
+			Username: a.config.AADClientID,
+			Password: a.config.AADClientSecret,
+		}
+		response.Auth[targetloginServer] = spCred
+		if sourceloginServer != "" {
+			response.Auth[sourceloginServer] = spCred
+		}
```
Liunardy, line 270: On the SP path, targetloginServer is always an ACR host, but sourceloginServer is an operator-configured mirror source that may be any host (e.g. mcr.microsoft.com, or anything else), so the raw AAD client secret is now emitted to an arbitrary mirror source.

NewAcrProvider states: "Assumption here is mirrored image is only used on NI cluster and for mirrored image, managed identity authentication is always used". Should the SP path emit a sourceloginServer entry at all?

pkg/credentialprovider/azure_credentials_test.go
lizMSFT: Made requested changes

tests/e2e/utils/container_registry_utils.go
lizMSFT: Made requested changes

pkg/credentialprovider/azure_credentials_test.go
lizMSFT: Made requested changes

pkg/credentialprovider/azure_credentials.go
lizMSFT: Good catch, it shouldn't. Removed the sourceloginServer emission on the SP path; it now only authenticates  targetloginServer  (always ACR). MI path is unchanged, so mirror scenarios still work there. Added a test asserting the geo target gets creds while the mirror source host doesn't.

nilo19:
Have we manually tested if it works?

pkg/credentialprovider/azure_credentials.go
```diff
@@ -240,6 +244,18 @@ func (a *acrProvider) GetCredentials(ctx context.Context, image string, _ []stri
...
+		// single label before the ACR suffix, so they do not cover regional
+		// (geo-replica) login servers such as foo.eastus2.geo.azurecr.io. Add an
+		// explicit entry for the exact parsed target login server so service-principal
+		// pulls from regional endpoints are authenticated, mirroring the managed
+		// identity path above.
+		spCred := v1.AuthConfig{
+			Username: a.config.AADClientID,
+			Password: a.config.AADClientSecret,
+		}
+		response.Auth[targetloginServer] = spCred
```
nilo19: Does it create a duplicated entry?
1. contoso.azurecr.io       → SP credentials
2. *.azurecr.io             → identical SP credentials
3. *.azurecr.*              → anonymous
When the first authenticated attempt succeeds, there is little visible impact. When it fails, he second attempt repeats the same credentials. So it seems introduce duplicated retries when failed.

pkg/credentialprovider/azure_credentials.go
nilo19: Can it be extracted out of for loop and reused?

pkg/credentialprovider/azure_credentials.go
```diff
@@ -49,8 +49,12 @@ const (
...
 var (
 	containerRegistryUrls = []string{"*.azurecr.io", "*.azurecr.cn", "*.azurecr.de", "*.azurecr.us"}
-	// a valid acr image starts with alphanumerics, followed by corresponding acr domain name.
-	acrRE = regexp.MustCompile(`^[a-zA-Z0-9]+\.(azurecr\.io|azurecr\.cn|azurecr\.de|azurecr\.us)`)
+	// a valid acr image starts with alphanumerics, followed by an optional
+	// regional login server segment (.<region>.geo), then the corresponding
+	// acr domain name. Examples:
+	//   foo.azurecr.io                 (global login server)
+	//   foo.eastus2.geo.azurecr.io     (regional login server)
+	acrRE = regexp.MustCompile(`^[a-zA-Z0-9]+(\.[a-zA-Z0-9]+\.geo)?\.(azurecr\.io|azurecr\.cn|azurecr\.de|azurecr\.us)`)
```
nilo19: Is it by design to not support hyphens? for exmaple `demo-abc123.eastus2.geo.azurecr.io`

pkg/credentialprovider/azure_credentials.go
lizMSFT: Made requested changes

pkg/credentialprovider/azure_credentials.go
lizMSFT: Fixed, the explicit entry is now only added for regional/geo hosts via  isRegionalACRLoginServer

pkg/credentialprovider/azure_credentials.go
lizMSFT: Not by design — this is intentional now to support ACR Domain Name Label (DNL) registries, whose login server has one hyphen ( <name>-<hash>.azurecr.io ). Updated the regex to allow a single hyphen and added test cases for both DNL and DNL+geo hosts.

PR #10731, opened 2026-07-25
fix: match ACR regional endpoints and DNL-enabled login servers in credential provider

### LINKED ISSUE #9975: ACR credential provider regex doesn't handle regional endpoints or registries w/ DNL enabled

pujitha24:
/kind bug

The acr-credential-provider's `acrRE` regex, used to detect whether an
image reference targets Azure Container Registry, only matched a single
alphanumeric label before the `azurecr.*` domain (e.g. `foo.azurecr.io`).
It rejected two valid ACR login server formats:

- Registries with Domain Name Label (DNL) enabled, whose login server
  includes a dash-separated hash, e.g. `myregistry-abc123.azurecr.io`.
- ACR regional/replica endpoints, which add one extra DNS label, e.g.
  `myregistry.eastus.azurecr.io`.

When `parseACRLoginServerFromImage` fails to recognize such a hostname,
`GetCredentials` returns an empty `Auth` map for that image (see
azure_credentials.go), so kubelet attempts an anonymous pull. For
private images hosted on a DNL-enabled or regional-endpoint registry,
that pull then fails with an authentication error. Public images and
registries using the existing plain `registry.azurecr.io` format are
unaffected.

#### Approach

Update `acrRE` to allow dashes within a DNS label and one optional
additional dot-separated label before the `azurecr.*` suffix, while
keeping the domain suffix itself a literal, anchored match. The
existing anti-spoofing behavior in `parseACRLoginServerFromImage`
(requiring whatever follows the matched registry to be empty or start
with `/`) is untouched, so hostnames like `foo.azurecr.io.evil.com` are
still correctly rejected.

Fixes #9975

Added test cases for a DNL-style login server (`foo-abc123.azurecr.io`)
and a regional endpoint (`foo.eastus.azurecr.io`), and updated the
`foo-azurecr-io.azurecr.cn` case, which previously documented the bug
(expected `""`) and now correctly matches. Verified the existing
`foo.azurecr.io.azurecr.cn` anti-spoofing test case still returns `""`
unchanged.

```release-note
Fixed the acr-credential-provider incorrectly treating images hosted on Azure Container Registry regional endpoints or registries with Domain Name Label (DNL) enabled as non-ACR images, which caused private image pulls from those registries to fail with an authentication error.
```

```docs

```

Validation: `go build ./pkg/credentialprovider/...` and
`go test ./pkg/credentialprovider/...` both pass (all existing and new
subtests of TestParseACRLoginServerFromImage pass).

Signed-off-by: Pujitha Paladugu <10557236+pujitha24@users.noreply.github.com>

### DIFF
--- a/pkg/credentialprovider/azure_credentials.go
+++ b/pkg/credentialprovider/azure_credentials.go
@@ -49,8 +49,11 @@ const (
 
 var (
 	containerRegistryUrls = []string{"*.azurecr.io", "*.azurecr.cn", "*.azurecr.de", "*.azurecr.us"}
-	// a valid acr image starts with alphanumerics, followed by corresponding acr domain name.
-	acrRE = regexp.MustCompile(`^[a-zA-Z0-9]+\.(azurecr\.io|azurecr\.cn|azurecr\.de|azurecr\.us)`)
+	// a valid acr image starts with a DNS label (allowing dashes, e.g. registries with a
+	// Domain Name Label suffix such as "registry-hash.azurecr.io"), optionally followed by
+	// one more label for regional endpoints (e.g. "registry.<region>.azurecr.io"), followed
+	// by the corresponding acr domain name.
+	acrRE = regexp.MustCompile(`^[a-zA-Z0-9](?:[a-zA-Z0-9-]*[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]*[a-zA-Z0-9])?)?\.(azurecr\.io|azurecr\.cn|azurecr\.de|azurecr\.us)`)
 )
 
 // CredentialProvider is an interface implemented by the kubelet credential provider plugin to fetch
--- a/pkg/credentialprovider/azure_credentials_test.go
+++ b/pkg/credentialprovider/azure_credentials_test.go
@@ -332,7 +332,15 @@ func TestParseACRLoginServerFromImage(t *testing.T) {
 		},
 		{
 			image:    "foo-azurecr-io.azurecr.cn",
-			expected: "",
+			expected: "foo-azurecr-io.azurecr.cn",
+		},
+		{
+			image:    "foo-abc123.azurecr.io/bar/image:version",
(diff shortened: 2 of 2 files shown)


pujitha24:
The two failing jobs here (`pull-cloud-provider-azure-e2e-ccm-capz` and `pull-cloud-provider-azure-e2e-ccm-vmss-capz`) look unrelated to this change — they're failing across other open PRs as well, e.g. #10737, #10693, #10730 and #10720. I retried them a few times over the last three days and the same two jobs fail each run.

This PR only touches `pkg/credentialprovider` (+5/-2) — ACR endpoint matching — so it shouldn't affect cluster e2e provisioning. All other 20 checks are green, including `pull-cloud-provider-azure-e2e-capz` and `pull-cloud-provider-azure-e2e-ccm-dualstack-vmss-capz`.

Happy to rebase or investigate further if it would help.

pujitha24:
Closing this — #10720 already landed a more complete fix for the same regex gap (issue #9975), including the correct `.geo.` regional-endpoint format validated against live AKS, plus the service-principal auth-path fix and e2e coverage that this PR didn't have. Thanks for the eyes on this; no need to carry two fixes for the same problem.

PR #10737, opened 2026-07-27
Add Config.VNetResourceGroupOrDefault to the ServiceGateway difftracker

georgeedward2000:
#### What type of PR is this?

/kind feature

`difftracker.Config` carries `VNetResourceGroup`, which is optional: it is only set for BYO-VNet
clusters whose Virtual Network lives outside the cluster resource group. When it is empty the VNet
is in `ResourceGroup`, so every consumer has to remember to fall back. Anyone who forgets builds an
ARM ID with an empty resource group segment:

    /subscriptions/<sub>/resourceGroups//providers/Microsoft.Network/virtualNetworks/<vnet>

This adds a single accessor so the fallback lives with the config instead of being repeated at each
call site, and documents the field.

The rest of the provider already resolves the VNet resource group this way:

- `getVnetResourceID` in `pkg/provider/azure_loadbalancer_backendpool.go`
- `pkg/provider/azure_privatelinkservice.go`

so this keeps the ServiceGateway config consistent with existing behaviour.

No existing behaviour changes: the method is additive and nothing calls it yet.

NONE

Small, self-contained piece of the in-progress ServiceGateway work, split out so the follow-up PR
that adds the Azure operations (which builds subnet and backend pool resource IDs) stays focused.
`config_test.go` is new because `config.go` had no test file yet; it covers the set, unset, equal
and both-empty cases.

```release-note
NONE
```
Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.:

```release-note
NONE
```

### DIFF
--- a/pkg/provider/servicegateway/difftracker/config.go
+++ b/pkg/provider/servicegateway/difftracker/config.go
@@ -44,7 +44,9 @@ type Config struct {
 	// Virtual Network name (required for backend pool configuration)
 	VNetName string
 
-	// Virtual Network resource group
+	// Resource group that contains the Virtual Network. Optional: when empty, the VNet is
+	// assumed to live in ResourceGroup (the common case). Set this for BYO-VNet clusters whose
+	// VNet is in a separate resource group (mirrors the cloud config's vnetResourceGroup).
 	VNetResourceGroup string
 }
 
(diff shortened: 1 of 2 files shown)


Liunardy:
[lgtm](https://github.com/kubernetes-sigs/cloud-provider-azure/pull/10737#issuecomment-5116498987) and [approved](https://github.com/kubernetes-sigs/cloud-provider-azure/pull/10737#issuecomment-5116503777) by David.
`pkg/provider/servicegateway` only changes. 
Only `Private link service should support service annotation 'service.beta.kubernetes.io/azure-pls-ip-configuration-ip-address' [PLS]` failed because of recent known issue. Merging manually.

PR #10761, opened 2026-07-28
ServiceGateway (SLB) reconciliation engine and e2e suite

georgeedward2000:
#### What type of PR is this?

/kind feature

The common-ground integration merged in #10714 wired ServiceGateway into the provider but left
`pkg/provider/servicegateway/difftracker` as placeholders: `InitializeFromCluster` returned an empty
`DiffTracker` and the state-diffing methods were no-ops. This PR fills that package in, so a cluster
running with `serviceGatewayEnabled` actually programs NRP.

The engine keeps the desired state (Kubernetes) and the actual state (NRP) side by side, diffs them,
and drives the difference through Azure:

- `initialization.go` rebuilds both states at startup from the cluster and from NRP, recovers
  finalizers and Service external IPs that a crash left behind, and cleans up orphaned Azure
  resources before serving traffic.
- `engine.go` / `service_updater.go` own the per-Service state machine: validate the inbound spec,
  add the cleanup finalizers, create the PIP / LoadBalancer / NAT gateway, then register the Service
  with the ServiceGateway.
- `pods.go` / `endpoints.go` track egress pods and EndpointSlices and translate them into backend
  pool and address-location updates.
- `locations_updater.go` batches address-location changes, with retry and jitter for transient NRP
  failures.
- `finalizers.go` holds Services and egress pods in Terminating until Azure has actually released
  their resources, so nothing leaks when a namespace is deleted.
- `metrics.go` reports engine state and Azure operation outcomes.

Two behaviours are worth calling out because they are not obvious from the diff:

- NRP answers `updateServices` and `updateAddressLocations` with `200 OK`, which the generated
  armnetwork client rejects because the spec only lists 202 and 204. Rather than patching the
  vendored SDK, `azure_operations.go` recognises that specific response and treats it as the
  synchronous completion it is. Any real failure keeps its own 4xx/5xx status and still propagates.
- A Service the engine cannot support (named targetPort, internal LoadBalancer, unsupported
  protocol) is rejected terminally and gets a warning Event naming the reason. Returning the error
  alone would surface only a generic `SyncLoadBalancerFailed`.

The e2e suite lives in `tests/e2e/servicegateway` as its own Ginkgo suite rather than inside
`tests/e2e/network`, so it can be run on its own against a ServiceGateway-enabled cluster:

    go test -c -o slb.test ./e2e/servicegateway

NONE

This PR is confined to `pkg/provider/servicegateway` and `tests/e2e/servicegateway`: no file outside
those two directories is changed. The two test adjustments that previously sat here landed
separately in #10738 and are already in master.

Suggested reading order: `types.go` and `engine.go` for the state model, then `initialization.go`
for startup and recovery, then `service_updater.go` for the per-Service lifecycle.

Testing: 492 unit tests across `pkg/provider/servicegateway` and its `difftracker` subpackage, plus
115 e2e specs run against live AKS clusters with ServiceGateway enabled — single-stack and
dual-stack, including CCM crash recovery, finalizer draining, bulk service churn, and a 100-pod
egress workload.

```release-note

```

```docs

```


pkg/provider/servicegateway/difftracker/resource_helpers.go
```diff
@@ -0,0 +1,539 @@
...
+	}
+	return inboundConfig, nil
+}
+
+// ValidateInboundConfig rejects inbound Service specs the PodIP backend pool cannot support,
+// returning an *InboundConfigValidationError whose Reason is the Event reason to surface. Without
+// this fast check the engine would terminal-park the create asynchronously with no visible cause.
+// A nil config (a port-less Service) is treated as valid; the caller handles that case separately.
+func ValidateInboundConfig(config *InboundConfig) error {
+	if config == nil {
```
david-kow, line 600: we should also check for not supported annotations. in separate pr though

tests/e2e/servicegateway/slb_basic_test.go
```diff
@@ -0,0 +1,347 @@
...
+		}
+		Expect(pipFound).To(BeTrue(), "Load Balancer should reference our Public IP")
+		utils.Logf("  Azure Portal: https://portal.azure.com/#@/resource%s", serviceLB.ID)
+
+		By("Verifying service details from Kubernetes")
+		updatedService, err := cs.CoreV1().Services(ns.Name).Get(context.TODO(), serviceName, metav1.GetOptions{})
+		Expect(err).NotTo(HaveOccurred())
+
+		// Check if external IP was eventually updated in Kubernetes (optional - might not be updated)
+		if len(updatedService.Status.LoadBalancer.Ingress) > 0 {
```
david-kow: why this is ok? shouldn't test fail here?

pkg/provider/servicegateway/difftracker/engine.go
```diff
@@ -0,0 +1,2134 @@
...
+				} else {
+					// Deterministic, spec-driven update failure (e.g. unsupported protocol, port
+					// out of range, dual-stack). Retrying cannot succeed, so park the service
+					// instead of looping forever. Its existing Azure resources keep the
+					// last-applied config; a later UpdateService with a changed spec clears the park.
+					opState.CreationFailedTerminal = true
+					opState.State = StateNotStarted
+					opState.LastAttempt = time.Now().Format(time.RFC3339)
+					dt.logger.V(4).Info("Parked service after non-retryable update error", "err", err, "service", serviceUID)
+					dt.checkInitializationCompleteLocked()
```
david-kow, line 901: should we log this in user-visible way?

pkg/provider/servicegateway/difftracker/engine.go
```diff
@@ -0,0 +1,2134 @@
...
+					opState.State = StateNotStarted
+					opState.LastAttempt = time.Now().Format(time.RFC3339)
+					dt.logger.V(5).Info("Re-dispatched service creation after desired config drifted during a terminal failure", "service", serviceUID)
+					dt.triggerServiceUpdater()
+					dt.checkInitializationCompleteLocked()
+				} else {
+					// Deterministic, spec-driven failure (e.g. unsupported protocol, port out
+					// of range). Retrying cannot succeed, so park the service instead of looping
+					// forever. A later UpdateService with a changed spec clears the park.
+					opState.CreationFailedTerminal = true
```
david-kow, line 1092: log in a way user can see it?

pkg/provider/servicegateway/runtime.go
david-kow, line 65: should we create runtime even if this is false?

pkg/provider/servicegateway/difftracker/engine.go
```diff
@@ -0,0 +1,2206 @@
...
+	// and the Azure delete here is asynchronous. Provisioning from there re-creates the LoadBalancer,
+	// Public IP and ServiceGateway registration for a Service that is going away; once our finalizer
+	// clears, the object disappears and those resources are left in Azure with nothing tracking them.
+	if service.DeletionTimestamp != nil {
+		dt.logger.V(4).Info("Skipped reconciling a Service that is being deleted",
+			"namespace", service.Namespace, "service", service.Name, "serviceUID", serviceUID)
+		return nil
+	}
+
+	dt.logger.V(2).Info("Reconciling inbound Service", "serviceUID", serviceUID)
```
david-kow, line 108: Should we also log namespace and name?

pkg/provider/servicegateway/difftracker/resource_helpers.go
```diff
@@ -0,0 +1,624 @@
...
+// Both the runtime path (ReconcileInboundService) and the startup path (reconcileServices) must
+// call this. A Service admitted by only one of them is provisioned inconsistently across a CCM
+// restart: an internal-LB request rejected at runtime would be rebuilt with Scope="Public".
+//
+// The internal-LB annotation is matched case-insensitively through the shared consts helper, as
+// every other Azure annotation in this provider is. An exact string comparison would treat "True"
+// as absent and let the request through.
+func AdmitInboundService(service *v1.Service) (*InboundConfig, error) {
+	if service == nil {
+		return nil, nil
```
david-kow: is that ever expected?

pkg/provider/servicegateway/difftracker/resource_helpers.go
```diff
@@ -0,0 +1,624 @@
...
+		set.Insert(item)
+	}
+	return set
+}
+
+// ExtractInboundConfigFromService creates InboundConfig from a Kubernetes Service
+// This is shared between initialization and the provider layer
+func ExtractInboundConfigFromService(service *v1.Service) *InboundConfig {
+	if service == nil || len(service.Spec.Ports) == 0 {
+		return nil
```
david-kow, line 432: do we have test coverage for service being created without ports, and then updated with ports to check if the reconciliation works correctly? what about deletion path - if ports are removed, do we only remove lb rule?

pkg/provider/servicegateway/difftracker/resource_helpers.go
```diff
@@ -0,0 +1,624 @@
...
+	// A malformed or out-of-range value is left nil here and rejected by ValidateInboundConfig,
+	// which carries the Event reason.
+	if idleTimeout, err := consts.Getint32ValueFromK8sSvcAnnotation(service.Annotations, consts.ServiceAnnotationLoadBalancerIdleTimeout, func(val *int32) error {
+		if *val < inboundIdleTimeoutMinMinutes || *val > inboundIdleTimeoutMaxMinutes {
+			return fmt.Errorf("idle timeout must be a whole number of minutes between %d and %d, got %d",
+				inboundIdleTimeoutMinMinutes, inboundIdleTimeoutMaxMinutes, *val)
+		}
+		return nil
+	}); err != nil {
+		config.InvalidIdleTimeout = to.Ptr(service.Annotations[consts.ServiceAnnotationLoadBalancerIdleTimeout])
```
david-kow, line 503: tbh this is a bit weird - extracting config fuct does the validation and we create an additional field in the config struct just so validate func can throw on something. why we can't just rewrite the service config to our config object (so we capture all fields we care about) and then validate errors if validation fails on any

pkg/provider/servicegateway/difftracker/engine.go
```diff
@@ -0,0 +1,2206 @@
...
+	inboundConfig, err := AdmitInboundService(service)
+	if err != nil {
+		return err
+	}
+	if inboundConfig == nil {
+		return nil
+	}
+
+	config := NewInboundServiceConfig(serviceUID, inboundConfig)
+	config.Namespace = service.Namespace
```
david-kow, line 119: why we can't have it in the NewInbound...? I see that sometimes these are not set, are they not always available? Might have been cleaner if this obj is always initialized

pkg/provider/servicegateway/difftracker/engine.go
```diff
@@ -0,0 +1,2206 @@
...
+	config.Name = service.Name
+
+	if dt.IsServiceTracked(serviceUID) {
```
david-kow: I have mixed feelings here - we take the lock, release the lock, make a decision and take look again. Isn't there something that can happen in-between that would violate our assumptions? Maybe do a single func under single lock?

pkg/provider/servicegateway/difftracker/engine.go
```diff
@@ -0,0 +1,2206 @@
...
+		Config:        config,
+		State:         StateNotStarted,
+		RetryCount:    0,
+		LastAttempt:   time.Now().Format(time.RFC3339),
+		CreatedAt:     time.Now(),
+		CorrelationID: uuid.NewString(),
+	}
+
+	// Release lock before triggering to avoid lock contention
+	dt.mu.Unlock()
```
david-kow, line 206: hm, that worries me for two reasons: a) not using defer is a risk of missing unlock in some branch somewhere; b) this makes much more difficult to reason about locking: what if something gets the lock between here and getting the lock in seedInboundEndpointsFromCache?

pkg/provider/servicegateway/difftracker/locations_updater.go
```diff
@@ -0,0 +1,374 @@
...
+		case isOperationSucceeded:
+			lu.failureCount = 0
+		case terminalSyncErr:
+			lu.failureCount = 0
+		case atomic.LoadInt32(&lu.diffTracker.isInitializing) == 1 &&
+			lu.failureCount >= getMaxInitLocationSyncAttempts():
+			// Stop retrying an NRP error that is not clearing. Work only a successful sync can
+			// advance still holds the initialization gate, so startup fails rather than running
+			// against NRP state this sync never reconciled.
+			lu.logger.Error(lastSyncErr, "Gave up initial NRP location sync after repeated transient failures",
```
david-kow: not sure what is supposed to happen next

pkg/provider/servicegateway/difftracker/pods.go
```diff
@@ -0,0 +1,469 @@
...
+	}
+}
+
+func (dt *DiffTracker) SetUpPodInformer(stopCh <-chan struct{}) error {
+	klog.V(2).Infof("setUpPodInformerForEgress: Setting up pod informer with label selector: %s", consts.PodLabelServiceEgressGateway)
+
+	// Create a separate informer factory with label selector to filter pods at the API server
+	podInformerFactory := informers.NewSharedInformerFactoryWithOptions(
+		dt.kubeClient,
+		ResyncPeriod(12*time.Hour)(),
```
david-kow: lets dial it down to 1 hour - we shouldn't need it at all, but I'd like to have this path exercised more often if we have it

david-kow (approved):
lgtm

david-kow (approved):
lgtm

Liunardy:
PR description for release notes (Does this PR introduce a user-facing change?) and docs (Additional documentation) should be inside these blocks:
````
```release-note

```

```docs

```
````
If there are no docs, leave it empty.

File: pkg/provider/servicegateway/difftracker/pods.go (hunk 13 above)
Comment (georgeedward2000) [reply pkg/provider/servicegateway/difftracker/pods.go]: I addressed it and am resolving it. Happy to reopen it if there's anything else we should discuss.

File: pkg/provider/servicegateway/difftracker/locations_updater.go (hunk 12 above)
Comment (georgeedward2000) [reply pkg/provider/servicegateway/difftracker/locations_updater.go]: Fixed: it now calls backoffAndRetry() instead of giving up. The attempt limit is only a reporting threshold. It logs once per outage, and the retry backoff is capped and jittered, so it can't hot-loop.

I also removed the init timeout. Failing startup would only result in repeated restarts against the same unavailable NRP. The initialization wait is now unbounded and emits an error-level log every minute, keeping CCM blocked and noisy rather than allowing it to start in an invalid state.

File: pkg/provider/servicegateway/difftracker/engine.go (hunk 10 above)
Comment (georgeedward2000) [reply pkg/provider/servicegateway/difftracker/engine.go]: Agreed, and fixed. That check-then-branch pattern is gone.

EnsureLoadBalancer now just calls ReconcileInboundService, which hands off to UpdateService(config). UpdateService determines tracked vs. untracked state under its own lock and delegates to AddService when the service is untracked. That removes the possibility of making a decision based on a stale read.

IsServiceTracked still exists, but only for GetLoadBalancer, where it's used as a read-only check. Nothing branches on it anymore.

Single code path, single lock, just as you suggested.

File: pkg/provider/servicegateway/difftracker/resource_helpers.go (hunk 1 above)
Comment (georgeedward2000) [reply pkg/provider/servicegateway/difftracker/resource_helpers.go:600]: Sounds good, we will defer it to a later PR.

File: tests/e2e/servicegateway/slb_basic_test.go (hunk 2 above)
Comment (georgeedward2000) [reply tests/e2e/servicegateway/slb_basic_test.go]: You're right, it shouldn't have been optional. That if len(...) > 0 check meant the test could pass even if the IP was never assigned. 

Fixed: the test now waits for the status to be populated instead of taking a single snapshot, and it asserts the expected IP unconditionally.

File: pkg/provider/servicegateway/difftracker/engine.go (hunk 4 above)
Comment (georgeedward2000) [reply pkg/provider/servicegateway/difftracker/engine.go:1092]: This is resolved now. It was previously log-only; now it emits a Warning event on the Service, so it shows up in kubectl describe svc:

```
Warning  ServiceGatewayConfigurationRejected
Service creation failed with a non-retryable error and will not be retried
until the Service spec changes: <reason>
```

The event is emitted outside the locked section via serviceParkNotice, so we don't make API calls while holding dt.mu.

Admission rejections (for example, named targetPort, internal LB, or unsupported protocol) follow the same pattern through WarningEventError, ensuring users get visible feedback directly on the Service instead of having to dig through controller logs.

File: pkg/provider/servicegateway/runtime.go
Comment (georgeedward2000) [reply pkg/provider/servicegateway/runtime.go:65]: That is right, that is unnecessary. I deferred it to a later PR.

File: pkg/provider/servicegateway/difftracker/engine.go (hunk 3 above)
Comment (georgeedward2000) [reply pkg/provider/servicegateway/difftracker/engine.go:901]: same as the other comment

File: pkg/provider/servicegateway/difftracker/engine.go (hunk 5 above)
Comment (georgeedward2000) [reply pkg/provider/servicegateway/difftracker/engine.go:108]: Sounds good.  I deferred it to a later PR.

File: pkg/provider/servicegateway/difftracker/resource_helpers.go (hunk 6 above)
Comment (georgeedward2000) [reply pkg/provider/servicegateway/difftracker/resource_helpers.go]: No, it was never expected. A nil Service reaching admission is a programming error, and the previous nil, nil return effectively treated it as "not ours" and silently ignored it.

Now we fail fast:

```
if service == nil {
    return nil, fmt.Errorf("cannot admit a nil Service")
}
```


ReconcileInboundService also guards against nil at the entry point, so this is mostly belt-and-braces protection rather than a real execution path. It ensures we surface the bug immediately instead of quietly skipping it.

pkg/provider/servicegateway/difftracker/resource_helpers.go
georgeedward2000: Yes on both. I added tests for those cases after this comment.

pkg/provider/servicegateway/difftracker/engine.go
georgeedward2000: We deferred. Nothing required for now, let's discuss for an eventual PR.

pkg/provider/servicegateway/difftracker/engine.go
georgeedward2000: (b) Benign. The operation is inserted into pendingServiceOps before the unlock, and updateEndpointsLocked re-resolves state under its own lock rather than relying on an earlier read. If a delete arrives during that window, the seed is applied against the current state. The unlock is required because both the seed path and triggerServiceUpdater take the same lock, so holding it would deadlock.

(a) Fair point. engine.go currently has 31 manual unlocks versus 8 defers, and there's nothing preventing a future early return from missing one. The common pattern is "mutate under lock, then call out," which doesn't fit well with a simple defer-based structure.

We could refactor this into a locked helper that returns a decision, followed by an unlocked execution path. That would allow defer to be used consistently throughout. Happy to do that as a follow-up if you'd prefer it before merge.

david-kow (approved):

Liunardy:
/lgtm
/approve
[lgtm and approved](https://github.com/kubernetes-sigs/cloud-provider-azure/pull/10761#pullrequestreview-4869210978) by David.
`pkg/provider/servicegateway`, new `tests/e2e/servicegateway/` folder and comment on previously added tests for service gateway only changes.

PR #10812, opened 2026-08-12
feat: support NFS encryption in transit for new storage accounts

rnhan:
#### What type of PR is this?

/kind feature

Azure Files now supports per-protocol encryption in transit (EiT) for NFS.  Enforcement is service side at the storage account file service level. The AzureFile CSI Driver needs to be able to enable this on accounts for storageclasses that request it, at create time only for safety.

Changes:
- Add `AccountOptions.IsNFSEncryptionInTransitEnabled` to configure per-protocol encryption in transit for NFS on a storage account's file service (`ProtocolSettings.Nfs.EncryptionInTransit.Required`).
- Updates account matching (`getStorageAccounts`) with `isNFSEncryptionInTransitEnabledEqual` so an account with NFS encryption in transit enabled is not reused for a request that does not require it (and vice versa), preventing plaintext NFS mounts from landing on an encryption-required account.

```release-note
Added `AccountOptions.IsNFSEncryptionInTransitEnabled` to enable per-protocol NFS encryption in transit on newly created storage accounts.  The field is optional, leaving it as nil preserves the existing behavior.
```

### DIFF
--- a/pkg/provider/storage/azure_storageaccount.go
+++ b/pkg/provider/storage/azure_storageaccount.go
@@ -80,6 +80,7 @@ type AccountOptions struct {
 	AllowCrossTenantReplication             *bool
 	IsMultichannelEnabled                   *bool
 	IsSmbOAuthEnabled                       *bool
+	IsNFSEncryptionInTransitEnabled         *bool
 	KeyName                                 *string
 	KeyVersion                              *string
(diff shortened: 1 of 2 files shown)


andyzhangx (approved):
/lgtm

PR #10814, opened 2026-08-13
fix: fall back to an IP-derived name for backend addresses with no cached node name

### LINKED ISSUE #9803: Node name in cache could be empty
#### What happened:

Node name in our cache could be empty for unknown reason. https://github.com/kubernetes-sigs/cloud-provider-azure/blob/5450c2a88e9aa1a1f92c05d24557eda8bb99114d/pkg/provider/azure.go#L830

In this case, we should assign a name for loadbalancerbackendaddress to prevent error. https://github.com/kubernetes-sigs/cloud-provider-azure/blob/master/pkg/provider/azure_loadbalancer_backendpool.go#L857

#### Environment:

- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

pujitha24:
#### What type of PR is this?

/kind bug

When `addNodeIPAddressesToBackendPool` builds a new `armnetwork.LoadBalancerBackendAddress`
for a node IP, it looks up the node name from `az.nodePrivateIPToNodeNameMap` and uses it
as the address `Name`. If that lookup misses, `Name` is set to a pointer to an empty
string.

This is reachable today: the local-service backend-pool-updater test suite
(`pkg/provider/azure_local_services_test.go`, exercising
`pkg/provider/azure_local_services.go`) seeds `nodePrivateIPs` directly without going
through `updateNodeCaches`, so `nodePrivateIPToNodeNameMap` stays empty and every backend
address built on that path previously got `Name: ""`. In production, both maps are
populated together by the same node-informer callback (`pkg/provider/azure.go`,
`updateNodeCaches`), so a persistently empty name there would only happen for a node whose
private-IP cache entry exists ahead of/without its name-map entry (e.g. a timing/race gap
during cache population), matching the "for unknown reason" note in the linked issue. This
PR does not claim to have reproduced that race live; the fix protects the code path (used by
both production and this test suite) that hits the empty-name case, and is validated by a
unit test that reproduces "cache miss -> empty name" directly.

An empty-named backend address risks being rejected, or colliding with other empty-named
addresses, when the pool is sent to the Azure API. This PR falls back to a name derived
from the IP address (e.g. `10.0.0.1` -> `ip-10-0-0-1`) whenever the cached node name is
empty, so the address always gets a non-empty, distinguishable name. The `ip-` prefix
avoids colliding with a real node whose name happens to already look IP-derived.

Fixes #

- This is a defensive fallback, not a fix for the root cause of why the cache entry could
  be missing in production (that root cause is not confirmed — see discussion above).
- Also updated `pkg/provider/azure_local_services_test.go`'s
  `getTestBackendAddressPoolWithIPs` helper, which previously hardcoded
  `Name: ptr.To("")` for every synthesized backend address, to generate the same
  fallback name production code now generates, since that test path never populates
  `nodePrivateIPToNodeNameMap`.
- Validation: `go build ./...` passes. `go test ./pkg/provider/...` passes (full
  package, no regressions). Added
  `TestAddNodeIPAddressesToBackendPoolGeneratesNameWhenNodeNameMissing`, confirmed it
  fails with `expected "ip-10-0-0-1", actual ""` when the fix is reverted and passes
  with the fix applied. `golangci-lint run ./pkg/provider/...` shows no new findings on
  the changed lines (pre-existing goconst/gosec/govet findings elsewhere in the package
  are unrelated to this diff).

```release-note
Fixed a potential issue where a load balancer backend address could be created with an empty name when the node name was missing from the internal node-IP cache, by falling back to a name derived from the node's IP address.
```

```docs

```

Fixes #9803

### DIFF
--- a/pkg/provider/azure_loadbalancer_backendpool.go
+++ b/pkg/provider/azure_loadbalancer_backendpool.go
@@ -707,6 +707,13 @@ func (az *Cloud) addNodeIPAddressesToBackendPool(backendPool *armnetwork.Backend
 	for _, ipAddress := range nodeIPAddresses {
 		if !hasIPAddressInBackendPool(backendPool, ipAddress) {
 			name := az.nodePrivateIPToNodeNameMap[ipAddress]
+			if name == "" {
+				// The node name in cache could be empty for unknown reasons. Fall back to a name
+				// derived from the IP address, prefixed to avoid colliding with a real node name,
+				// so the backend address is not created with an empty name.
+				name = "ip-" + strings.NewReplacer(".", "-", ":", "-").Replace(ipAddress)
+				logger.V(2).Info("Node name not found in cache for IP address, generated a name for the backend address", "ip", ipAddress, "generatedName", name)
+			}
 			logger.V(4).Info("adding node to the backend pool", "ip", ipAddress, "backendPoolName", ptr.Deref(backendPool.Name, ""))
 			addresses = append(addresses, &armnetwork.LoadBalancerBackendAddress{
 				Name: ptr.To(name),
--- a/pkg/provider/azure_loadbalancer_backendpool_test.go
+++ b/pkg/provider/azure_loadbalancer_backendpool_test.go
@@ -958,6 +958,24 @@ func TestAddNodeIPAddressesToBackendPoolDeduplicates(t *testing.T) {
 	}
 }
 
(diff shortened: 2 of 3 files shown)


pujitha24:
/assign @bridgetkromhout

Still waiting on ok-to-test and a first look whenever you have a chance.

pujitha24:
Closing this to keep my open PR queue manageable - it has been open a while without review, and I would rather not leave stale PRs sitting in your queue. The change itself still applies; happy to reopen and rebase if it is useful to you.

PR #10815, opened 2026-08-13
fix: include priority 4096 in managed security rule allocation range

### LINKED ISSUE #9906: Align max managed security priority behavior
#### What happened:

https://github.com/kubernetes-sigs/cloud-provider-azure/blob/master/pkg/provider/securitygroup/securityrule.go#L41

https://github.com/kubernetes-sigs/cloud-provider-azure/blob/5909ab1d1fa9ecafbb1c624cc18cee12c2fb65c1/pkg/provider/securitygroup/securitygroup.go#L113

We need to make the above behaviors the same. Max priority 4096 should/shoult not be counted in as the managed range.

#### Environment:

- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

pujitha24:
#### What type of PR is this?

/kind bug

`IsManagedSecurityRule` (pkg/provider/securitygroup/securityrule.go) treats the
managed security rule priority range as `[500, 4096]`, inclusive of the maximum
priority 4096. However, `RuleHelper.nextRulePriority`
(pkg/provider/securitygroup/securitygroup.go), which allocates priorities for
rules the cloud-provider creates, used an exclusive upper bound and could only
ever assign priorities in `[500, 4095]` in both allocation directions (from
start and from end). Azure's valid NSG rule priority range is 100-4096, so
priority 4096 is a legitimate allocatable value; excluding it from allocation
while still treating it as "managed" was an inconsistency between the two
functions.

This PR rewrites `nextRulePriority`'s loop to use inclusive bounds on both
ends, so it can now allocate priority 4096, matching `IsManagedSecurityRule`'s
range in both allocation directions.

This is a narrow allocation off-by-one affecting only the single boundary
priority 4096; it does not change behavior for any existing rule or any
priority below 4096, and is not a crash or data-loss issue.

Fixes #

Validation performed locally:
- `go build ./...` passes.
- Added `TestRuleHelper_NextRulePriority_IncludesMaximumPriority` in
  pkg/provider/securitygroup/securityrule_test.go, which fills every priority
  except 4096 and asserts `nextRulePriority` can still find it in both
  directions. This test fails on the pre-fix code (returns "security rule
  priority exhausted") and passes after the fix.
- `go test ./pkg/provider/securitygroup/...` passes, including the updated
  `TestSecurityGroupHelper_AddRuleForDenyAll`, whose expected priority for a
  deny-all rule allocated into an otherwise-empty managed range changes from
  the old (buggy) 4095 to the correct 4096.
- `golangci-lint run ./pkg/provider/securitygroup/...` shows only
  pre-existing `goconst` warnings in the test file, unchanged from master
  (verified via `git stash`).

Note: `pkg/provider/azure_standard.go` has a `getNextAvailablePriority`
function with the same exclusive-upper-bound pattern, but it appears to be
unreferenced outside its own unit test (dead code), so it was left out of
this narrowly-scoped fix.

```release-note
Fixed an inconsistency where the cloud-provider's security rule priority allocator could never assign the maximum valid priority (4096) to a new managed rule, even though that priority is considered part of the managed range for cleanup purposes.
```

```docs

```

Fixes #9906

### DIFF
--- a/pkg/provider/securitygroup/securitygroup.go
+++ b/pkg/provider/securitygroup/securitygroup.go
@@ -110,22 +110,24 @@ const (
 
 // nextRulePriority returns the next available priority for a new rule.
 // It takes a preference for whether to start from the beginning or end of the priority range.
+// The priority range is inclusive of both consts.LoadBalancerMinimumPriority and
+// consts.LoadBalancerMaximumPriority, matching the range considered managed by IsManagedSecurityRule.
 func (helper *RuleHelper) nextRulePriority(prefer rulePriorityPrefer) (int32, error) {
 	var (
-		init, end = consts.LoadBalancerMinimumPriority, consts.LoadBalancerMaximumPriority
-		delta     = 1
+		init, end = int32(consts.LoadBalancerMinimumPriority), int32(consts.LoadBalancerMaximumPriority)
+		delta     = int32(1)
 	)
 	if prefer == rulePriorityPreferFromEnd {
-		init, end, delta = end-1, init-1, -1
+		init, end, delta = end, init, -1
 	}
 
-	for init != end {
-		p := int32(init)
-		if _, found := helper.priorities[p]; found {
-			init += delta
-			continue
+	for p := init; ; p += delta {
+		if _, found := helper.priorities[p]; !found {
+			return p, nil
 		}
-		return p, nil
+		if p == end {
+			break
+		}
 	}
 
 	return 0, ErrSecurityRulePriorityExhausted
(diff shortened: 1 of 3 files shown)


pujitha24:
/assign @nilo19

Would appreciate a review when you have a chance.

pujitha24:
Closing this to keep my open PR queue manageable - it has been open a while without review, and I would rather not leave stale PRs sitting in your queue. The change itself still applies; happy to reopen and rebase if it is useful to you.

PR #10816, opened 2026-08-13
fix: Included ExtendedLocation when update VMSS

wonderyl:
#### What type of PR is this?

/kind bug

VMSS requires extended location when update the vmss.

This is error 
```
The resource 'aks-xxxxxxxx-vmss' already exists in extended location 'EdgeZone': 'xxxxxx' of location 'eastus'. A resource with the same name cannot be created in location 'eastus'. Please select a new resource name."
```
The VMSS team showed me the code, basically, when extendedLocaiton is set when the vmss is created, it will expect extendedLocaiton in subsequence update requests. The code was there since 2020, long before the first PR of extendedLocaiton in Cloud provider. My guess is the feature was half baked and never verified.

Fixes #

```release-note
Bug fix: sends extended location when update vmss.
```

### DIFF
--- a/pkg/provider/azure_vmss.go
+++ b/pkg/provider/azure_vmss.go
@@ -1347,7 +1347,7 @@ func (ss *ScaleSet) ensureVMSSInPool(ctx context.Context, _ *v1.Service, nodes [
 			})
 		primaryIPConfig.Properties.LoadBalancerBackendAddressPools = loadBalancerBackendAddressPools
 		newVMSS := armcompute.VirtualMachineScaleSet{
-			Location: vmss.Location,
+			Location:         vmss.Location,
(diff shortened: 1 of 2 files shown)


Liunardy (approved):

andyzhangx:
@wonderyl do you have the detailed error msg when hitting this issue? if yes, pls update in the PR description, I think that will be good for TSG.

And we are using this update vmss in long time, we don't see any issue without extendedlocation, so it seems this field is required in recently Azure platform API update?

wonderyl:
> @wonderyl do you have the detailed error msg when hitting this issue? if yes, pls update in the PR description, I think that will be good for TSG.
> 
> And we are using this update vmss in long time, we don't see any issue without extendedlocation, so it seems this field is required in recently Azure platform API update?

done

PR #10829, opened 2026-08-13
fix: register e2e Service cleanup defer before it can be skipped

### LINKED ISSUE #10126: tests/e2e: ensureloadbalancer can skip service cleanup when post-create Get fails
#### What happened:

Several e2e cases in `tests/e2e/network/ensureloadbalancer.go` create and expose a `Service`, then call `Get(...)`, and only register the cleanup `defer` after that `Get` returns. If that `Get` call fails, or if `Expect(err).NotTo(HaveOccurred())` fails immediately after it, the cleanup `defer` is never registered, so the test can leave the `Service` behind.

Confirmed occurrences:
- `It("should respect service with various configurations")` at `tests/e2e/network/ensureloadbalancer.go:986-992`
- `It("should respect service with BYO public IP with various configurations")` at `tests/e2e/network/ensureloadbalancer.go:1050-1056`
- `It("should respect service with BYO public IP prefix with various configurations")` at `tests/e2e/network/ensureloadbalancer.go:1114-1120`
- `It("should respect internal service with various configurations")` at `tests/e2e/network/ensureloadbalancer.go:1148-1154`

Cleanup should be registered as soon as the `Service` is known to exist, before later `Get(...)` calls or assertions that can abort the spec.

1. Run one of the listed tests.
2. Force the `Get(...)` call at one of the listed locations to return an error, or inject a failing assertion before the cleanup `defer` line.
3. Observe that the `DeleteService(...)` cleanup block is never executed because the `defer` was not registered yet.

I searched `tests/e2e` for the same `Get(...); defer cleanup; Expect(err)` ordering and found the four confirmed cases above.

I also checked nearby `WaitServiceExposureAndValidateConnectivity(...)` cases. Those are less urgent because that helper already deletes the `Service` on its own error path, so this issue is scoped to the confirmed post-create `Get(...)` gaps.

A straightforward fix would be to move the cleanup registration earlier, immediately after `Create(...)` succeeds or immediately after any helper that guarantees the `Service` exists.

#### Environment:

- Cloud provider or hardware configuration: N/A
- Install tools: N/A
- Others: N/A

pujitha24:
#### What type of PR is this?

/kind bug

In `tests/e2e/network/ensureloadbalancer.go`, four specs create/expose a
Service and only register the cleanup `defer DeleteService(...)` after a
later `Get(...)` call, with one or more `Expect(...)` assertions running
between successful service creation and that defer registration. In
Ginkgo v2 + Gomega, a failing `Expect(...)` unwinds the current goroutine
via `runtime.Goexit()`; any defer not yet registered at that point is
skipped. If the Service already exists by then and one of those
intervening assertions fails, the cleanup never runs and the Service is
left behind in the cluster.

This PR moves each `defer` to be registered immediately after the point
where the Service is first known to exist (right after `Create(...)`
succeeds, or right after the `createAndExposeDefaultServiceWithAnnotation(...)`
helper returns), before any subsequent `Expect(...)` calls. This mirrors
the fix suggested in the issue report. `utils.DeleteService` already
treats a NotFound error as success, so registering the deferred delete
this early is safe even on the path where creation itself failed.

Fixes #10126

- This is a test-only change; `tests/e2e/network/ensureloadbalancer.go` is
  the only file touched (4 hunks, 14 insertions / 14 deletions).
- Validation performed: `go build ./e2e/network/...`, `go vet ./e2e/network/...`,
  `gofmt -l`, and `golangci-lint run` against the `tests` module — all pass,
  and golangci-lint reports the same 33 pre-existing issues before and after
  this change, confirming no new lint issues were introduced.
- These specs are Ginkgo e2e tests that require a live Azure/AKS cluster to
  execute; that isn't available in this environment, so the fix was not
  exercised against a live cluster. The correctness argument rests on Go's
  documented defer/Goexit semantics and static review of the reordered
  control flow, not a live reproduction.
- CI on this repo requires a maintainer to add `ok-to-test` — happy to
  address anything it surfaces.

```release-note
NONE
```

```docs

```

### DIFF
--- a/tests/e2e/network/ensureloadbalancer.go
+++ b/tests/e2e/network/ensureloadbalancer.go
@@ -1012,14 +1012,14 @@ var _ = Describe("EnsureLoadBalancer should not update any resources when servic
 		}
 
 		ips := createAndExposeDefaultServiceWithAnnotation(cs, tc.IPFamily, testServiceName, ns.Name, labels, annotation, ports)
-		Expect(len(ips)).NotTo(BeZero())
-		ip := ips[0]
-		service, err := cs.CoreV1().Services(ns.Name).Get(context.TODO(), testServiceName, metav1.GetOptions{})
 		defer func() {
 			By("Cleaning up")
 			err := utils.DeleteService(cs, ns.Name, testServiceName)
 			Expect(err).NotTo(HaveOccurred())
 		}()
+		Expect(len(ips)).NotTo(BeZero())
+		ip := ips[0]
+		service, err := cs.CoreV1().Services(ns.Name).Get(context.TODO(), testServiceName, metav1.GetOptions{})
 		Expect(err).NotTo(HaveOccurred())
 
 		By("Update the service and without significant changes and compare etags")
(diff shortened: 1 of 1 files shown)


pujitha24:
/assign @bridgetkromhout

Small test-only fix, ready for review whenever you have a moment.

PR #10838, opened 2026-08-17
fix: one-directional NFS EiT account matching for account reuse

rnhan:
#### What type of PR is this?

/kind bug

Fixes # 

The isNFSEncryptionInTransitEnabledEqual logic would lead to not matching existing EiT accounts (without the flag having been set), so new EiT PVCs would always create a new account after this change.  This adds a one-directional match (via the isNFSEncryptionInTransitEnabledLogic) to skip the check for EiT requests (match any existing account) and only check for non-EiT requests s.t. non EiT requests don't get matched to accounts with Required=true.

https://github.com/kubernetes-sigs/azurefile-csi-driver/pull/3335#pullrequestreview-4933645188 

```release-note
none
```

### DIFF
--- a/pkg/provider/storage/azure_storageaccount.go
+++ b/pkg/provider/storage/azure_storageaccount.go
@@ -1162,6 +1162,15 @@ func (az *AccountRepo) isNFSEncryptionInTransitEnabledEqual(ctx context.Context,
 		return true, nil
 	}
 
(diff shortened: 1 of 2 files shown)


andyzhangx (approved):
/ok-to-test
/kind bug

andyzhangx (approved):
/lgtm

PR #10863, opened 2026-08-18
fix: skip public ip check for azure stack

patrickdillon:
#### What type of PR is this?

/kind bug
/kind regression

Azure Stack does not have a concept of public or private IP addresses, just IP addresses. Skip the function to find matched public IP addresses when running on azure stack. After #10507, provisioning LBs on Azure Stack fails with: 

```
(SyncLoadBalancerFailed: The service-controller component is reporting SyncLoadBalancerFailed events like: Error syncing load balancer: failed to ensure load balancer: external Service "openshift-ingress/router-default" cannot use non-public LoadBalancer IP "10.255.96.63"; use an internal LoadBalancer annotation or a valid Azure Public IP address
```

Fixes #

```release-note
NONE
```

```docs

```


sadasu:
LGTM

andyzhangx (approved):
/ok-to-test
/lgtm

andyzhangx (approved):
/lgtm

PR #10864, opened 2026-08-21
fix: emit access control warning events on ensure and never block deletion

Liunardy:
#### What type of PR is this?
/kind bug
/kind feature

- **Warning events were not emitted on create/update.** `reconcileSecurityGroup` wired the event emitter only on the delete path, so a rejected configuration produced no Kubernetes event on apply. Introduced by #7178, which swapped `SkipAnnotationValidation()` for `WithEventEmitter()` inside the existing `if !wantLb` branch, so events were only emitted when deleting. `InvalidAllowedIPRanges` and `ConflictConfiguration` are now visible to users.
- **Deletion could be blocked by configuration validation.** Setting both `spec.loadBalancerSourceRanges` and `azure-allowed-ip-ranges` returned `ErrSetBoth` on the delete path, so `EnsureLoadBalancerDeleted` failed and the cleanup finalizer was never
  removed. Follows the intended `When deleting LB, we don't need to validate the annotation`.
- **New `IPRangeFamilyMismatch` warning.** No traffic from an IP range whose IP family the Service does not have can reach the Service, because kube-proxy only programs rules for the Service's own families. The wrong-family proxier [drops the Service outright](https://github.com/kubernetes/kubernetes/blob/96cb9ab4201d88ce5e549fde047a686171838fdb/pkg/proxy/servicechangetracker.go#L145-L148), and the right-family one [filters `status.loadBalancer.ingress`](https://github.com/kubernetes/kubernetes/blob/96cb9ab4201d88ce5e549fde047a686171838fdb/pkg/proxy/serviceport.go#L216-L229). A security rule may still be created for such a range. This was silent; it now emits a warning event naming the ranges.
- **Warning events could describe a security group change that never happened.** `InvalidSourceRanges` and `InvalidAllowedIPRanges` were emitted during parsing, before the `ErrSetBoth` check. With both `spec.loadBalancerSourceRanges` and a partly-valid `azure-allowed-ip-ranges` set, the user saw "ignoring and adding a default DenyAll rule in security group" followed by `SyncLoadBalancerFailed`, and no rule was added at all. All fatal checks now run before any event is emitted.

Adds unit, integration and E2E coverage.

Related #6492. This unblocks deletion for a Service with both `spec.loadBalancerSourceRanges` and the `service.beta.kubernetes.io/load-balancer-source-ranges` annotation. The hard error on create/update is unchanged.

```release-note
fix: `InvalidSourceRanges`, `InvalidAllowedIPRanges` and `ConflictConfiguration` warning events are now emitted when a Service is created or updated instead of only during delete, and are no longer emitted when the configuration is rejected and no security group change is made. Deleting a Service is no longer blocked when both `spec.loadBalancerSourceRanges` and the `service.beta.kubernetes.io/azure-allowed-ip-ranges` annotation are set.
feat: a new `IPRangeFamilyMismatch` warning event reports IP ranges from `spec.loadBalancerSourceRanges`, `service.beta.kubernetes.io/load-balancer-source-ranges`, and `service.beta.kubernetes.io/azure-allowed-ip-ranges` whose IP family the Service does not have.
```

```docs

```

### DIFF
--- a/pkg/provider/azure_loadbalancer.go
+++ b/pkg/provider/azure_loadbalancer.go
@@ -3429,6 +3429,8 @@ func (az *Cloud) reconcileSecurityGroup(
 		var opts []loadbalancer.AccessControlOption
 		if !wantLb {
 			// When deleting LB, we don't need to validate the annotation
+			opts = append(opts, loadbalancer.SkipAnnotationValidation())
+		} else {
 			opts = append(opts, loadbalancer.WithEventEmitter(az.Event))
 		}
 		accessControl, err = loadbalancer.NewAccessControl(logger, service, sg, opts...)
--- a/pkg/provider/loadbalancer/accesscontrol.go
+++ b/pkg/provider/loadbalancer/accesscontrol.go
@@ -52,7 +52,8 @@ type AccessControl struct {
 }
 
 type accessControlOptions struct {
-	EventEmitter K8sEventEmitter
+	EventEmitter             K8sEventEmitter
+	SkipAnnotationValidation bool
 }
 
 var defaultAccessControlOptions = accessControlOptions{
@@ -64,6 +65,13 @@ type AccessControlOption func(*accessControlOptions)
 func WithEventEmitter(emitter K8sEventEmitter) AccessControlOption {
 	return func(o *accessControlOptions) {
 		o.EventEmitter = emitter
+	}
+}
+
+// SkipAnnotationValidation keeps a rejected annotation configuration from aborting the caller.
+func SkipAnnotationValidation() AccessControlOption {
+	return func(o *accessControlOptions) {
+		o.SkipAnnotationValidation = true
 	}
 }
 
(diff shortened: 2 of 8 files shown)


pkg/provider/loadbalancer/accesscontrol.go
```diff
@@ -124,7 +132,46 @@ func NewAccessControl(logger logr.Logger, svc *v1.Service, sg *armnetwork.Securi
 		AllowedServiceTags:                     allowedServiceTags,
 		invalidRanges:                          append(invalidSourceRanges, invalidAllowedIPRanges...),
 		securityRuleDestinationPortsByProtocol: securityRuleDestinationPortsByProtocol,
-	}, nil
+	}
+
+	if !options.SkipAnnotationValidation {
+		if ignored := ac.ignoredIPFamilyRanges(); len(ignored) > 0 {
+			ranges := fnutil.Map(func(p netip.Prefix) string { return p.String() }, ignored)
+			logger.V(2).Info("Ignored allowed IP ranges that do not match the Service's IP families",
```
nilo19: The mismatched ranges will not always be ignored. At least these two scenarios:
1. dualstack clusters, single stack service, turning off floating IP. In this case the destination will be derived from both ipv4 and ipv6 backend pools.
2. additional-public-ips annotation, we don't check if the ip family matches for this annotation.

pkg/provider/azure_loadbalancer.go
```diff
@@ -3429,6 +3429,8 @@ func (az *Cloud) reconcileSecurityGroup(
 		var opts []loadbalancer.AccessControlOption
 		if !wantLb {
 			// When deleting LB, we don't need to validate the annotation
+			opts = append(opts, loadbalancer.SkipAnnotationValidation())
+		} else {
 			opts = append(opts, loadbalancer.WithEventEmitter(az.Event))
```
nilo19, line 3434: As per my testing, when setting an invalid source range, together with spec.LoadBalancerSourceRanges, the event message is different from the genuine error.
err = cannot set both spec.LoadBalancerSourceRanges and service annotation
      service.beta.kubernetes.io/azure-allowed-ip-ranges
EVENT InvalidAllowedIPRanges: Found invalid service.beta.kubernetes.io/azure-allowed-ip-ranges
      ["bad"], ignoring and adding a default DenyAll rule in security group.

pkg/provider/loadbalancer/accesscontrol.go
Liunardy: Updated the event and log message to mention traffic cannot reach the service if the source ranges do not match the service's IP family.

pkg/provider/azure_loadbalancer.go
Liunardy: Reordered the validation so error comes first.

PR #10866, opened 2026-08-23
Build health-probe-proxy binary inside its image

### LINKED ISSUE #10632: health-probe-proxy: add multi-stage Dockerfile with pinned Go builder
health-probe-proxy's Dockerfile has no builder stage. It copies a pre-built binary from the host. The Go version compiling the binary depends on the build environment. Fixing stdlib CVEs requires bumping health-probe-proxy's go.mod directive rather than a Dockerfile change.

What to do:
1. Add a multi-stage build to Dockerfile using mcr.microsoft.com/oss/go/microsoft/golang (same as cloud-controller-manager/cloud-node-manager Dockerfiles).
2. Align the Makefile image build with how cloud-controller-manager/cloud-node-manager builds theirs (compilation inside the container, not on-host).
3. Add health-probe-proxy docker Dependabot entries for community-supported release branches so builder image digests are auto-bumped.

Why:
- Stdlib CVE fixes become a Dockerfile-only change (builder digest bump.
- Consistent with cloud-controller-manager/cloud-node-manager.
- Dependabot already auto-bumps the builder for cloud-controller-manager/cloud-node-manager. health-probe-proxy should get the same coverage.

yindia:
#### What type of PR is this?

/kind cleanup

The `health-probe-proxy` image copies a binary that is compiled on the host, so the Go toolchain baked into the binary depends on whoever built it. That makes stdlib CVE fixes awkward: there is no builder image to bump, you have to chase the version on the build machine.

This adds a proper builder stage to `health-probe-proxy/Dockerfile`, using the same pinned `mcr.microsoft.com/oss/go/microsoft/golang` image as the cloud-controller-manager and cloud-node-manager Dockerfiles. The binary is now compiled inside the container, so a stdlib CVE fix is just a builder digest bump.

A few related changes come along with it:

* The Makefile no longer builds the binary on the host before the image build, since the Dockerfile handles that now.
* Dependabot gets `health-probe-proxy` docker entries on the community-supported release branches, so the builder digest is bumped automatically the same way it already is for CCM and CNM.

Fixes #10632

The Windows image is left as is. It cross-compiles the `.exe` on the host and there is no Go builder stage for nanoserver, so it is out of scope here.

Verified the image builds and the binary runs from the builder stage.

```release-note
NONE
```

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -384,6 +384,30 @@ updates:
         - "minor"
         - "major"
   - package-ecosystem: "docker"
+    directory: "/health-probe-proxy"
+    schedule:
+      interval: "weekly"
+      time: "01:00"
+      timezone: "Asia/Shanghai"
+    target-branch: "release-1.36"
+    labels:
+      - "area/dependency"
+      - "release-note-none"
+      - "ok-to-test"
+      - "kind/cleanup"
+      - "lgtm"
+      - "approved"
+      - "release-1.36"
+    groups:
+      all:
+        applies-to: version-updates
+        patterns:
+        - "*"
+        update-types:
+        - "patch"
+        - "minor"
+        - "major"
+  - package-ecosystem: "docker"
     directory: "/"
     schedule:
       interval: "weekly"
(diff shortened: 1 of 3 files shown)


yindia:
cc: @Liunardy @elmiko @nilo19

Liunardy:
Please fix the build failure https://github.com/kubernetes-sigs/cloud-provider-azure/actions/runs/34417958827/job/104633033960?pr=10866. 

Curious how this was done:
> Verified the image builds and the binary runs

PR #10867, opened 2026-08-23
Fix retryrepectthrottled typo in azclient policy package

### LINKED ISSUE #10558: Fix retryrepectthrottled typo in azclient
https://github.com/kubernetes-sigs/cloud-provider-azure/blob/3f57aa450c856b9c5dbf0e2b5863fc220281602c/pkg/azclient/policy/retryrepectthrottled/throttle.go#L17
Should be retryrespectthrottled.

yindia:
#### What type of PR is this?

/kind cleanup

The throttling policy package under `pkg/azclient/policy` was named `retryrepectthrottled`, which misspells "respect". This renames the package and its directory to `retryrespectthrottled` and updates the two importers (`metrics` and `utils`) to match.

It is a pure rename with no behavior change. The policy, its exported types, and its logic stay exactly the same.

Fixes #10558

Only the azclient module is touched here. The root module still vendors azclient at its published tag, so the vendored copy keeps the old name until the next azclient version bump picks this up.

Ran `go build ./...`, `go vet`, and the package tests in `pkg/azclient` after the rename and they all pass.

```release-note
NONE
```

### DIFF
--- a/pkg/azclient/metrics/metrics.go
+++ b/pkg/azclient/metrics/metrics.go
@@ -29,7 +29,7 @@ import (
 	"go.opentelemetry.io/otel/metric/noop"
 
 	"sigs.k8s.io/cloud-provider-azure/pkg/azclient/policy/ratelimit"
-	"sigs.k8s.io/cloud-provider-azure/pkg/azclient/policy/retryrepectthrottled"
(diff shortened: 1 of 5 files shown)


yindia:
cc @feiskyer @nilo19 @Liunardy

PR #10897, opened 2026-08-26
fix: ignore controller-owned tag keys in pip tags annotation

Liunardy:
#### What type of PR is this?

/kind bug

The `service.beta.kubernetes.io/azure-pip-tags` annotation was applied to the Public IP without filtering, so it could set the tag keys the controller owns (`k8s-azure-cluster-name`, `k8s-azure-service`, `k8s-azure-dns-label-service`, and their legacy spellings).

`ensurePIPTagged` now skips those keys when merging the annotation and emits a `Warning` event with reason `IgnoredPIPTagKeys` on the Service. Matching is case-insensitive because [Azure resource tag names are case-insensitive](https://learn.microsoft.com/azure/azure-resource-manager/management/tag-resources).
All other annotation keys behave as before.

```release-note
The `service.beta.kubernetes.io/azure-pip-tags` annotation no longer sets the Public IP tag keys managed by the controller (`k8s-azure-cluster-name`, `k8s-azure-service`, `k8s-azure-dns-label-service`, and their legacy names). Such keys in the annotation are now ignored, in any letter case, and a `Warning` event with reason `IgnoredPIPTagKeys` is emitted on the Service.
```

```docs

```

### DIFF
--- a/pkg/provider/azure_loadbalancer.go
+++ b/pkg/provider/azure_loadbalancer.go
@@ -1529,6 +1529,25 @@ func getClusterFromPIPClusterTags(tags map[string]*string) string {
 	return ""
 }
 
+// Public IP tag keys whose values the controller owns.
+var reservedPIPTagKeys = []string{
+	consts.ClusterNameKey, consts.LegacyClusterNameKey,
+	consts.ServiceTagKey, consts.LegacyServiceTagKey,
+	consts.ServiceUsingDNSKey, consts.LegacyServiceUsingDNSKey,
+}
+
+// isPIPReservedTagKey reports whether key names a Public IP tag whose value the controller owns.
+// Matching ignores case: Azure resource tag names are case-insensitive.
+// https://learn.microsoft.com/azure/azure-resource-manager/management/tag-resources
+func isPIPReservedTagKey(key string) bool {
+	for _, reserved := range reservedPIPTagKeys {
+		if strings.EqualFold(key, reserved) {
+			return true
+		}
+	}
+	return false
+}
+
 type serviceIPTagRequest struct {
 	IPTagsRequestedByAnnotation bool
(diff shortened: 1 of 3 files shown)


pkg/provider/azure_loadbalancer.go
```diff
@@ -3614,6 +3626,9 @@ func (az *Cloud) ensurePIPTagged(service *v1.Service, pip *armnetwork.PublicIPAd
 	}
 
 	for k, v := range annotationTags {
+		if isPIPReservedTagKey(k) {
```
nilo19, line 3637: Worth adding a new e2e or enriching an existing one?

Liunardy: Added

tests/e2e/network/service_annotations.go
```diff
@@ -487,6 +487,109 @@ var _ = Describe("Service with annotation", Label(utils.TestSuiteLabelServiceAnn
 		testPIPTagAnnotationWithTags(cs, tc, ns, serviceName, labels, ports, expectedTags)
 	})
 
+	DescribeTable("should ignore controller-owned tag keys in service annotation `service.beta.kubernetes.io/azure-pip-tags`",
```
nilo19, line 490: Do we have any existing DNS label e2e cases? If so can we cover it from there? I think this could make the PR shorter.

Liunardy: Yes, there is an existing DNS label E2E case (`should support service annotation 'service.beta.kubernetes.io/azure-dns-label-name'`). But that case performs DNS-specific checks and create/delete/update some services with managed and BYO PIPs. I think covering this controller-owned tag checks there would make it checks too many things.

This covers exactly the case where controller-owned tags are ignored on new and existing managed PIPs.

PR #10906, opened 2026-08-28
Fix case-insensitive duplicate tags

### LINKED ISSUE #10902: Tag keys should be case insensitive
#### What happened:
Azure resource tag names are [case-insensitive](https://learn.microsoft.com/azure/azure-resource-manager/management/tag-resources). 

https://github.com/kubernetes-sigs/cloud-provider-azure/blob/6c17ad5b5128f986ad00f566aad4a80c43a75eb3/pkg/provider/azure_utils.go#L65 keys its result map by exact string, so two tag keys differing only in case become two separate entries, even though both refer to the same tag:
```text
"foo=1,FOO=2"
```
The duplicate entries are later resolved case-insensitively while ranging over a Go map, so which value ends up on the resource depends on map iteration order and can differ between reconciles, so the controller may issue an update even when nothing should change.

`azure-pip-tags`, cloud config `tags`, and `tagsMap` go through the same function and reach load balancers, route tables and private link services. `tagsMap` is a Go map, so ordering isn't defined there and there is no "first occurrence".

Only 1 entry should be created if the tag keys match case-insensitively.
Maybe warn when there are duplicate keys and only the first one is used. An event on the Service for `azure-pip-tags`, and a log for the cloud config paths, which have no object to attach an event to.

#### Environment:

- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

kavix:
#### What type of PR is this?

/kind bug

Azure resource tag names are case-insensitive. The duplicate entries were resolved case-insensitively while ranging over a Go map, so which value ended up on the resource depended on map iteration order and could differ between reconciles. This caused the controller to issue an update even when nothing should change.
This PR fixes the issue by keeping only the first case-insensitive tag occurrence and discarding duplicates. It also logs a warning or issues an event when duplicates are found.

Fixes #10902

```release-note
Fix Azure resource tags duplication due to case-insensitivity in cloud config `tags`/`tagsMap` and the `service.beta.kubernetes.io/azure-pip-tags` service annotation by preserving the first occurrence and ignoring duplicates.
```

```docs
```

### DIFF
--- a/pkg/provider/azure_loadbalancer.go
+++ b/pkg/provider/azure_loadbalancer.go
@@ -3628,7 +3628,7 @@ func shouldReleaseExistingOwnedPublicIP(
 func (az *Cloud) ensurePIPTagged(service *v1.Service, pip *armnetwork.PublicIPAddress) bool {
 	configTags, err := parseTags(az.Tags, az.TagsMap)
 	if err != nil {
-		klog.Warningf("ensurePIPTagged: %v", err)
+		log.Background().WithName("ensurePIPTagged").Error(err, "Found duplicate tag keys in cloud configuration")
 	}
 	annotationTags := make(map[string]*string)
 	if _, ok := service.Annotations[consts.ServiceAnnotationAzurePIPTags]; ok {
@@ -3636,7 +3636,7 @@ func (az *Cloud) ensurePIPTagged(service *v1.Service, pip *armnetwork.PublicIPAd
 		annotationTags, annErr = parseTags(service.Annotations[consts.ServiceAnnotationAzurePIPTags], map[string]string{})
 		if annErr != nil {
 			az.Event(service, v1.EventTypeWarning, "DuplicatePIPTags", annErr.Error())
-			klog.Warningf("ensurePIPTagged: %v", annErr)
+			log.Background().WithName("ensurePIPTagged").Error(annErr, "Found duplicate tag keys in service annotation", "service", getServiceName(service))
 		}
 	}
 
@@ -4385,7 +4385,7 @@ func (az *Cloud) ensureLoadBalancerTagged(lb *armnetwork.LoadBalancer) bool {
 	}
 	tags, err := parseTags(az.Tags, az.TagsMap)
 	if err != nil {
-		klog.Warningf("ensureLoadBalancerTagged: %v", err)
+		log.Background().WithName("ensureLoadBalancerTagged").Error(err, "Found duplicate tag keys in cloud configuration")
 	}
 	if lb.Tags == nil {
 		lb.Tags = make(map[string]*string)
@@ -4404,7 +4404,7 @@ func (az *Cloud) ensureSecurityGroupTagged(sg *armnetwork.SecurityGroup) bool {
 	}
 	tags, err := parseTags(az.Tags, az.TagsMap)
 	if err != nil {
-		klog.Warningf("ensureSecurityGroupTagged: %v", err)
+		log.Background().WithName("ensureSecurityGroupTagged").Error(err, "Found duplicate tag keys in cloud configuration")
 	}
 	if sg.Tags == nil {
 		sg.Tags = make(map[string]*string)
--- a/pkg/provider/azure_loadbalancer.go
+++ b/pkg/provider/azure_loadbalancer.go
@@ -3626,10 +3626,18 @@ func shouldReleaseExistingOwnedPublicIP(
 
 // ensurePIPTagged ensures the public IP of the service is tagged as configured
 func (az *Cloud) ensurePIPTagged(service *v1.Service, pip *armnetwork.PublicIPAddress) bool {
-	configTags := parseTags(az.Tags, az.TagsMap)
+	configTags, err := parseTags(az.Tags, az.TagsMap)
+	if err != nil {
+		klog.Warningf("ensurePIPTagged: %v", err)
+	}
 	annotationTags := make(map[string]*string)
 	if _, ok := service.Annotations[consts.ServiceAnnotationAzurePIPTags]; ok {
-		annotationTags = parseTags(service.Annotations[consts.ServiceAnnotationAzurePIPTags], map[string]string{})
+		var annErr error
+		annotationTags, annErr = parseTags(service.Annotations[consts.ServiceAnnotationAzurePIPTags], map[string]string{})
+		if annErr != nil {
+			az.Event(service, v1.EventTypeWarning, "DuplicatePIPTags", annErr.Error())
+			klog.Warningf("ensurePIPTagged: %v", annErr)
+		}
 	}
 
 	var ignoredReservedKeys bool
@@ -4375,7 +4383,10 @@ func (az *Cloud) ensureLoadBalancerTagged(lb *armnetwork.LoadBalancer) bool {
 	if az.Tags == "" && len(az.TagsMap) == 0 {
 		return false
 	}
-	tags := parseTags(az.Tags, az.TagsMap)
+	tags, err := parseTags(az.Tags, az.TagsMap)
+	if err != nil {
+		klog.Warningf("ensureLoadBalancerTagged: %v", err)
+	}
 	if lb.Tags == nil {
 		lb.Tags = make(map[string]*string)
 	}
@@ -4391,7 +4402,10 @@ func (az *Cloud) ensureSecurityGroupTagged(sg *armnetwork.SecurityGroup) bool {
 	if az.Tags == "" && (len(az.TagsMap) == 0) {
 		return false
 	}
-	tags := parseTags(az.Tags, az.TagsMap)
+	tags, err := parseTags(az.Tags, az.TagsMap)
+	if err != nil {
+		klog.Warningf("ensureSecurityGroupTagged: %v", err)
+	}
 	if sg.Tags == nil {
 		sg.Tags = make(map[string]*string)
 	}
--- a/pkg/provider/azure_loadbalancer.go
+++ b/pkg/provider/azure_loadbalancer.go
@@ -3626,17 +3626,18 @@ func shouldReleaseExistingOwnedPublicIP(
 
 // ensurePIPTagged ensures the public IP of the service is tagged as configured
 func (az *Cloud) ensurePIPTagged(service *v1.Service, pip *armnetwork.PublicIPAddress) bool {
-	configTags, err := parseTags(az.Tags, az.TagsMap)
-	if err != nil {
-		log.Background().WithName("ensurePIPTagged").Error(err, "Found duplicate tag keys in cloud configuration")
+	configTags, droppedConfigKeys := parseTags(az.Tags, az.TagsMap)
+	if len(droppedConfigKeys) > 0 {
+		log.Background().WithName("ensurePIPTagged").V(2).Info("Found duplicate tag keys in cloud configuration", "duplicateKeys", droppedConfigKeys)
 	}
 	annotationTags := make(map[string]*string)
 	if _, ok := service.Annotations[consts.ServiceAnnotationAzurePIPTags]; ok {
-		var annErr error
-		annotationTags, annErr = parseTags(service.Annotations[consts.ServiceAnnotationAzurePIPTags], map[string]string{})
-		if annErr != nil {
-			az.Event(service, v1.EventTypeWarning, "DuplicatePIPTags", annErr.Error())
-			log.Background().WithName("ensurePIPTagged").Error(annErr, "Found duplicate tag keys in service annotation", "service", getServiceName(service))
+		var droppedAnnotationKeys []string
+		annotationTags, droppedAnnotationKeys = parseTags(service.Annotations[consts.ServiceAnnotationAzurePIPTags], nil)
+		if len(droppedAnnotationKeys) > 0 {
+			az.Event(service, v1.EventTypeWarning, "DuplicatePIPTags", fmt.Sprintf(
+				"Ignoring duplicate tag keys (case-insensitive) in the %s annotation; keeping the first occurrence: %s",
+				consts.ServiceAnnotationAzurePIPTags, strings.Join(droppedAnnotationKeys, ", ")))
 		}
 	}
 
@@ -4383,9 +4384,9 @@ func (az *Cloud) ensureLoadBalancerTagged(lb *armnetwork.LoadBalancer) bool {
 	if az.Tags == "" && len(az.TagsMap) == 0 {
 		return false
 	}
-	tags, err := parseTags(az.Tags, az.TagsMap)
-	if err != nil {
-		log.Background().WithName("ensureLoadBalancerTagged").Error(err, "Found duplicate tag keys in cloud configuration")
+	tags, droppedKeys := parseTags(az.Tags, az.TagsMap)
+	if len(droppedKeys) > 0 {
+		log.Background().WithName("ensureLoadBalancerTagged").V(2).Info("Found duplicate tag keys in cloud configuration", "duplicateKeys", droppedKeys)
 	}
 	if lb.Tags == nil {
 		lb.Tags = make(map[string]*string)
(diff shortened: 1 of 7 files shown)


kavix:
Can u re review this @andyzhangx

pkg/provider/azure_utils_test.go
```diff
@@ -280,10 +281,58 @@ func TestParseTags(t *testing.T) {
...
+			expectedTags: map[string]*string{
+				"FOO": ptr.To("baz"),
+			},
+			expectedErr: true,
+		},
+		{
+			description: "parseTags should handle Unicode case-folded duplicate keys in tagsMap",
+			tagsMap: map[string]string{
+				"S": "first",
+				"ſ": "second",
```
Liunardy, line 322: S != ſ

Image

pkg/provider/azure_loadbalancer.go
```diff
@@ -3626,10 +3626,18 @@ func shouldReleaseExistingOwnedPublicIP(
...
+		log.Background().WithName("ensurePIPTagged").Error(err, "Found duplicate tag keys in cloud configuration")
+	}
 	annotationTags := make(map[string]*string)
 	if _, ok := service.Annotations[consts.ServiceAnnotationAzurePIPTags]; ok {
-		annotationTags = parseTags(service.Annotations[consts.ServiceAnnotationAzurePIPTags], map[string]string{})
+		var annErr error
+		annotationTags, annErr = parseTags(service.Annotations[consts.ServiceAnnotationAzurePIPTags], map[string]string{})
+		if annErr != nil {
+			az.Event(service, v1.EventTypeWarning, "DuplicatePIPTags", annErr.Error())
+			log.Background().WithName("ensurePIPTagged").Error(annErr, "Found duplicate tag keys in service annotation", "service", getServiceName(service))
```
Liunardy: Probably don't need to log since event is already emitted.

pkg/provider/azure_loadbalancer.go
```diff
@@ -3626,10 +3626,18 @@ func shouldReleaseExistingOwnedPublicIP(
 
 // ensurePIPTagged ensures the public IP of the service is tagged as configured
 func (az *Cloud) ensurePIPTagged(service *v1.Service, pip *armnetwork.PublicIPAddress) bool {
-	configTags := parseTags(az.Tags, az.TagsMap)
+	configTags, err := parseTags(az.Tags, az.TagsMap)
+	if err != nil {
+		log.Background().WithName("ensurePIPTagged").Error(err, "Found duplicate tag keys in cloud configuration")
```
Liunardy: Don't log as error. Log as info V(2) since it is a warning.

pkg/provider/azure_utils.go
```diff
@@ -85,13 +89,25 @@ func parseTags(tags string, tagsMap map[string]string) map[string]*string {
 				klog.Warning("parseTags: empty key, ignoring this key-value pair")
 				continue
 			}
+			if found, kExisting := findKeyInMapCaseInsensitive(formatted, k); found {
+				errs = append(errs, fmt.Errorf("found duplicate tag key %q (case-insensitive) in tags, ignoring this key-value pair and keeping the first occurrence %q", k, kExisting))
```
Liunardy: With many duplicate keys this builds one very long message mentioning the same thing. Best to collect the ignored or kept keys first then build the message.

pkg/provider/azure_utils.go
```diff
@@ -100,15 +116,40 @@ func parseTags(tags string, tagsMap map[string]string) map[string]*string {
...
 
-			if found, k := findKeyInMapCaseInsensitive(formatted, key); found && k != key {
-				logger.V(4).Info("found identical keys from tags and tagsMap (case-insensitive), will replace", "identical keys", k, "keyFromTagsMap", key)
-				delete(formatted, k)
+			var alreadyInTagsMap bool
+			var existingTagsMapKey string
+			for _, k := range tagsMapKeys {
+				if strings.EqualFold(k, key) {
+					alreadyInTagsMap = true
+					existingTagsMapKey = k
```
Liunardy: Can use a map that normalize the key instead of slice to check for duplicates.

Liunardy:
Can we extend existing E2E case "should ignore controller-owned tag keys in service annotation `service.beta.kubernetes.io/azure-pip-tags`" to check the event is emitted?
I think we can add ",A=should-be-dropped" to
https://github.com/kubernetes-sigs/cloud-provider-azure/blob/ec98c027e9af53f6651d9035baee67377d89061d/tests/e2e/network/service_annotations.go#L505
And then check the duplicate tag warning event is emitted.

pkg/provider/azure_loadbalancer.go
```diff
@@ -3626,10 +3626,18 @@ func shouldReleaseExistingOwnedPublicIP(
...
+	}
 	annotationTags := make(map[string]*string)
 	if _, ok := service.Annotations[consts.ServiceAnnotationAzurePIPTags]; ok {
-		annotationTags = parseTags(service.Annotations[consts.ServiceAnnotationAzurePIPTags], map[string]string{})
+		var annErr error
+		annotationTags, annErr = parseTags(service.Annotations[consts.ServiceAnnotationAzurePIPTags], map[string]string{})
+		if annErr != nil {
+			az.Event(service, v1.EventTypeWarning, "DuplicatePIPTags", annErr.Error())
```
Liunardy: The DuplicatePIPTags event message is the raw parseTags error ("...in tags..."), which doesn't name the annotation the user actually set. Recommend making it self-contained and annotation-specific. 

Maybe have parseTags return the dropped keys and format the message here.

pkg/provider/azure_utils_test.go
kavix: Removed this test case.

pkg/provider/azure_loadbalancer.go
kavix: Done. Removed the error log since the DuplicatePIPTags warning event is already emitted on the Service.

pkg/provider/azure_loadbalancer.go
kavix: Updated to log at .V(2).Info(...) across all cloud configuration tag reconciliation call sites.

pkg/provider/azure_utils.go
kavix: Refactored parseTags to collect and return the slice of dropped duplicate keys ([]string), allowing callers to construct concise, consolidated messages.

pkg/provider/azure_utils.go
kavix: Updated to use a normalized key map (map[string]string) for O(1) duplicate checking instead of iterating over a slice.

pkg/provider/azure_loadbalancer.go
kavix: Updated the event message to be self-contained and reference ServiceAnnotationAzurePIPTags directly alongside the list of dropped keys: "Ignoring duplicate tag keys (case-insensitive) in the service.beta.kubernetes.io/azure-pip-tags annotation; keeping the first occurrence: ..."

kavix:
Thanks @Liunardy for the review!

I have addressed all the comments in commit ca0cbfa:
1. **Refactored `parseTags`**: Using a normalized map for (1)$ duplicate checking and returning the slice of dropped duplicate keys (`[]string`).
2. **Annotation-Specific Event**: Updated the `DuplicatePIPTags` warning event to be self-contained and reference `ServiceAnnotationAzurePIPTags` directly alongside the list of dropped duplicate keys. Removed the redundant error log.
3. **Cloud Config Log Level**: Log duplicate configuration keys as `.V(2).Info(...)` across all cloud configuration call sites.
4. **Extended E2E Test**: Added duplicate tag `,A=should-be-dropped` to the `service.beta.kubernetes.io/azure-pip-tags` E2E test case and added assertions for the `DuplicatePIPTags` warning event.
5. **Cleaned up tests**: Removed the Unicode case-folding test case and updated unit test suites to verify dropped keys and event emission.

pkg/provider/azure_utils_test.go
Liunardy: Should keep it, but expect no dropped keys.

pkg/provider/azure_loadbalancer.go
```diff
@@ -3626,10 +3626,19 @@ func shouldReleaseExistingOwnedPublicIP(
...
+	}
 	annotationTags := make(map[string]*string)
 	if _, ok := service.Annotations[consts.ServiceAnnotationAzurePIPTags]; ok {
-		annotationTags = parseTags(service.Annotations[consts.ServiceAnnotationAzurePIPTags], map[string]string{})
+		var droppedAnnotationKeys []string
+		annotationTags, droppedAnnotationKeys = parseTags(service.Annotations[consts.ServiceAnnotationAzurePIPTags], nil)
+		if len(droppedAnnotationKeys) > 0 {
+			az.Event(service, v1.EventTypeWarning, "DuplicatePIPTags", fmt.Sprintf(
+				"Ignoring duplicate tag keys (case-insensitive) in the %s annotation; keeping the first occurrence: %s",
+				consts.ServiceAnnotationAzurePIPTags, strings.Join(droppedAnnotationKeys, ", ")))
```
Liunardy, line 3640: This reads like the keys after ":" are the first occurrences. Probably better to say we keep the first occurrences and show these are the ignored/dropped duplicates/keys.

pkg/provider/azure_loadbalancer.go
```diff
@@ -3626,10 +3626,19 @@ func shouldReleaseExistingOwnedPublicIP(
 
 // ensurePIPTagged ensures the public IP of the service is tagged as configured
 func (az *Cloud) ensurePIPTagged(service *v1.Service, pip *armnetwork.PublicIPAddress) bool {
-	configTags := parseTags(az.Tags, az.TagsMap)
+	configTags, droppedConfigKeys := parseTags(az.Tags, az.TagsMap)
+	if len(droppedConfigKeys) > 0 {
+		log.Background().WithName("ensurePIPTagged").V(2).Info("Found duplicate tag keys in cloud configuration", "duplicateKeys", droppedConfigKeys)
```
Liunardy: Maybe rename duplicateKeys to droppedKeys/ignoredKeys so it is more precise.

pkg/provider/azure_utils_test.go
kavix: Restored the Unicode test case with `"S"` and `"ſ"` in `tagsMap`, asserting that both keys are preserved with zero dropped keys.

pkg/provider/azure_loadbalancer.go
kavix: Updated the event message to explicitly clarify that the trailing list shows the dropped duplicate keys:
`"Ignoring duplicate tag keys (case-insensitive) in the %s annotation; keeping the first occurrence, dropped duplicate keys: %s"`

pkg/provider/azure_loadbalancer.go
kavix: Renamed `"duplicateKeys"` to `"droppedKeys"` across all cloud configuration tag reconciliation call sites.

Liunardy:
Can we add to the the release note to mention the tag keys in cloud config tags/tagsMap and the service.beta.kubernetes.io/azure-pip-tags annotation and we are keeping the first occurrence.

pkg/provider/azure_utils_test.go
Liunardy: Can we add this case to TestReconcileTags
https://github.com/kubernetes-sigs/cloud-provider-azure/blob/ec98c027e9af53f6651d9035baee67377d89061d/pkg/provider/azure_utils_test.go#L37 
and the E2E?
https://github.com/kubernetes-sigs/cloud-provider-azure/blob/ec98c027e9af53f6651d9035baee67377d89061d/tests/e2e/network/service_annotations.go#L505
Expect the tags to contain both. I think we may need to fix findKeyInMapCaseInsensitive.

kavix:
Updated the release note in the PR description to explicitly mention tag keys in cloud config `tags`/`tagsMap` and the `service.beta.kubernetes.io/azure-pip-tags` annotation, and that the first occurrence is kept. Thanks @Liunardy!

pkg/provider/azure_utils_test.go
Liunardy: @kavix please take a look at this and the [latest copilot comment](https://github.com/kubernetes-sigs/cloud-provider-azure/pull/10906#discussion_r3934010552).

kavix:
Thanks @Liunardy! Addressed the feedback in commit 46e4504:

1. **Aligned `findKeyInMapCaseInsensitive` with `parseTags`**: Switched from `strings.EqualFold(k, key)` to `strings.ToLower(k) == strings.ToLower(key)` in `pkg/provider/azure_utils.go`. In Azure ARM, tags are not Unicode case-folded (`"S" != "ſ"`). Using `strings.ToLower` ensures downstream reconciliation preserves both keys without collision, resolving the discrepancy flagged in the recent Copilot comment.
2. **Extended Unit Tests**:
   - Added the Unicode test case with `"S"` and `"ſ"` to `TestReconcileTags` in `pkg/provider/azure_utils_test.go`, verifying both keys are preserved simultaneously on the resource.
   - Added a dedicated `TestFindKeyInMapCaseInsensitive` unit test suite to verify ASCII case-insensitivity while preserving distinct Unicode characters.
3. **Extended E2E Test**: Added `,S=first,ſ=second` to the `service.beta.kubernetes.io/azure-pip-tags` annotation in `tests/e2e/network/service_annotations.go` across both create and update flows, asserting that both keys are preserved on the public IP while `A=should-be-dropped` is dropped and triggers the `DuplicatePIPTags` warning event.

pkg/provider/azure_utils.go
```diff
@@ -85,13 +87,27 @@ func parseTags(tags string, tagsMap map[string]string) map[string]*string {
 				klog.Warning("parseTags: empty key, ignoring this key-value pair")
 				continue
 			}
+			kLower := strings.ToLower(k)
```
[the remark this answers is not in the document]
kavix: The discrepancy between `parseTags` and `reconcileTags` has been resolved in commit `46e4504`.

In Azure ARM, tags are not Unicode case-folded; characters like `"S"` and `"ſ"` are distinct valid tag keys (`"S" != "ſ"`). Rather than collapsing them in `parseTags`, `findKeyInMapCaseInsensitive` has been updated to use `strings.ToLower(k) == strings.ToLower(key)`.

Because both `parseTags` and `findKeyInMapCaseInsensitive` (and by extension `reconcileTags`) now consistently use `strings.ToLower`, `"S"` and `"ſ"` are treated as distinct keys across the entire lifecycle, eliminating any map-order nondeterminism.

pkg/provider/azure_utils_test.go
kavix: Addressed in commit `46e4504`:

1. **Fixed `findKeyInMapCaseInsensitive`**: Updated the comparison from `strings.EqualFold(k, key)` to `strings.ToLower(k) == strings.ToLower(key)` in `pkg/provider/azure_utils.go`. This aligns with `parseTags` and Azure ARM tag semantics (`"S" != "ſ"`), ensuring `reconcileTags` preserves both keys without collision. This also directly resolves Copilot's comment.
2. **Unit Tests**:
   - Added the Unicode test case with `"S"` and `"ſ"` to `TestReconcileTags` in `pkg/provider/azure_utils_test.go`, asserting that both keys are preserved on the resource.
   - Added a dedicated `TestFindKeyInMapCaseInsensitive` unit test suite.
3. **E2E Test**: Added `,S=first,ſ=second` to the `service.beta.kubernetes.io/azure-pip-tags` annotation in `tests/e2e/network/service_annotations.go` across both create and update scenarios, asserting that both keys are retained on the Public IP and the `DuplicatePIPTags` warning event is emitted for duplicate `A`.

Liunardy:
/lgtm
Will let another pair of eyes to check the changes and then we'll be good to merge this.

pkg/provider/azure_utils.go
```diff
@@ -100,20 +116,31 @@ func parseTags(tags string, tagsMap map[string]string) map[string]*string {
...
-				logger.V(4).Info("found identical keys from tags and tagsMap (case-insensitive), will replace", "identical keys", k, "keyFromTagsMap", key)
-				delete(formatted, k)
+			keyLower := strings.ToLower(key)
+			if _, exists := tagsMapNormalized[keyLower]; exists {
+				duplicateKeys = append(duplicateKeys, key)
+				continue
+			}
+			tagsMapNormalized[keyLower] = key
+
+			if kExisting, found := normalizedKeys[keyLower]; found {
```
nilo19, line 126: tags:          "owner=team,OWNER=team"
systemTags:    "aks-managed"
existing tags: {"OWNER": "team"}
In this case it seems OWNER=team will be deleted. Can we keep the OWNER unchanged, but ignore it and only respect the first occurence?

kavix: Thanks @nilo19! Fixed this by making the deletion check in `reconcileTags` case-insensitive using
  `findKeyInMapCaseInsensitive`. 

  Now, when `systemTags` is set, `OWNER` on the resource matches `owner` from the first occurrence, keeping `OWNER=team` unchanged without deleting it or triggering unnecessary updates. Added unit tests for this scenario and rebased on `master`.

kavix:
@Liunardy @nilo19 can u re-review this when u have a chance?

PR #10907, opened 2026-08-28
Feat: Fault domain behavior enhancement

### LINKED ISSUE #10222: Fault domain behavior enhancement

Change cloud-node-manager's topology labeling behavior so Azure availability zones and Azure platform fault domains are represented as separate topology concepts.

Specifically:

1. For standalone/regional VMs that are not running in an availability zone, stop using Azure `platformFaultDomain` as the fallback value for `topology.kubernetes.io/zone`.

   Today, this can result in labels such as:

   ```text
   topology.kubernetes.io/zone=0
   ```

   If the VM has no real availability zone, the Kubernetes zone label should be omitted rather than populated with the platform fault domain.

2. Expose Azure `platformFaultDomain` as a separate node label, including when availability zones are enabled.

   Suggested provider-scoped label:

   ```text
   topology.kubernetes.azure.com/fault-domain
   ```

   For example, a zonal VM could be labeled as:

   ```text
   topology.kubernetes.io/region=eastus
   topology.kubernetes.io/zone=eastus-1
   topology.kubernetes.azure.com/fault-domain=0
   ```

   A non-zonal standalone/regional VM could be labeled as:

   ```text
   topology.kubernetes.io/region=eastus
   topology.kubernetes.azure.com/fault-domain=0
   ```

   without also receiving `topology.kubernetes.io/zone=0`.

This likely needs compatibility and migration design. Existing nodes may already have `topology.kubernetes.io/zone=0`, and existing users may depend on that value for node selection, scheduling, monitoring, or automation. If this behavior changes, it should likely happen at a Kubernetes version boundary and be called out clearly in release notes.

#### Why is this needed:

Availability zone and platform fault domain are different failure-domain concepts, but the current behavior can mix them under the Kubernetes zone label.

This makes native Kubernetes topology spread constraints harder to reason about. For example, a user may try to spread across three real availability zones with:

```yaml
topologyKey: topology.kubernetes.io/zone
minDomains: 3
```

If the cluster also has non-zonal nodes labeled `topology.kubernetes.io/zone=0`, Kubernetes treats `0` as another topology domain. That can lead to surprising placement, reduced effective spread across real zones, or scheduling/provisioning failures caused by unexpected skew calculations.

The inverse use case is also important: some customers want to spread workloads across Azure platform fault domains within a single availability zone. Today, `platformFaultDomain` is available from IMDS, but when availability zones are enabled it is not exposed as a separate node label. This prevents users from expressing "spread across fault domains inside this zone" directly with Kubernetes topology spread.

Separating the labels would make the topology model clearer:

- `topology.kubernetes.io/zone` represents an availability zone.
- `topology.kubernetes.azure.com/fault-domain` represents Azure platform fault domain.
- Workloads can choose the topology dimension they actually care about.
- Users do not need to special-case unexpected `zone=0` values when they only intend to spread across real zones.

kavix:
#### What type of PR is this?

/kind feature

This PR changes the topology labeling behavior so that Azure availability zones and Azure platform fault domains are represented as separate topology concepts.
Specifically:
- For standalone/regional VMs that are not running in an availability zone, it stops using Azure `platformFaultDomain` as the fallback value for `topology.kubernetes.io/zone`. If the VM has no real availability zone, the Kubernetes zone label is omitted rather than populated with the platform fault domain.
- Exposes Azure `platformFaultDomain` as a separate node label `topology.kubernetes.azure.com/fault-domain`.

Fixes #10222

```release-note
Nodes no longer use platformFaultDomain as a fallback for topology.kubernetes.io/zone if an availability zone is not present. Instead, a new label topology.kubernetes.azure.com/fault-domain is used for fault domains.
```

```docs
```

### DIFF
--- a/patch_mock.go
+++ b/patch_mock.go
@@ -1,0 +1,58 @@
+package main
+
+import (
+	"fmt"
+	"os"
+	"strings"
+)
+
+func main() {
+	content, err := os.ReadFile("pkg/provider/azure_mock_vmsets.go")
+	if err != nil {
+		fmt.Println(err)
+		return
+	}
+
+	mockCode := `// GetPlatformFaultDomainByNodeName mocks base method.
+func (m *MockVMSet) GetPlatformFaultDomainByNodeName(ctx context.Context, name string) (string, error) {
+	m.ctrl.T.Helper()
+	ret := m.ctrl.Call(m, "GetPlatformFaultDomainByNodeName", ctx, name)
+	ret0, _ := ret[0].(string)
+	ret1, _ := ret[1].(error)
+	return ret0, ret1
+}
+
+// GetPlatformFaultDomainByNodeName indicates an expected call of GetPlatformFaultDomainByNodeName.
+func (mr *MockVMSetMockRecorder) GetPlatformFaultDomainByNodeName(ctx, name any) *MockVMSetGetPlatformFaultDomainByNodeNameCall {
+	mr.mock.ctrl.T.Helper()
+	call := mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "GetPlatformFaultDomainByNodeName", reflect.TypeOf((*MockVMSet)(nil).GetPlatformFaultDomainByNodeName), ctx, name)
+	return &MockVMSetGetPlatformFaultDomainByNodeNameCall{Call: call}
+}
+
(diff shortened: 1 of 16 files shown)


kavix:
Can u review this @Liunardy @cheftako

kavix:
Can u review this @Liunardy @cheftako

PR #10938, opened 2026-09-02
fix: classify standalone VMs on node name paths

### LINKED ISSUE #10217: bug: standalone VM nodes are still misclassified as VMSS on nodeName-based paths
#### What happened:

PR #10194 fixed standalone VM classification for `providerID` and `ipConfigurationID`, but `getVMManagementTypeByNodeName` still has the old unconditional short-circuit:

```go
if ss.DisableAvailabilitySetNodes && !ss.EnableVmssFlexNodes {
    return ManagedByVmssUniform, nil
}
```

That means a standalone VM node can still be misclassified as VMSS uniform on nodeName-based paths when:

- `vmType=vmss`
- `DisableAvailabilitySetNodes=true`
- `EnableVmssFlexNodes=false`

The remaining affected callers include at least:

- `GetPowerStatusByNodeName`
- `GetProvisioningStateByNodeName`
- `GetInstanceTypeByNodeName`
- `GetZoneByNodeName`
- `EnsureHostsInPool`
- `GetNodeVMSetName`

So after #10194, some standalone-VM flows are fixed, but nodeName-based flows can still take the VMSS handler incorrectly.

Standalone VM nodes should not be classified as `ManagedByVmssUniform` just because `DisableAvailabilitySetNodes=true`.

NodeName-based classification should be consistent with the providerID/ipConfigurationID fixes from #10194, so standalone VMs route to the availability-set/standalone-VM handler instead of the VMSS-uniform handler.

1. Configure Azure CCM with:
   - `vmType=vmss`
   - `DisableAvailabilitySetNodes=true`
   - `EnableVmssFlexNodes=false`
2. Have at least one real standalone VM node in the cluster (providerID format: `/providers/Microsoft.Compute/virtualMachines/...`, not a VMSS instance).
3. Trigger any nodeName-based CCM path, for example:
   - `GetPowerStatusByNodeName` / `GetProvisioningStateByNodeName`
   - `GetInstanceTypeByNodeName`
   - `GetZoneByNodeName`
   - `EnsureHostsInPool`
   - `GetNodeVMSetName`
4. Observe that `getVMManagementTypeByNodeName` returns `ManagedByVmssUniform` before consulting any non-VMSS cache/lookup, so the request is routed down the VMSS-uniform path.

Code pointers on current master:

- `pkg/provider/azure_vmss_cache.go`: `getVMManagementTypeByNodeName`
- `pkg/provider/azure_vmss_cache.go`: `getVMManagementTypeByProviderID`
- `pkg/provider/azure_vmss_cache.go`: `getVMManagementTypeByIPConfigurationID`

- This looks like the remaining half of the same standalone-VM problem family addressed by #10194.
- I validated live that providerID-based node lifecycle paths are active in CCM (`GetNodeNameByProviderID` from `node_lifecycle_controller`), and then checked the source to find the remaining unconditional nodeName short-circuit.
- On the live clusters I used for comparison, `DisableAvailabilitySetNodes` was not enabled, so this remaining bug was dormant there. That is why I did not see a before/after behavior change from the master image on those clusters.

#### Environment:

  - Affected by source on current master; live comparison was done on AKS/HCP clusters running custom CCM images.
  - Azure
  - `vmType=vmss`
  - standalone VM nodes present
  - bug requires `DisableAvailabilitySetNodes=true` and `EnableVmssFlexNodes=false`
  - Linux nodes
- Kernel (e.g. `uname -a`):
  - N/A
- Install tools:
  - CCM in managed AKS/HCP-style deployment
  - N/A
- Others:
  - Related fix: #10194

HaPhanBaoMinh:
#### What type of PR is this?

/kind bug

Classifies standalone VM nodes using the Kubernetes Node provider ID when availability-set nodes and VMSS Flex nodes are disabled. This preserves the no-VM-list optimization while preventing standalone VMs from being routed through the VMSS Uniform handler.

It also classifies nodes by provider ID in `EnsureHostsInPool` for the affected configuration so standalone VM NICs are updated through the correct handler.

Fixes #10217

AI assistance was used to analyze the routing flow and draft tests. I reviewed the changes and validated them with targeted, package, race, static, and lint checks.

Tests:

- `GOWORK=off go test -v ./pkg/provider -run '^(TestGetVMManagementTypeByNodeName|TestEnsureHostsInPool.*)$' -count=1`
- `GOWORK=off go test ./pkg/provider/... -count=1`
- `GOWORK=off go test -race ./pkg/provider -run '^(TestGetVMManagementTypeByNodeName|TestEnsureHostsInPoolRoutesStandaloneVM)$' -count=1`
- `make test-check`
- `GOWORK=off make lint`

```release-note
Fixed standalone VM nodes being misclassified as VMSS Uniform on node-name and load balancer backend-pool paths when availability-set and VMSS Flex nodes are disabled.
```

```docs

```

### DIFF
--- a/pkg/provider/azure_vmss.go
+++ b/pkg/provider/azure_vmss.go
@@ -1546,9 +1546,6 @@ func (ss *ScaleSet) ensureHostsInPool(ctx context.Context, service *v1.Service, 
 // participating in the specified LoadBalancer Backend Pool.
 func (ss *ScaleSet) EnsureHostsInPool(ctx context.Context, service *v1.Service, nodes []*v1.Node, backendPoolID string, vmSetNameOfLB string) error {
 	logger := log.FromContextOrBackground(ctx).WithName("EnsureHostsInPool")
-	if ss.DisableAvailabilitySetNodes && !ss.EnableVmssFlexNodes {
-		return ss.ensureHostsInPool(ctx, service, nodes, backendPoolID, vmSetNameOfLB)
-	}
 	vmssUniformNodes := make([]*v1.Node, 0)
 	vmssFlexNodes := make([]*v1.Node, 0)
 	vmasNodes := make([]*v1.Node, 0)
@@ -1571,9 +1568,14 @@ func (ss *ScaleSet) EnsureHostsInPool(ctx context.Context, service *v1.Service, 
 			continue
 		}
 
-		vmManagementType, err := ss.getVMManagementTypeByNodeName(ctx, localNodeName, azcache.CacheReadTypeDefault)
+		var vmManagementType VMManagementType
+		if ss.DisableAvailabilitySetNodes && !ss.EnableVmssFlexNodes && node.Spec.ProviderID != "" {
+			vmManagementType, err = ss.getVMManagementTypeByProviderID(ctx, node.Spec.ProviderID, azcache.CacheReadTypeDefault)
+		} else {
+			vmManagementType, err = ss.getVMManagementTypeByNodeName(ctx, localNodeName, azcache.CacheReadTypeDefault)
+		}
 		if err != nil {
-			logger.Error(err, "Failed to check vmManagementType by node name", "node", localNodeName)
+			logger.Error(err, "Failed to check VM management type", "node", localNodeName)
 			errors = append(errors, err)
 			continue
 		}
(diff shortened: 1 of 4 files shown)


HaPhanBaoMinh:
/cc @andyzhangx @anndono

Could you please take a look when you have time? Thanks.

HaPhanBaoMinh:
@andyzhangx @anndono Friendly ping — CI is green. Could you please review when you have time? Thanks!

PR #10943, opened 2026-09-03
fix: reject service reusing an existing private link service name

Liunardy:
#### What type of PR is this?
/kind bug

The PLS lookup is keyed by LB frontend IP configuration, so a Service whose `azure-pls-name` matches a PLS on a *different* frontend looks like "no PLS exists". CCM then PUTs that name, which repoints the existing PLS to the new Service's frontend and overwrites its `k8s-azure-owner-service` tag. The original Service keeps its IP but silently loses its PLS.

Reject the Service instead when the requested name is already used by a PLS on another frontend. A Service's own PLS is exempt via the `k8s-azure-owner-service` tag.

```release-note
[BREAKING CHANGE] fix: a Service whose `service.beta.kubernetes.io/azure-pls-name` matches a private link service already in use is now rejected, instead of repointing that private link service to the new Service. Action required: correct the annotation on any PLS-enabled Service reporting SyncLoadBalancerFailed after upgrade.
```

```docs

```

### DIFF
--- a/pkg/provider/azure_privatelinkservice.go
+++ b/pkg/provider/azure_privatelinkservice.go
@@ -149,6 +149,14 @@ func (az *Cloud) reconcilePrivateLinkService(
 			return false, err
 		}
 
+		if !exists {
+			// The lookup above is keyed by LB frontend, so it cannot see the name already
+			// taken on another frontend; creating would silently repoint that PLS here.
+			if err := az.checkPrivateLinkServiceNameConflict(ctx, service, plsName); err != nil {
+				return false, err
+			}
+		}
+
 		dirtyPLS, err := az.getExpectedPrivateLinkService(ctx, existingPLS, &plsName, &clusterName, service, fipConfig)
 		if err != nil {
 			return false, err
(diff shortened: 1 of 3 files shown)


nilo19:
Please include "BREAKING CHANGE" in the release note.

pkg/provider/azure_privatelinkservice.go
```diff
@@ -149,6 +149,14 @@ func (az *Cloud) reconcilePrivateLinkService(
 			return false, err
 		}
 
+		if !exists {
```
nilo19, line 152: I think it will block the internel->external transition. What's the current behavior? We should keep the behavior unchanged.

Liunardy: Current behaviour ARM responds back with `PrivateLinkServiceLoadBalancerReferenceCannotBeChanged`. Fixed to check the PLS owner, it was showing as PLS exist and owned by this service but different frontend and not sending the ARM request. Now it is the same behaviour.

Also checked the supported PLS sharing using the same frontend still works https://cloud-provider-azure.sigs.k8s.io/topics/pls-integration/#sharing-managed-privatelinkservice, PE connection to the PLS and different ports. 

This check will block the invalid way where a service uses a different frontend IP and the same pls name as an original PLS owned by another service, which silently repoints the PLS and makes the existing PE serve the new service's pods instead.

Liunardy:
> Please include "BREAKING CHANGE" in the release note.

Updated the release note with "BREAKING CHANGE" and "Action required".

nilo19:
Please update the doc on the correct way to reuse an existing PLS.

nilo19:
@Liunardy please cherrypick.

PR #10945, opened 2026-09-03
fix: retry loadbalancer metadata on transient IMDS failures

### LINKED ISSUE #10944: Transient IMDS failure on /metadata/loadbalancer causes node to briefly lose its ExternalIP
#### What happened:

On a cluster using a Standard SKU load balancer with `useInstanceMetadata`
enabled, a node intermittently lost its `NodeExternalIP` and then had it
restored a few minutes later, with no configuration change in between.

Digging into the node manager path, the address drop lines up with a
transient IMDS failure. When the node manager reconciles node addresses it
calls IMDS `/metadata/loadbalancer`. If that call comes back with a
transient error (in our case an HTTP 503), `getLoadBalancerMetadata`
returns an error, and `getMetadata` currently treats that error the same
as the benign "this VM isn't in a standard LB backend pool" case:

```go
loadBalancerMetadata, err := ims.getLoadBalancerMetadata()
if err != nil || loadBalancerMetadata == nil || loadBalancerMetadata.LoadBalancer == nil {
    logger.V(4).Info("Warning: failed to get loadbalancer metadata", "error", err)
    return instanceMetadata, nil
}
```

So on a transient blip it returns instance metadata without the load
balancer public IP. The node addresses then get published as Hostname +
InternalIP only, dropping the NodeExternalIP. On the next successful
reconcile the ExternalIP comes back.

From the outside it looks like the node loses its ExternalIP and gets it
back roughly 300s later (which matches the default node-status update
frequency). We confirmed the address was removed and re-added exactly one
reconcile interval apart.

A transient IMDS failure while fetching load balancer metadata shouldn't
cause the node to publish addresses without its ExternalIP. Ideally the
transient error is retried (the reconcile keeps the last good value)
rather than silently dropping the public IP, while the genuinely benign
"not in a standard LB backend pool" case still proceeds as today.

Hard to reproduce cleanly on a live cluster since it depends on IMDS
returning a transient error at the moment the node manager reconciles.
Conceptually:

- Standard SKU load balancer, node in the backend pool with a public IP,
useInstanceMetadata enabled.
- Wait for a transient failure from IMDS /metadata/loadbalancer
(e.g. a 503) during a node-address reconcile.
- Observe the node's EXTERNAL-IP drop to <none>, then return on the
next reconcile (~5m later with defaults).

It's more straightforward to demonstrate at the unit level by having the
fake IMDS return a non-200 on the loadbalancer endpoint and asserting the
node addresses aren't published without the ExternalIP.

N/A

#### Environment:

- Kubernetes version (use `kubectl version`): 1.3x.x
- Cloud provider or hardware configuration: Azure, Standard SKU LB
- Install tools: N/A
- Others:

jovieir:
#### What type of PR is this?
/kind bug

Right now `getMetadata` collapses two very different situations into the
same "benign" branch. When it calls `getLoadBalancerMetadata` and gets an
error, it just logs a `V(4)` warning and returns instance metadata without
the load balancer public IP. The comment says this is for the case where
the VM isn't part of a standard LoadBalancer backend pool, which is fair,
but the same branch also catches transient IMDS failures.

The way I read it, that's the bug: on a transient IMDS blip (a 503, a 429,
a request timeout, a dropped connection) the node ends up publishing its
addresses without the `NodeExternalIP`. The address only comes back on the
next successful reconcile, which with the default node-status update
frequency is about 5 minutes later. So from the outside it looks like a
node randomly loses its ExternalIP and gets it back exactly ~300s later.

This PR keeps the benign behaviour but stops swallowing the transient
failures:

- `getLoadBalancerMetadata` returns a typed `imdsResponseError` (carrying
  the status code) on a non-200 response, instead of a formatted string.
- `isTransientIMDSError` classifies `5xx`, `429`, `408`, and any non-HTTP
  error (network failure, timeout, JSON parse) as transient. Other `4xx`
  responses are treated as benign.
- On a transient error, `getMetadata` propagates it (wrapped) so the caller
  retries and the cache keeps the last good value. The benign path is
  untouched: it still logs the `V(4)` warning and returns instance metadata
  as before.

One thing I want to be upfront about: in practice the genuine "not in an LB
backend pool" case doesn't even hit the error branch. IMDS returns `200`
with an empty `publicIpAddresses` list rather than a `4xx`. So the benign
`4xx` handling here is really a safeguard for robustness, not the common
path. I kept it (and added a test for it) so the intent is explicit and
nobody folds it back into the transient branch later.

Fixes #10944

New unit tests in `TestNodeAddresses` cover both cases:
- a `503` on the loadbalancer metadata endpoint now surfaces an error and
  the node addresses are not published without the ExternalIP.
- a `404` (standing in for "not in a standard LB backend pool") returns no
  error and keeps Hostname + InternalIP, as before.

`go build`, `go vet`, `gofmt -l`, and `go test ./pkg/provider/ -run
TestNodeAddresses` are all clean locally.

```release-note
Fixed an issue where a transient IMDS failure while fetching load balancer
metadata could cause a node to temporarily lose its ExternalIP until the
next reconcile. Transient failures are now retried instead of publishing
node metadata without the load balancer public IP.
```

```docs

```

### DIFF
--- a/.agents/references/etag-cache.md
+++ b/.agents/references/etag-cache.md
@@ -1,5 +1,15 @@
 # ETag & Cache Invalidation
 
+## IMDS Partial Metadata
+
+`InstanceMetadataService` caches valid instance metadata together with an
+`LBMetadataError` when the load balancer endpoint returns a transient failure.
+Compute and identity reads remain usable. Local `NodeAddresses` returns the
+cached error without invalidating the entry, so node status addresses remain
+unchanged and repeated reads do not trigger an IMDS request burst. Normal cache
+expiry refreshes both endpoints; a successful refresh clears the error. Remote
+node address lookups through ARM are not blocked by the local LB error.
+
 ## Problem Overview
 
 Azure uses ETags for optimistic concurrency control. When the SDK sends a
--- a/pkg/provider/azure_instances_test.go
+++ b/pkg/provider/azure_instances_test.go
@@ -466,6 +466,7 @@ func TestNodeAddresses(t *testing.T) {
 		useCustomImsCache   bool
 		nilVMSet            bool
 		expectedErrMsg      error
+		lbStatusCode        int
 	}{
 		{
 			name:                "NodeAddresses should report error if metadata.Network is nil",
@@ -602,6 +603,39 @@ func TestNodeAddresses(t *testing.T) {
 				},
 			},
 		},
+		{
+			name:                "NodeAddresses should report error when loadbalancer metadata returns a transient failure",
+			nodeName:            "vm1",
+			metadataName:        "vm1",
+			vmType:              consts.VMTypeStandard,
+			ipV4:                "10.240.0.1",
+			ipV4Public:          "192.168.1.12",
+			loadBalancerSKU:     "standard",
+			lbStatusCode:        http.StatusServiceUnavailable,
+			useInstanceMetadata: true,
+			expectedErrMsg:      fmt.Errorf("failed to get loadbalancer metadata: %w", &imdsResponseError{statusCode: http.StatusServiceUnavailable}),
+		},
+		{
+			name:                "NodeAddresses should not report error and keep instance addresses when the VM is not in a standard LB backend pool",
+			nodeName:            "vm1",
+			metadataName:        "vm1",
+			vmType:              consts.VMTypeStandard,
+			ipV4:                "10.240.0.1",
+			ipV4Public:          "192.168.1.12",
+			loadBalancerSKU:     "standard",
+			lbStatusCode:        http.StatusNotFound,
+			useInstanceMetadata: true,
+			expectedAddress: []v1.NodeAddress{
+				{
+					Type:    v1.NodeHostName,
+					Address: "vm1",
+				},
+				{
+					Type:    v1.NodeInternalIP,
+					Address: "10.240.0.1",
+				},
+			},
+		},
 	}
 
 	for _, test := range testcases {
@@ -620,6 +654,10 @@ func TestNodeAddresses(t *testing.T) {
 		mux := http.NewServeMux()
 		mux.Handle("/", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 			if strings.Contains(r.RequestURI, consts.ImdsLoadBalancerURI) {
+				if test.lbStatusCode != 0 {
+					w.WriteHeader(test.lbStatusCode)
+					return
+				}
 				fmt.Fprintf(w, loadbalancerTemplate, test.ipV4Public, test.ipV4, test.ipV6Public, test.ipV6)
 				return
 			}
--- a/pkg/provider/azure_instances_test.go
+++ b/pkg/provider/azure_instances_test.go
@@ -21,7 +21,9 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"net/http/httptest"
 	"strings"
+	"sync/atomic"
 	"testing"
 	"time"
 
@@ -536,6 +538,17 @@ func TestNodeAddresses(t *testing.T) {
 			expectedAddress: expectedNodeAddress,
 		},
 		{
+			name:                "NodeAddresses should get remote instance addresses from Azure API despite a local loadbalancer metadata failure",
+			nodeName:            "vm1",
+			metadataName:        "vm2",
+			vmType:              consts.VMTypeStandard,
+			ipV4:                "10.240.0.2",
+			loadBalancerSKU:     "standard",
+			lbStatusCode:        http.StatusServiceUnavailable,
+			useInstanceMetadata: true,
+			expectedAddress:     expectedNodeAddress,
+		},
+		{
 			name:                "NodeAddresses should get IP addresses from local IMDS if node's name is equal to metadataName",
 			nodeName:            "vm1",
 			metadataName:        "vm1",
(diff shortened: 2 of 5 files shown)


pkg/provider/azure_instances_test.go
```diff
@@ -602,6 +603,39 @@ func TestNodeAddresses(t *testing.T) {
...
+			metadataName:        "vm1",
+			vmType:              consts.VMTypeStandard,
+			ipV4:                "10.240.0.1",
+			ipV4Public:          "192.168.1.12",
+			loadBalancerSKU:     "standard",
+			lbStatusCode:        http.StatusServiceUnavailable,
+			useInstanceMetadata: true,
+			expectedErrMsg:      fmt.Errorf("failed to get loadbalancer metadata: %w", &imdsResponseError{statusCode: http.StatusServiceUnavailable}),
+		},
+		{
```
nilo19, line 618: Other than testing new 503/504, can we also cover regression scenarios?
e.g., Existing ExternalIP → 503 → cache unchanged and no status patch → successful response with a different IP → updated ExternalIP.

nilo19:
Could we refactor the cache to retain valid instance metadata together with an `LBMetadataError` field? `NodeAddresses` could return that cached error until the next refresh, leaving the node’s addresses unchanged.

The current fix protects existing nodes, but for new nodes an LB failure also breaks VM identity lookup. That triggers the [[startup retry loop](https://github.com/kubernetes-sigs/cloud-provider-azure/blob/b2a724a43f3f2b724f53a5c29f3420cef10fae8e/pkg/nodemanager/nodemanager.go#L429-L435)](https://github.com/kubernetes-sigs/cloud-provider-azure/blob/b2a724a43f3f2b724f53a5c29f3420cef10fae8e/pkg/nodemanager/nodemanager.go#L429-L435), configured for [[20 attempts with 50-ms-plus-jitter delays](https://github.com/kubernetes-sigs/cloud-provider-azure/blob/b2a724a43f3f2b724f53a5c29f3420cef10fae8e/pkg/nodemanager/nodemanager.go#L109-L114)](https://github.com/kubernetes-sigs/cloud-provider-azure/blob/b2a724a43f3f2b724f53a5c29f3420cef10fae8e/pkg/nodemanager/nodemanager.go#L109-L114), risking IMDS throttling.

Keeping identity reads successful and caching the LB failure would avoid that request burst while waiting for a successful refresh.

pkg/provider/azure_instances_test.go
jovieir: Added regression coverage for ExternalIP preservation during transient failures and updates after recovery. The tests also verify that repeated reads reuse the cached error until expiry and that a successful refresh clears it. :-) lmk!

jovieir:
@nilo19 good catch on the startup retry loop. The cache now retains valid instance metadata alongside LBMetadataError, so identity reads succeed while NodeAddresses returns the cached error until refresh + added tests verifying that repeated identity/address reads make only one request per IMDS endpoint during the cache lifetime.

.agents/references/etag-cache.md
```diff
@@ -1,5 +1,15 @@
 # ETag & Cache Invalidation
 
+## IMDS Partial Metadata
```
nilo19, line 3: How is this related to ETag?

nilo19:
After some investigation I found this change is more of a tradeoff but a fix. Let's further discuss if this tradeoff is worth to be done.
For a new node init phase:
Before: valid instance metadata plus an LB 503 allowed initialization using the private IP; the cloud taint was removed.
With this change: [the LB error blocks address lookup](https://github.com/kubernetes-sigs/cloud-provider-azure/blob/96bb525a8b3bb001fb90def3b4f9cb14037969ce/pkg/provider/azure_instances_v1.go#L106-L108), so [initialization stops](https://github.com/kubernetes-sigs/cloud-provider-azure/blob/96bb525a8b3bb001fb90def3b4f9cb14037969ce/pkg/nodemanager/nodemanager.go#L498-L501) and the cloud taint remains.
This is intentional protection against publishing incomplete addresses, but it also affects nodes that have only private IPs.

jacobbaek:
Hi @nilo19 ,
What if the error only occurs when the external IP already exists?
Since a new node wouldn't have an external IP, it seems there wouldn't be an issue.

From what I've checked, it actually contains information about an external IP that was used previously.

`for _, address := range node.Status.Addresses {
  if address.Type == v1.NodeExternalIP {
    return XXXX
  }
}`

nilo19:
> Hi @nilo19 , What if the error only occurs when the external IP already exists? Since a new node wouldn't have an external IP, it seems there wouldn't be an issue.
> 
> From what I've checked, it actually contains information about an external IP that was used previously.
> 
> `for _, address := range node.Status.Addresses { if address.Type == v1.NodeExternalIP { return XXXX } }`

Have you checked the case where a new nodepool is created with node public IP enabled?

jacobbaek:
> > Hi @nilo19 , What if the error only occurs when the external IP already exists? Since a new node wouldn't have an external IP, it seems there wouldn't be an issue.
> > From what I've checked, it actually contains information about an external IP that was used previously.
> > `for _, address := range node.Status.Addresses { if address.Type == v1.NodeExternalIP { return XXXX } }`
> 
> Have you checked the case where a new nodepool is created with node public IP enabled?

I didn't check that.
IMHO, Isn't there a previously used public IP in the newly created nodepool?
After all, the nodes are being created from scratch.

nilo19:
> > > Hi @nilo19 , What if the error only occurs when the external IP already exists? Since a new node wouldn't have an external IP, it seems there wouldn't be an issue.
> > > From what I've checked, it actually contains information about an external IP that was used previously.
> > > `for _, address := range node.Status.Addresses { if address.Type == v1.NodeExternalIP { return XXXX } }`
> > 
> > 
> > Have you checked the case where a new nodepool is created with node public IP enabled?
> 
> I didn't check that. IMHO, Isn't there a previously used public IP in the newly created nodepool? After all, the nodes are being created from scratch.

Public IP will be created when the VM is created, before kubernetes node is created.

jacobbaek:
> > > > Hi @nilo19 , What if the error only occurs when the external IP already exists? Since a new node wouldn't have an external IP, it seems there wouldn't be an issue.
> > > > From what I've checked, it actually contains information about an external IP that was used previously.
> > > > `for _, address := range node.Status.Addresses { if address.Type == v1.NodeExternalIP { return XXXX } }`
> > > 
> > > 
> > > Have you checked the case where a new nodepool is created with node public IP enabled?
> > 
> > 
> > I didn't check that. IMHO, Isn't there a previously used public IP in the newly created nodepool? After all, the nodes are being created from scratch.
> 
> Public IP will be created when the VM is created, before kubernetes node is created.

I thought that the external IP was not included in `node.Status.Addresses` while the node was being created.
By "externalIP," you're referring to the public IP on the VM, right?

How about checking if the "node.cloudprovider.kubernetes.io/uninitialized" taint exists?
IMHO, It will be possible to prevent failure to remove taint that you mentioned.

Please correct me if I'm mistaken.

nilo19:
> > > > > Hi @nilo19 , What if the error only occurs when the external IP already exists? Since a new node wouldn't have an external IP, it seems there wouldn't be an issue.
> > > > > From what I've checked, it actually contains information about an external IP that was used previously.
> > > > > `for _, address := range node.Status.Addresses { if address.Type == v1.NodeExternalIP { return XXXX } }`
> > > > 
> > > > 
> > > > Have you checked the case where a new nodepool is created with node public IP enabled?
> > > 
> > > 
> > > I didn't check that. IMHO, Isn't there a previously used public IP in the newly created nodepool? After all, the nodes are being created from scratch.
> > 
> > 
> > Public IP will be created when the VM is created, before kubernetes node is created.
> 
> I thought that the external IP was not included in `node.Status.Addresses` while the node was being created. By "externalIP," you're referring to the public IP on the VM, right?
> 
> How about checking if the "node.cloudprovider.kubernetes.io/uninitialized" taint exists? IMHO, It will be possible to prevent failure to remove taint that you mentioned.
> 
> Please correct me if I'm mistaken.

Could you elaborate your plan?

jacobbaek:
I think it would be better to explain my suggestion with code.
I published the change in my fork.
https://github.com/jacobbaek/cloud-provider-azure/commit/e27a4418ab9bbff4b2385fcfa1015b8c3c59743f

nilo19:
> I think it would be better to explain my suggestion with code. I published the change in my fork. [jacobbaek@e27a441](https://github.com/jacobbaek/cloud-provider-azure/commit/e27a4418ab9bbff4b2385fcfa1015b8c3c59743f)

Two behavior changes observed.
During the node init phase:
If there is a successful instance-metadata response, but an LB-metadata 503, the proposal ignores the LB error during initialization, removes the cloud taint, then checks addresses again. The same error now causes failure because the taint is gone.
The node keeps its existing private IP, but CNM skips any needed address update and reports an initialization error. Retrying initialization does nothing; the update waits for a later successful sync. Previously, initialization could complete and publish the private addresses despite the LB error.
After the node init phase:
An LB metadata failure now blocks CNM from updating a node’s private IP, even when that private IP was retrieved successfully and there is no ExternalIP to protect. Previously, the private-IP update could proceed; now it may remain stale until a later successful reconciliation.
Plus, seems this change does not cover [instance v2](https://github.com/jacobbaek/cloud-provider-azure/blob/e27a4418ab9bbff4b2385fcfa1015b8c3c59743f/pkg/provider/azure_instances_v2.go#L172-L176).

jacobbaek:
> > I think it would be better to explain my suggestion with code. I published the change in my fork. [jacobbaek@e27a441](https://github.com/jacobbaek/cloud-provider-azure/commit/e27a4418ab9bbff4b2385fcfa1015b8c3c59743f)
> 
> Two behavior changes observed. During the node init phase: If there is a successful instance-metadata response, but an LB-metadata 503, the proposal ignores the LB error during initialization, removes the cloud taint, then checks addresses again. The same error now causes failure because the taint is gone. The node keeps its existing private IP, but CNM skips any needed address update and reports an initialization error. Retrying initialization does nothing; the update waits for a later successful sync. Previously, initialization could complete and publish the private addresses despite the LB error. After the node init phase: An LB metadata failure now blocks CNM from updating a node’s private IP, even when that private IP was retrieved successfully and there is no ExternalIP to protect. Previously, the private-IP update could proceed; now it may remain stale until a later successful reconciliation. Plus, seems this change does not cover [instance v2](https://github.com/jacobbaek/cloud-provider-azure/blob/e27a4418ab9bbff4b2385fcfa1015b8c3c59743f/pkg/provider/azure_instances_v2.go#L172-L176).

Could you please check this commit again?
https://github.com/jacobbaek/cloud-provider-azure/commit/d58485776b0b5c44a3e8cdd877339e0c0448b197

jovieir:
@nilo19 

Thanks for the review . I agree the cache/ETag behavior needs to be explicit here.

The path forward I see is:

- keep the transient IMDS failure path separate from the benign “not in a standard LB backend pool” path
- on transient errors, return the error so the caller retries instead of publishing node metadata without the ExternalIP
- keep the last good value in cache while the retry is pending
- only replace it after a successful refresh
- add/keep regression tests covering both the transient-failure case and the successful recovery/update case

This is not really an ETag-specific change. The issue is that the metadata fetch itself can fail transiently, and we should avoid treating that as “no load balancer metadata” and then publishing incomplete node addresses. The cache semantics here are about preserving the last successful value until a subsequent successful fetch, not about invalidating on ETag mismatches.

So my intent is only to:

- keep the benign 4xx behavior as a separate safeguard
- retry transient failures, not silently fall through

lmk

Issue #10729, opened 2026-07-24
Potential race condition on nodePrivateIPs between node informer and EndpointSlice informer / service reconciliation


Liunardy:
#### What happened:
az.nodePrivateIPs is a map[string]*utilsets.IgnoreCaseSet accessed concurrently by multiple informer goroutines without consistent locking, causing a data race that can crash the cloud-controller-manager with fatal error: concurrent map read and map write.

Writer (protected)
updateNodeCaches acquires nodeCachesLock before mutating the map. It is called by the node informer AddFunc/UpdateFunc/DeleteFunc handlers registered in SetInformers.
https://github.com/kubernetes-sigs/cloud-provider-azure/blob/4128235349f81a3f45d1f9e89f6c72a6a43a1ea2/pkg/provider/azure.go#L749-L752
https://github.com/kubernetes-sigs/cloud-provider-azure/blob/4128235349f81a3f45d1f9e89f6c72a6a43a1ea2/pkg/provider/azure.go#L701-L711

Readers (unprotected)
Example code paths reading nodePrivateIPs without lock:

1. EndpointSlice informer UpdateFunc (runs on the EndpointSlice informer processor goroutine): https://github.com/kubernetes-sigs/cloud-provider-azure/blob/4128235349f81a3f45d1f9e89f6c72a6a43a1ea2/pkg/provider/azure_local_services.go#L389-L396
2. checkAndApplyLocalServiceBackendPoolUpdates (called from EnsureLoadBalancer on a service controller worker goroutine): https://github.com/kubernetes-sigs/cloud-provider-azure/blob/4128235349f81a3f45d1f9e89f6c72a6a43a1ea2/pkg/provider/azure_local_services.go#L617
3. GetBackendPrivateIPs (called from reconcileSecurityGroup → EnsureLoadBalancer on a service controller worker goroutine):
https://github.com/kubernetes-sigs/cloud-provider-azure/blob/4128235349f81a3f45d1f9e89f6c72a6a43a1ea2/pkg/provider/azure_loadbalancer_backendpool.go#L281
4. ReconcileBackendPools (called from reconcileLoadBalancer → EnsureLoadBalancer on a service controller worker goroutine):
https://github.com/kubernetes-sigs/cloud-provider-azure/blob/4128235349f81a3f45d1f9e89f6c72a6a43a1ea2/pkg/provider/azure_loadbalancer_backendpool.go#L537

[...]

PR #10957, opened 2026-09-08
fix: retain deny-all NSG rule destinations other Services still need

Liunardy:
#### What type of PR is this?
/kind bug

There is one deny-all rule per IP family (e.g. `k8s-azure-lb_deny-all_IPv4`) shared by every Service of the same IP family. Reconciling a Service strips its own destinations from that rule, but `listSharedIPPortMapping` only reported the *ports* other Services still need, never their deny-all destinations. A destination another Service still required was therefore dropped, making that Service reachable from the whole virtual network.

`listSharedIPPortMapping` now also returns the deny-all destinations other Services still require, and `reconcileSecurityGroup` restores them with `EnsureDenyAllRules` after cleanup. This covers shared frontend IPs, backend node IPs when the floating IP is disabled, and `azure-additional-public-ips`.

Under multiple standard load balancers, a Service can have a backend pool of its own, so the nodes behind another Service are resolved per Service: from its endpoints when it is local, otherwise from the load balancer it is placed on. `EnsureLoadBalancerDeleted` records that placement before reconciling the security group, and a nil node list during init no longer clears the recorded nodes

```release-note
fix: keep the deny-all NSG rule intact for Services using `azure-deny-all-except-load-balancer-source-ranges` when another Service sharing one of its destinations is reconciled or deleted. Previously that destination could be dropped from the rule, leaving the Service reachable from anywhere in the virtual network rather than only its configured source ranges.
```

```docs

```

### DIFF
--- a/pkg/provider/azure_loadbalancer_accesscontrol.go
+++ b/pkg/provider/azure_loadbalancer_accesscontrol.go
@@ -50,18 +50,31 @@ func filterServicesByDisableFloatingIP(services []*v1.Service) []*v1.Service {
 	}, services)
 }
 
-// listSharedIPPortMapping lists the shared IP port mapping for the service excluding the service itself.
+// listSharedIPPortMapping lists the shared IP port mapping and the shared deny all destinations for
+// the service excluding the service itself.
 // There are scenarios where multiple services share the same public IP,
 // and in order to clean up the security rules, we need to know the port mapping of the shared IP.
+// The deny all rules are shared the same way, so a destination has to survive the cleanup while any
+// of those services still requires it. backendNodeIPs belong to every service that disables the floating
+// IP, because they all target the same nodes.
 func (az *Cloud) listSharedIPPortMapping(
 	ctx context.Context,
 	svc *v1.Service,
-	ingressIPs []netip.Addr,
-) (map[armnetwork.SecurityRuleProtocol][]int32, error) {
+	dstAddresses []netip.Addr,
+	backendNodeIPs []netip.Addr,
+) (map[armnetwork.SecurityRuleProtocol][]int32, []netip.Addr, error) {
 	var (
 		logger = log.FromContextOrBackground(ctx).WithName("listSharedIPPortMapping")
 		rv     = make(map[armnetwork.SecurityRuleProtocol][]int32)
+
+		isDstAddress = make(map[netip.Addr]bool, len(dstAddresses))
+		// The same address may be appended more than once. SetDestinationPrefixes will deduplicate.
+		denyAllDestinations []netip.Addr
+		denyAllRequired     bool
 	)
+	for _, addr := range dstAddresses {
+		isDstAddress[addr] = true
+	}
 
 	var services []*v1.Service
 	{
@@ -70,7 +83,7 @@ func (az *Cloud) listSharedIPPortMapping(
 		services, err = az.serviceLister.List(labels.Everything())
 		if err != nil {
 			logger.Error(err, "Failed to list all services")
-			return nil, fmt.Errorf("list all services: %w", err)
+			return nil, nil, fmt.Errorf("list all services: %w", err)
 		}
 		logger.V(5).Info("Listed all services", "num-all-services", len(services))
 
@@ -80,7 +93,7 @@ func (az *Cloud) listSharedIPPortMapping(
 			services = filterServicesByDisableFloatingIP(services)
 		} else {
 			logger.V(5).Info("Filter service by external IPs")
-			services = filterServicesByIngressIPs(services, ingressIPs)
+			services = filterServicesByIngressIPs(services, dstAddresses)
 		}
 	}
 	logger.V(5).Info("Filtered services", "num-filtered-services", len(services))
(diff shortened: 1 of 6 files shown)


pkg/provider/azure_loadbalancer_accesscontrol.go
```diff
@@ -94,17 +107,46 @@ func (az *Cloud) listSharedIPPortMapping(
...
+
+		for _, ing := range s.Status.LoadBalancer.Ingress {
+			if addr, err := netip.ParseAddr(ing.IP); err == nil && isDstAddress[addr] {
+				denyAllDestinations = append(denyAllDestinations, addr)
+			}
+		}
+	}
+
+	if denyAllRequired {
+		denyAllDestinations = append(denyAllDestinations, backendNodeIPs...)
```
nilo19, line 144: This assumes all services with floating IP disabled share the same set of backend nodes, but with multi-slb, cluster or local, this is not the case.

pkg/provider/azure_loadbalancer_accesscontrol.go
```diff
@@ -94,17 +107,46 @@ func (az *Cloud) listSharedIPPortMapping(
...
+				if isDstAddress[addr] {
+					denyAllDestinations = append(denyAllDestinations, addr)
+				}
+			}
+		}
+
+		// A service that disables the floating IP has no rule for its frontend IP; its nodes are
+		// appended after the loop.
+		if consts.IsK8sServiceDisableLoadBalancerFloatingIP(s) {
+			continue
```
nilo19, line 133: One corner case not being fixed:
service A has IP1, with additional IP X.
service B has IP2, with additional IP X.
If removing A, B will be skipped here so the deny-all rule will not be retained.
But since additional IP is designed to be pointed to IPs outside of this cluster, need to consider wether this needs to be fixed.

pkg/provider/azure_loadbalancer_accesscontrol.go
Liunardy: Added backendNodeIPsOfService to get IPs of the nodes backing the service under multi-slb.

Added early return for reconcileMultipleStandardLoadBalancerConfigurations during init if the nodes are nil and added this multi-slb config reconcile before reconcileSecurityGroup during EnsureLoadBalancerDeleted to initialise the multi-slb config if delete is first processed after restart.

pkg/provider/azure_loadbalancer_accesscontrol.go
Liunardy: This was partially fixed/covered. I changed the services filter, now filterServicesSharingDestinations, to fix one case.
Test cases covering this:
- `keeps the shared additional public IP when both deny-all Services use the floating IP`
- `keeps the backend node IP and the shared additional public IP when both deny-all Services disable the floating IP`
- `keeps the shared additional public IP when only the deleted deny-all Service disables the floating IP` <- this case was failing if the service with floating IP disabled just check for other services with floating IP disabled as well (missing the services that uses floating IP and share the same additional IP).
- `keeps the backend node IP and the shared additional public IP when only the other deny-all Service disables the floating IP`

pkg/provider/azure_loadbalancer_accesscontrol.go
```diff
@@ -44,24 +53,65 @@ func filterServicesByIngressIPs(services []*v1.Service, ips []netip.Addr) []*v1.
...
+			if az.isLoadBalancerInUseByService(svc, multiSLBConfig) {
+				nodeNames = az.getActiveNodesByLoadBalancerName(multiSLBConfig.Name)
+				break
+			}
+		}
+	}
+
+	rv := make([]netip.Addr, 0, nodeNames.Len())
+	for _, nodeName := range nodeNames.UnsortedList() {
+		for _, ip := range az.nodePrivateIPs[strings.ToLower(nodeName)].UnsortedList() {
```
nilo19, line 74: Since nodePrivateIPs is updated from a standalone informer, I think it's worth guarding all read cache calls under an RLock. Could be done in a following PR.

pkg/provider/azure_loadbalancer_accesscontrol.go
```diff
@@ -94,17 +144,46 @@ func (az *Cloud) listSharedIPPortMapping(
...
+				if isDstAddress[addr] {
+					denyAllDestinations = append(denyAllDestinations, addr)
+				}
+			}
+		}
+
+		// A service that disables the floating IP has no rule for its frontend IP; its rules
+		// target the nodes backing it instead.
+		if consts.IsK8sServiceDisableLoadBalancerFloatingIP(s) {
+			for _, addr := range sharedBackendNodeIPs(s) {
```
nilo19, line 169: Could you test the following case in a dual stack cluster?
service A and B run on the same node, A is ipv6, B is ipv4.
B request deny-all rule, should only protect ipv4 node IPs.
Reconcile A. Will this add ipv6 IPs to the rule?

pkg/provider/azure_loadbalancer_accesscontrol.go
Liunardy: Yeah, it's known issue https://github.com/kubernetes-sigs/cloud-provider-azure/issues/10729. Read under lock will need to be a different PR since there are many callers.

PR #10970, opened 2026-09-11
refactor: use lowest fixed versions in image CVE plans

nilo19:
#### What type of PR is this?

/kind cleanup

Use the lowest reported fixed version strictly newer than the installed version for each Go module finding, instead of the highest reported alternative. Combine findings for the same module by taking the maximum of those selected minima so every finding's upgrade requirement is retained.

- Stop planning with module and CVE context when version metadata is invalid or no newer fix is available.
- Add regression coverage for version ordering, multi-finding aggregation, Go directive selection, toolchain exclusions, and rescan safeguards.
- Simplify the skill instructions by removing implementation details already enforced by the helper while retaining agent-owned decisions and reporting boundaries.

Related: #10054

This changes only the shared skill, its helper, and its tests. It does not update provider runtime code, application dependencies, or container images. Go dependency resolution can still select a higher version, and rebuilt-image rescanning remains necessary to verify the planned CVEs are gone.

Validation:

- `python3 -B -m unittest discover -s .agents/skills/fix-image-cves/scripts -p 'test_*.py'` — all 47 tests passed, including the offline Go dependency-resolution test.
- Skill YAML metadata and placeholder checks passed.
- `git diff --check` passed.

No live images were built or scanned for this tooling change; the rescan regression uses mocked Trivy output.

```release-note
NONE
```

```docs

```

### DIFF
--- a/.agents/skills/fix-image-cves/SKILL.md
+++ b/.agents/skills/fix-image-cves/SKILL.md
@@ -5,164 +5,71 @@ description: Scan a built container image with Trivy, classify fixable Go-module
 
 # Fix Image CVEs
 
-## When To Use
+Use `scripts/fix_image_cves.py` for dependency selection, source updates, and
+verification. The helper implements the lowest-fixed-version policy and manages
+module, vendor, and license updates; it does not build or push images.
 
-Use this skill when you need to scan a built container image with Trivy,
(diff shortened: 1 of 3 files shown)


Liunardy (commented):
/lgtm

PR #10971, opened 2026-09-14
fix(e2e): ensure cleanup runs on service validation and setup failures

### LINKED ISSUE #10126: tests/e2e: ensureloadbalancer can skip service cleanup when post-create Get fails

anndono:
#### What type of PR is this?

/kind bug
/kind flake
Fixes resource cleanup in load-balancer e2e tests when setup or connectivity validation fails.
- Uses named return values so deferred Service cleanup observes the returned error, including errors from shadowed variables.
- Registers cleanup earlier to cover failed assertions and partially completed dual-stack setup.
- Ensures Service deletion completes before deleting public IPs and their resource group.
- Logs helper cleanup failures without replacing the original error and corrects pod cleanup error logging.
- Adds regression tests for failure cleanup, successful-call behavior, error preservation, and Service finalization.
These changes prevent skipped cleanup from leaving test resources behind.

Fixes #10126

```release-note
NONE
```

```docs

```


tests/e2e/network/ensureloadbalancer.go
```diff
@@ -267,6 +269,14 @@ var _ = Describe("Ensure LoadBalancer", Label(utils.TestSuiteLabelLB), func() {
 		v4Enabled, v6Enabled := utils.IfIPFamiliesEnabled(tc.IPFamily)
 		targetIPs := []*string{}
 		deleteFuncs := []func(){}
+		defer func() {
+			By("Cleaning up Service")
+			err := utils.DeleteService(cs, ns.Name, testServiceName)
+			Expect(err).NotTo(HaveOccurred())
+			for _, deleteFunc := range deleteFuncs {
+				deleteFunc()
```
nilo19, line 277: Seems if one deletion fails, the following deletions will all be skipped, leaving many orphaned resources. e.g., ipv6 pip failed, ipv4 pip deletion will be skipped.

Liunardy:
add Fixes #10126 ?

tests/e2e/network/ensureloadbalancer.go
Liunardy, line 1041: If IPv6 PIP creation fails, will the IPv4 PIP be deleted? Also, createPIPPrefix in "should respect service with BYO public IP prefix with various configurations".

tests/e2e/network/ensureloadbalancer.go
anndono: Added `runCleanupActions` to replace the original deletion logic to ensure one deletion failure no longer skips subsequent deletions.

tests/e2e/network/ensureloadbalancer.go
anndono: Moved the deferred cleanup before resource creation, so IPv4 cleanup runs even if IPv6 creation fails.

PR #11008, opened 2026-09-18
fix: reject managed frontend IPs in additional public IP annotations

Liunardy:
#### What type of PR is this?
/kind bug

The `service.beta.kubernetes.io/azure-additional-public-ips` annotation is intended for IPs outside managed load balancer frontends, such as Azure Global-tier public IPs. Allowing a Service to list a managed frontend IP can result in unintended NSG rule configuration.

This PR rejects such conflicts before updating Service status, the load balancer, or NSG state, while preserving valid additional public IP scenarios. Deletion remains allowed so invalid annotations do not block Service cleanup.

Also normalizes public IP cache keys for case-insensitive resource group lookup.

```release-note
[BREAKING CHANGE] fix: reject LoadBalancer Services that list a managed load balancer frontend IP in the `service.beta.kubernetes.io/azure-additional-public-ips` annotation. The annotation remains supported for IPs outside managed frontends, such as Azure Global-tier public IPs. Action required: remove managed frontend IPs from this annotation.
```

```docs

```


pkg/provider/azure_loadbalancer.go
```diff
@@ -1171,6 +1171,98 @@ func (az *Cloud) getServiceLoadBalancerStatus(ctx context.Context, service *v1.S
 	return &v1.LoadBalancerStatus{Ingress: lbIngresses}, lbIPsPrimaryPIPs, fipConfigs, nil
 }
 
+func (az *Cloud) validateAdditionalPublicIPs(ctx context.Context, service *v1.Service, managedLBs []*armnetwork.LoadBalancer) error {
```
nilo19, line 1174: Instead of just checking existing frontend IPs, also need to check existing static PIP that is going to be associated with the service in this reconciliation.

